PR #146, opened 2026-04-08
feat(sessions): session artifact metadata and list API

d-padmanabhan:
## Description

Session listing and detail APIs could not reliably expose attribution or lifecycle fields stored beside session artifacts, and `ListSessions` had no pagination or filters over those fields. This PR adds `metadata.json` under each session artifact directory, writes and completes it from `SaveSession` / delete paths (SQLite and PostgreSQL), merges it into gRPC and HTTP session payloads, and extends `ListSessions` with filters, `total_count`, and bounded `offset`/`limit`. On-disk `context` in `metadata.json` is allowlisted before API merge; session IDs used in artifact paths are validated to reject unsafe segments.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🔧 Refactoring (no functional changes)
- [ ] 🧪 Test update

## Related Issues

Closes #111

[Add Session Metadata Tracking for Agent Attribution](https://github.com/teradata-labs/loom/issues/111)

## Changes Made

- `pkg/artifacts/session_metadata.go`: `metadata.json` schema, allowlisted `context` keys (`user_id`, `project_id`, `conversation_id`), path validation, read/write helpers.
- Session stores and server: persist metadata on `SaveSession`, finalize on delete where applicable, preserve completed / `ended_at` and optional artifact stats on resync (`session_attribution.go`, `server.go`, `multi_agent.go`, `pkg/agent/session_store.go`, Postgres session store).
- `proto/loom/v1/loom.proto` plus regenerated `gen/go` and OpenAPI: new `Session` fields (`agent_name`, `started_at`, `ended_at`, `metadata_status`, `artifact_count`); `ListSessionsRequest` filters and pagination; `ListSessionsResponse.total_count`.
- `pkg/server/list_sessions.go`: reject negative `offset`; default limit 50, max 500; in-memory filter and page; conditional disk read of `metadata.json` when `metadata_status` or `project_id` filters need merged fields.
- Tests: `session_metadata_test.go`, `list_sessions_test.go`, `server_test.go` updates.
- Docs: `docs/architecture/artifacts.md`, `docs/guides/artifacts-usage.md`, `docs/reference/http-server.md` (not root `ARCHITECTURE.md`).

## Testing

**During development:** `GOWORK=off go test -tags fts5 -short ./pkg/server/... ./pkg/artifacts/...`

**Local full pass (2026-04-08):** With `buf` on `PATH` (`brew install buf`, typically `/opt/homebrew/bin`):

- `just test` completed successfully (`go test -tags fts5 -race -short -v ./...`).
- `just build` completed successfully (proto generate + all binaries).

**Not green locally:**

- `just race-check`: failed in `pkg/agent` with `panic: test timed out after 10m0s` under `-race -count=50` (recipe stops there).
- `just check`: failed at `golangci-lint` with 127 repo-wide staticcheck QF1012 (`WriteString`/`fmt.Sprintf`) hits, not specific to this feature branch.

No `.pre-commit-config.yaml` at repo root.

### Test Checklist

- [x] All new and existing tests pass (`just test`)
- [ ] Race detector shows zero race conditions (`just race-check`)
- [x] Code builds successfully (`just build`)
- [ ] All checks pass (`just check`)

### Test Coverage

- [ ] No tests needed (explain why):

## Proto Changes

- [ ] No proto changes in this PR
- [x] Proto changes included:
  - [x] `buf generate` executed successfully
  - [x] `buf lint` passes
  - [x] `buf format --diff --exit-code` passes
  - [ ] `buf breaking --against .git#branch=main` passes (or breaking changes justified)

## Documentation

- [ ] README.md updated (if user-facing changes)
- [ ] ARCHITECTURE.md updated (if architectural changes)
- [ ] TODO.md updated (if implementation plan changed)
- [ ] No documentation changes needed

## Security

- [ ] Security review required because:
- [ ] Added/updated secrets handling
- [ ] Gosec scan passes (`just security`)
- [ ] CodeQL analysis passes

## Performance

- [ ] No performance impact
- [ ] Performance benchmarks included
- [x] Performance impact justified:

`ListSessions` requests that filter on `metadata_status` or `project_id` may read `metadata.json` per session candidate so filters and `total_count` match merged state. Lists that do not need those fields avoid the per-session disk merge path.

## Deployment Notes

- [x] No special deployment steps needed
- [ ] Database migrations required
- [ ] Configuration changes required
- [ ] Other deployment notes:

### Required

- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [ ] My code follows the project's code style (gofmt, proto format)
- [ ] I have performed a self-review of my own code
- [x] I have commented my code in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings or errors
- [ ] All CI checks pass (proto, lint, test, build, security)
- [ ] Zero race conditions detected
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with `go test -race ./...`
- [x] Commit messages follow convention: `<type>(<scope>): <subject>`

### Optional

- [ ] I have updated the TODO.md with implementation notes
- [ ] I have added usage examples
- [ ] I have updated relevant diagrams
- [ ] I have tested on multiple platforms (Linux, macOS, Windows)

## Screenshots/Examples

Not applicable (API and on-disk artifact layout). Guide and reference updates live under `docs/`.


ilsiepotamus (changes requested):
see PR comment

d-padmanabhan:
> > > @d-padmanabhan lets discuss this implementation- mainly because this adds disk i/o into the hot path, and it breaks the pattern of keeping everything in the relational database. My preference would be to add columns to the sessions table (SQLite and Postgres): agent_name, ended_at, metadata_status, artifact_count, plus a context JSON column for the allowlisted attribution keys and eliminate the need for a metadata.json.
> > 
> > 
> > agreed @ilsiepotamus, disk reads on `ListSessions` are a real concern, and list/filter fields should live in the DB. I used `metadata.json` so metadata stayed colocated with artifacts for [#111](https://github.com/teradata-labs/loom/issues/111). I’m happy to move the canonical fields onto `sessions` (including a JSON `context` for allowlisted keys) and only touch disk when we need export/backfill - or drop `metadata.json` entirely if we don’t care about portability. Which way do you want to optimize?
> 
> I think `metadata.json` could still be useful, but I leave it to your discretion!

Thanks for the flexibility on `metadata.json`. I agree with the direction: make the relational store the source of truth for list and filter fields (agent_name, ended_at, metadata_status, artifact_count, plus a JSON column for the allowlisted context keys) so `ListSessions` stays off the disk hot path. I will treat `metadata.json` as optional: either a write-through copy for portability and inspection next to artifacts, or dropped if we decide export alone is enough. I will follow up with a concrete schema and migration plan (SQLite + Postgres) on this thread or in a follow-up PR after #146 lands, unless you prefer we fold the DB work into #146.

ilsiepotamus:
@d-padmanabhan can you address the gosec errors please?

d-padmanabhan:
> @d-padmanabhan can you address the gosec errors please?

fixed the gosec findings: int -> int32 conversions now use `types.SafeInt32`, and `ReadSessionArtifactMetadata` validates the path is under the session root before reading. `gosec ./...` is clean locally. Re-requesting review @ilsiepotamus

ilsiepotamus (changes requested):
Nice work on the allowlist + path-traversal defenses. Before this lands, a few issues need changes grouped by priority. The biggest ask: `metadata.json` must be **opt-out by default** so operators on Postgres/multi-replica deployments aren't forced onto the disk hot path.

### Must fix

**1. Add a config flag (default: disabled).**
There's no way to turn `metadata.json` off today — both session stores call `SyncSessionArtifactMetadata` unconditionally. Add a viper key (e.g. `artifacts.session_metadata_enabled`, env `LOOM_SESSION_METADATA_ENABLED`) that defaults to **false** and short-circuits inside `SyncSessionArtifactMetadata`, `CompleteSessionArtifactMetadata`, `mergeArtifactMetadataIntoProto`, and `listSessionsNeedsArtifactDisk`. Document the flag in `docs/reference/http-server.md` and `docs/architecture/artifacts.md`. When disabled, the API fields derived only from disk (`ended_at`, `artifact_count`, `metadata_status` from disk) simply stay empty.

**2. Move disk I/O out of the Postgres transaction.**
`pkg/storage/postgres/session_store.go:102` calls `SyncSessionArtifactMetadata` inside `execInTx`. That holds a pool connection + row lock across fsync. Move the call to after `execInTx` returns (gate on the new flag).

**3. Move disk I/O out of the SQLite global mutex.**
`pkg/agent/session_store.go:450` runs the same sync under `s.mu.Lock()`. `SaveSession` fires on every chat turn (`agent.go:1253`, `agent.go:1462`); holding the mutex across fsync serializes all session writes in the process. Release `s.mu` before calling `SyncSessionArtifactMetadata` (gate on the new flag).

**4. Fix the cross-tenant `DeleteSession` window.**
`pkg/server/server.go:459` and `pkg/server/multi_agent.go:2747` call `CompleteSessionArtifactMetadata(req.SessionId)` **before** the ownership/exists check. On RLS-enabled Postgres, a caller can flip another tenant's on-disk `metadata_status`/`ended_at` just by guessing a session id. Move the call inside the `found == true` branch after the tenant-scoped `GetSession` confirms ownership.

### Should fix

**5. Serialize read-modify-write on `metadata.json`.**
`SyncSessionArtifactMetadata` (and `CompleteSessionArtifactMetadata`) do read → mutate → atomic-rename without locking. Concurrent Sync + Complete can silently revert `completed` → `active`, so the "preserve completed" invariant in the docstring isn't actually enforced. Add a per-session mutex map (or flock on a sibling `.lock` file) around the read+write.

**6. Release the server RLock before the N disk reads in filtered `ListSessions`.**
`pkg/server/multi_agent.go:2706` holds `s.mu.RLock()` while looping per-session disk reads when `metadata_status` or `project_id` is set. For large session counts this blocks concurrent list ops. Gather the `*agent.Session` pointers under the lock, release, then merge from disk.

**7. Drop the hardcoded `MetadataStatus: "active"` when disk enrichment is off.**
`convertSession` sets `"active"` unconditionally. When `enrichFromArtifactDisk=false` (admin `ListAllSessions`, default `ListSessions`), completed sessions still show `metadata_status=active` — and the filter-vs-no-filter paths disagree. Leave the field empty when not enriched; clients can treat empty as "unknown".

**8. Lift patch coverage on `session_attribution.go` (currently 48.6%).**
Add tests for: malformed `metadata.json`, context keys outside the allowlist not leaking into `Session.metadata`, merge when `AgentID`/`AgentName` is empty on disk, merge when `metadata_status`/`project_id` filter is active but no file exists.

### Follow-up (OK to defer, please file)

**9. Multi-replica Postgres caveat.**
`metadata.json` is pod-local — on K8s with >1 replica, the pod that serves a read is rarely the pod that wrote. With the flag default-off (item 1) this is no longer a surprise, but document it explicitly in `artifacts.md` and note that the forthcoming DB-columns migration will resolve it.

**10. DB-columns migration thread.**
Per the earlier thread: `sessions.agent_name`, `sessions.ended_at`, `sessions.metadata_status`, `sessions.artifact_count`, plus a JSON `context` column (allowlisted keys). Once those land, `metadata.json` becomes a pure optional write-through for portability. Fine as a follow-up PR.

### Nits

- `ReadSessionArtifactMetadata` — wrap the `filepath.Rel` cause with `%w` instead of returning a bare `"invalid session metadata path"`; preserves security posture and improves debuggability.
- `SyncSessionArtifactMetadata(_ context.Context, ...)` — either thread `ctx` through or drop it from the signature.
- `BuildSessionArtifactMetadata` only preserves `"completed"`; any future status (`paused`, `abandoned`) gets silently overwritten. Either preserve any non-empty existing status or call out the limitation in the docstring.
- `ListSessionsResponse.total_count` — mention the `types.SafeInt32` saturation behavior so clients don't read 2^31-1 as "exactly this many".

ilsiepotamus:
@d-padmanabhan just bumping for visibility.  I can also add the proposed changes if you agree

d-padmanabhan:
> @d-padmanabhan just bumping for visibility. I can also add the proposed changes if you agree

@ilsiepotamus - sorry for the slow turnaround; this got sidelined behind other work priorities. I’ve just pushed `a2c4929` on `feat/session-metadata-attribution` with the review follow-ups (opt-in artifacts.session_metadata_enabled, safer tx/mutex boundaries around disk sync, delete ordering + tenant checks, list path lock scope, per-session metadata RMW locking, tests/docs). Thanks for the bump and for offering to land changes - if anything still looks off on your side, happy to adjust in a follow-up commit. I also pushed `c7706c7` to fix a flaky `FuzzPromptInterpolation` job (context deadline exceeded on large fuzz inputs), unrelated to the session-metadata behavior.

ilsiepotamus (changes requested):
hello @d-padmanabhan ! Sorry for such a long delay.  Just two more things to address —

  1. ListSessions is now silently paginated even with the feature flag off.

  listSessionsDefaultPageSize = 50 and listSessionsMaxPageSize = 500 are applied in normalizeListSessionsLimit regardless of
  artifacts.session_metadata_enabled. Existing callers that didn't pass limit previously got every session in one response and now
  get at most 50. That's a quiet breaking change visible to every client, not just opt-in ones, and it isn't flagged as breaking in
   the PR description.
  
  Could you either (a) preserve the old "return everything" behavior when limit is omitted (e.g. treat 0 as unbounded only when no
  pagination filters are set), or (b) call this out explicitly in the PR body / changelog as a breaking change for ListSessions? My
   preference is (a) since the pagination is independently useful when callers opt into it, and there's no need to force it on
  everyone.
  
  2. DeleteSession now does an extra sessionStore.LoadSession round-trip on every call, even when the metadata flag is off — and in
   Server.DeleteSession it runs even when the session is already known in-memory.

  In pkg/server/server.go:

  _, inMemory := s.agent.GetSession(req.SessionId)
  inStore := false
  if s.sessionStore != nil {
      if loaded, err := s.sessionStore.LoadSession(ctx, req.SessionId); err == nil && loaded != nil {
          inStore = true
      }
  }
  
  The LoadSession should be gated on !inMemory — if we already found the session in memory there's no reason to also query the
  store before deciding whether to delete. On Postgres-backed deployments this adds a network round-trip plus pulling the session
  row (and likely its messages, depending on LoadSession's impl) to every single delete on the happy path. A one-line fix:
  
  if !inMemory && s.sessionStore != nil {
      if loaded, err := s.sessionStore.LoadSession(ctx, req.SessionId); err == nil && loaded != nil {
          inStore = true
      }
  }
  
  MultiAgentServer.DeleteSession already gates this correctly (if deleteAgent == nil && s.sessionStore != nil), so the fix is to
  mirror that pattern.

  Two smaller things worth surfacing while you're in there:

  - The lock pattern in MultiAgentServer.DeleteSession flipped from Lock+defer Unlock wrapping the whole op to RLock-find → Unlock 
  → Lock-delete. Two concurrent DeleteSession(id) calls for the same ID can now both pass the lookup phase. The downstream ops
  (CompleteSessionArtifactMetadata and the store delete) are idempotent so this is benign, but the second call will return Success:
   true instead of NotFound. Worth a comment in code at least.
  - The new "session exists only in store, not in memory" case used to return 404 and now returns success. That's a correctness
  improvement (clients can finally clean up evicted sessions), but it's a semantics shift for clients keying off the response code.
   Should be in the changelog.

  For context on what I checked locally on 0ebf4a9:

  - just test: ✅ passes
  - just check: ✅ passes (the 127 QF1012 hits you saw aren't reproducible on the current head; the only failure I hit was a
  pre-existing TAVILY_API_KEY env leak in TestWebSearchTool_DefaultProvider_NoAPIKey, which fails identically on main)
  - just race-check: ✅ passes, zero races across all 50 iterations for pkg/agent (ok 593s) and the other four packages. The 10m
  panic you saw isn't reproducible — likely a one-off flake. Worth re-running on your end to confirm and ticking that box.

  Nice work otherwise — the metadata.json gating itself is genuinely zero-cost when off, which is what matters for default-off
  rollout. Thanks!

ilsiepotamus:
Pushed fixes for the 2026-05-22 review in `8d98abc` and merged latest `main` (the branch was 29 commits behind, now even).

1. **ListSessions pagination is now opt-in** — `pageProtoSessions` returns every session when neither `limit` nor `offset` is set (preserves the old "return everything" behavior; no silent 50-row truncation). Setting either parameter applies the default page size / 500 cap. Added `TestPageProtoSessions_PaginationOptIn`.
2. **`Server.DeleteSession`** now gates `sessionStore.LoadSession` on `!inMemory`, mirroring `MultiAgentServer.DeleteSession`, so we don't add a store round-trip on the happy path.
3. Documented the `MultiAgentServer.DeleteSession` RLock-find → Lock-delete window in a code comment (benign: idempotent ops; the losing concurrent call returns `Success` rather than `NotFound`).
4. Added a CHANGELOG `[Unreleased]` entry covering the opt-in pagination and the `DeleteSession` store-only success semantics change.

`go test -tags fts5 -race ./pkg/server/...` passes locally; CI is re-running. @d-padmanabhan — flagging since I pushed directly to your branch.

ilsiepotamus (approved):

PR #239, opened 2026-06-30
fix(quickstart): Bedrock API-key auth + non-exhaustive model picker (#166, #168)

ilsiepotamus:
Closes #166. Closes #168.

> **Note:** #166 and #168 both live in the same Bedrock block of `quickstart.sh`; separate PRs would conflict line-for-line, so they're combined here. Say the word if you'd rather split them.

## #166 — missing Bedrock API key / bearer-token option
The provider menu offered AWS Bedrock via **SSO/profile** and **access keys**, but not via a **Bedrock long-term API key** (bearer token) — the form needed for TD setup. The bearer reporter had to skip provider setup and hand-edit a yaml.

The client already supports it: `bedrock_bearer_token` is a recognized keyring key (`cmd/looms/config.go`) and the Bedrock client treats it as the highest-priority auth path, bypassing SigV4 (`pkg/llm/bedrock/client_sdk.go`). The only gap was the installer.

Added menu **option 3 — "AWS Bedrock (with API Key / Bearer Token)"**, which prompts for region + model and stores the key in the OS keyring via `config set-key bedrock_bearer_token`. The menu is renumbered (Anthropic 3→4 … Ollama 9→10, Skip 10→11), including the Ollama post-setup check.

## #168 — stale, hardcoded model subset
The picker hardcoded a small list (including the **deprecated Sonnet 3.5 v2**) and didn't surface that you can enter a full inference-profile ID, so users couldn't pick a newer model they have access to (the reporter wanted `us.anthropic.claude-sonnet-4-6`).

Extracted a shared `select_bedrock_model()` used by all three Bedrock auth flows. It lists current defaults (Sonnet 4.5 / Opus 4.5) and **prominently offers entering any inference profile ID** (option 3 → free-form, plus a paste-at-the-prompt fallback), with explicit text that the list is not exhaustive. This also de-duplicates the model prompt that was copy-pasted across the flows.

## Verification
- `bash -n quickstart.sh` passes.
- Menu options 1–11 map 1:1 to the `if/elif` branch chain (verified); Skip (11) falls through as before.
- `shellcheck` is not installed in this environment, and the interactive flow needs a live AWS account + keyring, so **please run the quickstart against Bedrock (each of the three auth methods) before merging.**

## Not done here (deliberately)
[...]

PR #242, opened 2026-07-01
test: de-flake unit CI — SSE happy path + live DuckDuckGo web-search calls

ilsiepotamus:
## Problem
`TestStreamableHTTPServer_POST_SSE_HappyPath` (added in #219) flakes in CI — e.g. it red-lit #239 (an unrelated `quickstart.sh`-only PR) with:

```
streamable_http_server_sse_test.go:162: Should be true   (stream handler should be invoked)
```

## Root cause (test-only race)
`handleStreamingPost` flushes the SSE response headers **before** invoking the stream handler:

```
w.WriteHeader(http.StatusOK)          // headers
flusher.Flush()
final, err := s.streamHandler.HandleMessageStream(...)   // sets sh.called = true
```

`http.Client.Do` (in the `postSSE` helper) returns as soon as those headers arrive. The test then asserted `sh.wasCalled()` **immediately**, racing the handler goroutine that sets `called = true`. It passes locally (handler wins the race) but flakes under CI load. No production bug — the transport behaves correctly.

## Fix
Move the `sh.wasCalled()` assertion to **after** `io.ReadAll(resp.Body)`. Once the full SSE body has been read, the handler has necessarily run and returned (the body *is* its output), so the check is deterministic. Added a comment explaining the ordering so it isn't moved back. Test-only change.

## Verification
```
go test -tags fts5 -race -run 'TestStreamableHTTPServer_POST_SSE|TestStreamableHTTPServer_SSE_ClientDisconnect' -count=100 ./pkg/mcp/transport/   # ok
```
100 iterations under `-race`, all green.

Branched off latest `main` (`2aab425`), since #219's test isn't in older branch points.


ilsiepotamus:
## Scope expanded: second flake fixed (a37f547)

This PR's own first CI run failed on a **different** flake — `TestWebSearchTool_MinimalUsage` + `TestWebSearchTool_SearchDuckDuckGo` (`pkg/shuttle/builtin`), both at exactly 30.03s, with the latter **panicking the test binary** via an unchecked `result.Data.(map[string]interface{})` on the nil result. (The SSE fix held — no SSE failure in that run.)

**Root cause:** four tests (`SearchDuckDuckGo`, `MinimalUsage`, `Execute_WithParameters`, `Concurrent` — the last firing **10 concurrent live calls**) queried the real DuckDuckGo Instant Answer API from unit CI. DDG throttles hosted runners → 30s client-timeout flakes. Brave/Tavily/SerpAPI tests were already `httptest`-mocked; DDG wasn't, because it needs no API key.

**Fix (test-only):**
- Shared `newMockedDDGTool` helper — local `httptest` server with a canned Instant Answer payload, injected via the tool's already-injectable `duckDuckGoEndpoint`. All four tests converted; `SearchDuckDuckGo` now deterministically asserts the parsed abstract + related topics.
- `require` + comma-ok on the casts, so a failure reports instead of panicking the binary.
- `Concurrent`: `require` → `assert` inside goroutines (`require` calls `t.FailNow`, which must only run on the test goroutine).

**Verified hermetic:** the full web-search test set passes with `HTTP_PROXY`/`HTTPS_PROXY` blackholed to `127.0.0.1:1` (`NO_PROXY=localhost`) — any remaining live-network call would fail instantly. Also `-race -count=5` on the set and the full `pkg/shuttle/builtin` package: green.

PR now covers both unit-CI flakes: the SSE happy-path race and the live-DDG network dependency.

Vasu2803k (approved):
Solid fix. Race condition moved to the right place (after body read), mocking pattern is clean, and concurrency test bug is fixed. No blockers.

PR #249, opened 2026-07-10
feat(agent): add ChatWithContentBlocks for multimodal user turns

Josh-Schoen:
## Description

Hosts embedding Loom cannot deliver images to the model: `Agent.Chat` and `Agent.ChatWithProgress` only accept a plain `string`, even though `types.Message.ContentBlocks` and all three provider clients (anthropic, bedrock, openai) already fully support multimodal image content blocks.

This adds `Agent.ChatWithContentBlocks(ctx, sessionID, userMessage, contentBlocks, progressCallback)` — it mirrors `ChatWithProgress` but sets `ContentBlocks` on the user turn. `userMessage` remains the canonical text (persistence, graph-memory extraction, providers without multimodal support); `progressCallback` may be nil, making it also the blocks-capable equivalent of plain `Chat`. Also exports the `ContentBlock` alias from `pkg/agent` alongside the existing `Message` alias.

Motivating use case: loom-cloud (Tera) chat attachments — user-uploaded images resolved from the artifact store and delivered to vision-capable models. Verified end-to-end there against Bedrock Claude Sonnet (model accurately described attached test images).

## Type of Change

## Changes Made

- Add `Agent.ChatWithContentBlocks` in `pkg/agent/agent.go`, with `ContentBlocks` set on the user `Message` and a distinct span name `agent.chat_with_content_blocks`.
- Export `type ContentBlock = types.ContentBlock` alias in `pkg/agent/types.go`.
- **Refactor (addresses review):** extract a private `(*Agent).chat(ctx, sessionID, userMessage, chatParams)` helper carrying the shared lifecycle (span setup, persistence, graph-memory kick-off, conversation loop, success/error telemetry). `Chat`, `ChatWithProgress`, and `ChatWithContentBlocks` are now thin wrappers over it — removing ~300 lines of duplication and reconciling with #255: all three now use `truncatePreview` (previously this method still used `truncateString(…,100)`). `chatParams.spanName` plus `reportProgress`/`reportContentBlocks` keep each entry point's span/telemetry byte-identical to before (the sole exception, documented in the commit, is that `ChatWithProgress`'s `has_progress` event field is now `callback != nil` rather than a hardcoded `true`).

### Test Checklist

- [x] All new and existing tests pass (`go test -tags fts5 ./pkg/agent/` — ok)
- [x] Race detector shows zero race conditions (`go test -tags fts5 -race ./pkg/agent/` — ok; CI also runs `-race`)
- [x] All checks pass (CI)

### Test Coverage

- [x] Unit tests added — `pkg/agent/chat_content_blocks_test.go`:
  - **PropagatesBlocksToLLM** — a capturing mock LLM asserts at the provider boundary that text + image blocks actually reach the request (the core guarantee).
  - **PersistsCanonicalText** — `userMessage` stays the canonical text on the stored turn while `ContentBlocks` ride along; assistant reply appended.
  - **NilProgressCallback / EmptyBlocks** — degrade cleanly to a plain text turn.
  - **SuccessWithProgressCallback** — success path with a non-nil callback emits no failure event.
  - **ErrorEmitsFailureEvent / ErrorNilCallback** — failing LLM propagates the error, returns nil, and emits `StageFailed` only when a callback exists.
  - **ContentBlockAlias** — documents the re-exported `pkg/agent.ContentBlock`.

## Proto Changes

## Documentation

## Security

## Performance

- [x] No performance impact

## Deployment Notes

- [x] No special deployment steps needed

## Additional Context

Follow-up (not in this PR): `ContentBlocks` are **not** serialized by either session store — `pkg/agent/session_store.go` (SQLite) and `pkg/storage/postgres/session_store.go` both persist only `Content` and the tool fields — so image blocks are dropped on persistence. A session rehydrated after a restart loses the images and the model sees only the canonical text mid-conversation. Separately, token accounting counts only `msg.Content`, so within a live session images ride in L1 history and are re-sent on every subsequent provider call while the memory manager sees a ~10-token message. Both belong in a tracking issue (block survival across rehydration + image-aware token accounting).

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -1488,10 +1488,16 @@ func (a *Agent) ChatWithProgress(ctx context.Context, sessionID string, userMess
 // ChatWithContentBlocks is like ChatWithProgress but the user turn carries
 // multimodal content blocks (text and/or images) alongside the plain-text
 // content. userMessage remains the canonical text (used for persistence,
-// graph-memory extraction, and providers without multimodal support);
-// contentBlocks take precedence when building the provider request for
-// providers that support them. progressCallback may be nil, in which case no
-// progress events are emitted (equivalent to Chat).
+// graph-memory extraction, and providers without multimodal support).
+//
+// When contentBlocks is non-empty, providers build the request from the blocks
+// only — Content is not sent. To keep userMessage canonical, if contentBlocks
+// contains no text block, userMessage is automatically prepended as one, so an
+// image-only call still delivers the question text to the model. Callers that
+// include their own text block are unaffected.
+//
+// progressCallback may be nil, in which case no progress events are emitted
+// (equivalent to Chat).
 func (a *Agent) ChatWithContentBlocks(ctx context.Context, sessionID string, userMessage string, contentBlocks []ContentBlock, progressCallback ProgressCallback) (*Response, error) {
 	return a.chat(ctx, sessionID, userMessage, chatParams{
 		spanName:            "agent.chat_with_content_blocks",
@@ -1528,12 +1534,31 @@ type chatParams struct {
 	reportContentBlocks bool
 }
 
+// hasTextBlock reports whether any of the blocks is a text block.
+func hasTextBlock(blocks []ContentBlock) bool {
+	for _, b := range blocks {
+		if b.Type == "text" {
+			return true
+		}
+	}
+	return false
+}
+
 // chat runs the full conversation lifecycle — span setup, user-message
 // persistence, graph-memory kick-off, the conversation loop, and success/error
 // telemetry — shared by the three public chat entry points. See chatParams for
 // how each entry point tailors span name, multimodal content, and progress
 // reporting.
 func (a *Agent) chat(ctx context.Context, sessionID string, userMessage string, p chatParams) (*Response, error) {
+	// Providers build the request exclusively from ContentBlocks when present,
+	// dropping Content — so an image-only block set would silently send the
+	// image without the question text. Keep userMessage truly canonical by
+	// prepending it as a text block when the caller didn't include one. This is
+	// a no-op for callers that already carry their prompt in a text block.
+	if len(p.contentBlocks) > 0 && !hasTextBlock(p.contentBlocks) && userMessage != "" {
+		p.contentBlocks = append([]ContentBlock{{Type: "text", Text: userMessage}}, p.contentBlocks...)
+	}
+
 	// Inject session ID into context for tool access
 	ctx = session.WithSessionID(ctx, sessionID)
 
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -1469,18 +1469,86 @@ func truncatePreview(s string) string {
 // Chat processes a user message and returns a response.
 // This is the main entry point for conversational interaction.
 func (a *Agent) Chat(ctx context.Context, sessionID string, userMessage string) (*Response, error) {
+	return a.chat(ctx, sessionID, userMessage, chatParams{
+		spanName: "agent.chat",
+	})
+}
+
+// ChatWithProgress is like Chat but supports streaming progress updates.
+// The progressCallback will be called at key execution stages to report progress.
+// This is used by StreamWeave to provide real-time feedback to clients.
+func (a *Agent) ChatWithProgress(ctx context.Context, sessionID string, userMessage string, progressCallback ProgressCallback) (*Response, error) {
+	return a.chat(ctx, sessionID, userMessage, chatParams{
+		spanName:         "agent.chat_with_progress",
+		progressCallback: progressCallback,
+		reportProgress:   true,
+	})
+}
+
+// ChatWithContentBlocks is like ChatWithProgress but the user turn carries
+// multimodal content blocks (text and/or images) alongside the plain-text
+// content. userMessage remains the canonical text (used for persistence,
+// graph-memory extraction, and providers without multimodal support);
+// contentBlocks take precedence when building the provider request for
+// providers that support them. progressCallback may be nil, in which case no
+// progress events are emitted (equivalent to Chat).
+func (a *Agent) ChatWithContentBlocks(ctx context.Context, sessionID string, userMessage string, contentBlocks []ContentBlock, progressCallback ProgressCallback) (*Response, error) {
+	return a.chat(ctx, sessionID, userMessage, chatParams{
+		spanName:            "agent.chat_with_content_blocks",
+		contentBlocks:       contentBlocks,
+		progressCallback:    progressCallback,
+		reportProgress:      true,
+		reportContentBlocks: true,
+	})
+}
+
+// chatParams configures the shared conversation lifecycle behind Chat,
+// ChatWithProgress, and ChatWithContentBlocks. Each public entry point sets only
+// the fields it needs so its span and telemetry stay identical to the
+// pre-refactor, hand-written form.
+type chatParams struct {
+	// spanName is the trace span name for this entry point.
+	spanName string
+
+	// contentBlocks, when present, is attached to the user turn so multimodal
+	// content reaches providers that support it. nil for text-only entry points.
+	contentBlocks []ContentBlock
+
+	// progressCallback, when non-nil, is threaded into context so nested
+	// operations (tools, backends, sub-agents) can report progress, and it
+	// receives a StageFailed event if the conversation loop errors.
+	progressCallback ProgressCallback
+
+	// reportProgress adds the "has_progress" field to the conversation.started
+	// event. Chat omits it; ChatWithProgress and ChatWithContentBlocks set it.
+	reportProgress bool
+
+	// reportContentBlocks adds the "message.content_blocks" span attribute and
+	// the "content_blocks" started-event field. ChatWithContentBlocks only.
+	reportContentBlocks bool
+}
+
+// chat runs the full conversation lifecycle — span setup, user-message
+// persistence, graph-memory kick-off, the conversation loop, and success/error
+// telemetry — shared by the three public chat entry points. See chatParams for
+// how each entry point tailors span name, multimodal content, and progress
+// reporting.
+func (a *Agent) chat(ctx context.Context, sessionID string, userMessage string, p chatParams) (*Response, error) {
 	// Inject session ID into context for tool access
 	ctx = session.WithSessionID(ctx, sessionID)
 
 	// Start trace span — always created; NoOpTracer handles disabled case
 	startTime := time.Now()
-	ctx, span := a.tracer.StartSpan(ctx, "agent.chat")
+	ctx, span := a.tracer.StartSpan(ctx, p.spanName)
 	defer a.tracer.EndSpan(span)
 
 	// Set initial attributes
 	span.SetAttribute(observability.AttrSessionID, sessionID)
 	span.SetAttribute("message.length", len(userMessage))
 	span.SetAttribute("message.preview", truncatePreview(userMessage))
+	if p.reportContentBlocks {
+		span.SetAttribute("message.content_blocks", len(p.contentBlocks))
+	}
 	a.mu.RLock()
 	currentLLM := a.llm
 	a.mu.RUnlock()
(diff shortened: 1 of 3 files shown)


ilsiepotamus:
@Josh-Schoen need unit tests

Josh-Schoen:
> @Josh-Schoen need unit tests

@ilsiepotamus done.

ilsiepotamus (changes requested):
The feature, API shape, and especially the tests are in good shape — but this adds a third verbatim copy of a ~215-line lifecycle, and there's an in-flight collision that makes the duplication concrete rather than stylistic.

## What's good

- **The tests are the strongest part** (`chat_content_blocks_test.go`): the `capturingLLM` asserts at the provider boundary that blocks actually reach the LLM request — the core guarantee, tested at the right seam. Coverage includes canonical-text persistence, nil callback, empty blocks degrading to a text turn, error paths with and without callback, and the alias. Testify, mutex-guarded mocks, deterministic (classifier disabled).
- **Telemetry hygiene**: only the block *count* goes on the span (`message.content_blocks`); no base64 image bytes leak into traces.
- **Zero regression risk as written**: existing `Chat`/`ChatWithProgress` paths untouched, and the method doc comment is clear about the `userMessage`-as-canonical-text contract.

## 🔴 Blocking: third verbatim copy of the chat lifecycle

`ChatWithContentBlocks` is `ChatWithProgress` copied wholesale with two functional lines changed (`ContentBlocks:` on the user message, span name + one attribute). Diffing `Chat` vs `ChatWithProgress` on main shows they too differ *only* in progress-callback handling and span name — so after this merges the repo carries **three identical copies** of span setup, persistence, graph-memory kickoff, error handling, and ~15 metric emissions.

This is already colliding, not hypothetically: **PR #255 (in flight) changes the `message.preview`/`response.preview` truncation in `Chat` and `ChatWithProgress`** (`truncateString(…,100)` → 200-rune `truncatePreview`). This PR's copy uses `truncateString(…,100)`, so whichever merges second leaves the third copy silently divergent — exactly the tech-debt pattern CLAUDE.md prohibits.

**Ask**: extract one internal helper —

```go
func (a *Agent) chat(ctx context.Context, sessionID, userMessage string,
    blocks []ContentBlock, cb ProgressCallback, spanName string) (*Response, error)
```

— and make all three public methods thin wrappers (passing `"agent.chat"`, `"agent.chat_with_progress"`, `"agent.chat_with_content_blocks"` through, so existing telemetry is byte-identical). Mechanical refactor; the existing test suites already cover all three entry points.

## 🟡 Minor

- **PR description Testing section is stale**: it still says "No tests needed (explain why)" with the race-detector box unchecked, but the second commit added the 235-line test file and CI (which runs `-race`) is green. Please update the checklist.
- **Branch is BEHIND main** — needs a rebase before merge regardless.
- No test for the *success* path with a non-nil progress callback (only the failure path exercises one) — trivial add while refactoring.
- The unchecked type assertions (`response.Metadata["turns"].(int)`) are copied from the siblings — fine for consistency, and the helper refactor gives one place to harden them later.
- **Follow-up issue, not blocking**: persisted user messages now carry base64 image payloads into session storage, and token counting (tiktoken over `Content`) doesn't account for image bytes toward context budgets. A large-attachment policy (strip/reference blocks on persistence) deserves a tracking issue.

With the internal-helper refactor and a rebase this is ready — happy to re-review quickly.

Josh-Schoen:
Thanks for the thorough review — all points addressed. Rebased onto latest `main` (which now includes #255) and pushed.

## 🔴 Duplication — done

Extracted a private `(*Agent).chat(ctx, sessionID, userMessage, chatParams)` helper carrying the whole lifecycle. `Chat`, `ChatWithProgress`, and `ChatWithContentBlocks` are now thin wrappers (~300 lines of duplication removed). `chatParams` passes the span name through plus two flags (`reportProgress`, `reportContentBlocks`) so each entry point's span, attributes, and events stay **byte-identical** to before.

This also makes the #255 collision you flagged concrete and resolves it: after rebasing, `Chat`/`ChatWithProgress` were on `truncatePreview` while this method was still on `truncateString(…,100)`. All three now share `truncatePreview` through the helper, so there's a single place going forward.

One intentional, documented telemetry change: `ChatWithProgress`'s `conversation.started` event now reports `has_progress = (callback != nil)` instead of a hardcoded `true`. It only differs in the degenerate nil-callback case and is strictly more correct. Everything else (span names, all attributes, all event fields, all metric emissions, the `StageFailed` emission) is unchanged — verified by diffing the telemetry surface and by the existing instrumentation tests passing under `-race`.

## 🟡 Minor

- **Stale Testing section** — updated: tests added, race box checked (`go test -tags fts5 -race ./pkg/agent/` green locally + CI).
- **Behind main** — rebased.
- **Success-path callback test** — added `TestAgent_ChatWithContentBlocks_SuccessWithProgressCallback` (success with a non-nil callback emits no failure event).
- **Unchecked type assertions** (`response.Metadata["turns"].(int)`) — now in the one shared helper, so there's a single place to harden later if we want.
- **Large-attachment follow-up** — noted in the PR description; happy to open a tracking issue for a strip/reference-on-persistence policy + image-aware token accounting. Out of scope here.

Ready for re-review.

Josh-Schoen:
@ilsiepotamus updated. Please review when you have time.

ilsiepotamus (dismissed):
**Verdict: Approve, with two comments worth addressing before merge** (one PR-description inaccuracy, one API footgun). CI is fully green including race detector, fuzz, and lint. The duplication flagged in my earlier review was genuinely addressed — this is now a net −206 line change.

## Refactor correctness — verified ✅

I diffed the new shared helper against both deleted bodies:

- `Chat`'s behavior is preserved exactly: no `has_progress` event field (`reportProgress: false`), no `ContextWithProgressCallback` injection (nil callback), no `StageFailed` on error, and identical success telemetry including the `"status": "success"` tag on `MetricAgentConversations`.
- The only behavior change is the one documented in the PR body: `ChatWithProgress`'s `has_progress` is now `callback != nil` instead of hardcoded `true`. That's a correctness improvement, not a regression.
- No fourth copy was missed — `Chat` and `ChatWithProgress` were the only lifecycle duplicates on main.
- Propagation path is sound: `llmtypes.Message` is an alias of `types.Message`, so blocks flow from the session history straight into the provider clients, all of which (anthropic, bedrock ×3, openai, gemini, ollama) already branch on `len(msg.ContentBlocks) > 0`.

## Tests — strong ✅

`chat_content_blocks_test.go` asserts at the provider boundary with a mutex-guarded `capturingLLM` (race-safe), and covers the right matrix: block propagation, canonical-text persistence, nil callback, empty blocks, success-with-callback (no spurious `StageFailed`), error with and without callback, and the alias. This is the right shape of test for the feature's core guarantee.

## Issues

### 1. The PR body's persistence claim is inaccurate (docs honesty) — medium

The "Additional Context" section says *"persisted user messages now carry base64 image payloads into session storage."* They don't. Neither store serializes `ContentBlocks`:

- `pkg/agent/session_store.go` (SQLite `SaveMessage`) — inserts `content`, `tool_calls_json`, etc.; no `ContentBlocks` column or marshal.
- `pkg/storage/postgres/session_store.go` — same.

So the real follow-up gap is the **opposite** of what's described: blocks are silently dropped at persistence, meaning a session rehydrated after restart loses the images and the model sees only canonical text mid-conversation. That may be acceptable by design (canonical-text philosophy), but the follow-up tracking issue should target "blocks don't survive rehydration" rather than "storage bloat." Please correct the PR body before merge — and actually file that tracking issue rather than leaving it as prose.

### 2. API footgun: blocks *replace* Content at the provider, they don't augment it — medium/low

When `ContentBlocks` is non-empty, every provider ignores `msg.Content` entirely (e.g., the blocks branch in `pkg/llm/anthropic/client.go` never includes `Content`). So this call:

```go
ag.ChatWithContentBlocks(ctx, id, "What is in this image?", []ContentBlock{imageOnlyBlock}, nil)
```

sends the image with **no question text** to a multimodal model — silently. The doc comment says blocks "take precedence" but doesn't state the caller obligation. Two options:

- **(a) Defensive (preferred):** in `chat()`, when `len(contentBlocks) > 0` and none has `Type == "text"`, prepend `{Type: "text", Text: userMessage}`. This makes `userMessage` truly canonical and costs nothing for callers that already include a text block.
- **(b) Minimal:** extend the doc comment to say "callers must include the prompt text as a text block; Content is not sent when blocks are present."

### 3. Known limitation, acknowledged — token budget blind to images

`segmented_memory.go` counts tokens on `msg.Content` only, and images ride in L1 history and are re-sent on every subsequent provider call in the session. A few MB of base64 grows every request while the memory manager sees a ~10-token message. The PR acknowledges this — fine to defer, but it belongs in the same tracking issue as #1.

### Minor / non-blocking

- `turns := response.Metadata["turns"].(int)` — the pre-existing unchecked type assertions were carried over verbatim; a panic here now flows through all three entry points (it flowed through both before, so no regression). Optional hardening for a future pass.
- `chatParams.reportProgress`/`reportContentBlocks` existing solely to keep telemetry byte-identical is slightly awkward, but it's well-commented and the right call for a refactor PR — keeping behavior provable was worth two booleans.

## Summary

The feature is well-scoped, the dedup refactor is faithful, and the tests are the best part. Good to merge once the PR description is corrected (item 1) and the text-block precedence contract is either fixed defensively or explicitly documented (item 2).

Josh-Schoen:
Both pre-merge items addressed in `75c884a` (and the PR description is corrected):

**1. PR-description inaccuracy — fixed.** You're right, and I verified it: neither `pkg/agent/session_store.go` (SQLite) nor `pkg/storage/postgres/session_store.go` serializes `ContentBlocks` — blocks are silently dropped on persistence, not stored. The Additional Context section now describes the real gap (blocks don't survive session rehydration; the model sees only canonical text mid-conversation after a restart) plus the in-session L1/token-accounting concern you noted in item 3, as one combined follow-up.

**2. Text-block precedence footgun — fixed defensively (your option a).** `chat()` now prepends `{Type: "text", Text: userMessage}` when `contentBlocks` is non-empty but contains no text block (skipped when `userMessage` is empty). No-op for callers that already include their prompt as a text block. The `ChatWithContentBlocks` doc comment now states the contract explicitly. Two tests added at the provider boundary:
- `ImageOnlyPrependsText` — image-only blocks arrive at the provider with the canonical text prepended as the first block.
- `ExistingTextBlockNotDuplicated` — caller-supplied text blocks pass through unchanged (no duplicate prompt).

The guard runs before span setup, so `message.content_blocks` reports the effective count sent to the model.

Full `pkg/agent` suite green with `-tags fts5 -race` on top of the updated branch (main was merged in via GitHub in the meantime; this commit sits on top of that merge).

Tracking issue for the persistence/token-accounting follow-up: deferred for the moment per repo owner — will file separately.

ilsiepotamus (approved):
**Round 3 — Approve.** Everything asked for in round 2 is now in place; this round reviewed what changed since the round-2-verified commit (`d5d7bee`): two merges of main and the footgun fix `75c884a`.

## Footgun fix — verified ✅

`chat()` now prepends `userMessage` as a text block when `contentBlocks` has none (round 2's preferred option (a)), and the implementation gets the edges right:

- Builds a **fresh slice** — the caller's slice is never mutated.
- The `userMessage != ""` guard means an empty prompt never injects an empty text block — which the Anthropic API would reject (verified against the blocks branch in `pkg/llm/anthropic/client.go`, which drops `Content` entirely when blocks are present — exactly the behavior that motivated the fix).
- No-op when the caller already includes a text block; `hasTextBlock`'s `"text"` matches the type string every provider branches on.
- Two new tests cover both sides: `ImageOnlyPrependsText` and `ExistingTextBlockNotDuplicated`.
- Doc comment now states the contract explicitly.

## Round-2 items — closed out

1. **PR body corrected** ✅ — Additional Context now accurately says blocks are *dropped* at persistence rather than bloating storage.
2. **Tracking issue filed** — the body promised one but none existed; filed as #276 (block survival across rehydration + image-aware token accounting).

## Verification on head `bd67abb`

- Full `go test -tags fts5 -race -count=1 ./pkg/agent/` passes locally on the exact head.
- Isolated diff of the PR's three files since round 2 contains only the documented changes (doc comment, `hasTextBlock`, guarded prepend, two tests) — the merge commits are pure main catch-up.
- Branch is current with main; CI re-running on the merge commit.

Good to merge once CI completes.

PR #252, opened 2026-07-12
fix(task): record real conversation session on claims and task creation

Josh-Schoen:
## Summary

`task_board claim` previously synthesized `"<agentID>-session"` as the claiming session because the tool never saw the real conversation id — making `claimed_by_session` useless for session-scoped task filtering downstream. Task creation recorded no session at all.

- **Claims record the real conversation session**: `executeClaim` reads the session id from ctx (`session.SessionIDFromContext` — every `Chat`/`ChatWithProgress`/`ChatWithContentBlocks` entry point already stamps it via `session.WithSessionID`), falling back to the legacy synthetic id for headless usage so standalone behavior is unchanged.
- **Creation stamps attribution metadata**: `task_board create`, `decompose` (new `DecomposeRequest.SessionID`), and the skill task emitter stamp `created_by_session` task metadata (new canonical constant `task.CreatedBySessionMetadataKey`). This is metadata attribution, deliberately **not** a pre-claim: `ClaimTask` requires `claimed_by_session IS NULL` and the ready-front only checks `status = OPEN`, so pre-claiming would make the agent's own ready → claim loop fail.

## Tests

New coverage: ctx-session claim, headless fallback, create-stamps-metadata-and-stays-claimable, emitter template-path stamping, emitter decomposer-path stamping. Full `pkg/agent` / `pkg/task` / `pkg/skills/tasks` suites pass (`-tags fts5`).

## Notes

- Based on `47d9655` (the commit downstream tera-backend pins) rather than current `main` — rebases cleanly onto `main`; happy to rebase if preferred.
- Downstream consumer: Teradata-PE/avmo-tera-cloud#142 (extends its ListTasks session filter to match the new metadata).


ilsiepotamus (approved):
Reviewed and verified locally:

- `Chat`/`ChatWithProgress` stamp the session into ctx before the conversation loop, so the tool's ctx read is sound.
- The new `Metadata` assignments are on freshly constructed structs (no clobbering); the emitter's decomposed path nil-guards before indexed writes.
- Both `DecomposeRequest` constructors in the repo were updated; no gRPC surface maps to this struct, so no proto change needed.
- No drift in the six touched files between base 47d9655 and current main — update-branch/rebase before merge is all that's needed (GitHub reports MERGEABLE/BEHIND).
- Ran `pkg/task`, `pkg/skills/tasks`, and the TaskBoardTool tests in `pkg/agent` at the PR head with `-tags fts5 -race` — all pass. CI fully green.

Non-blocking nits:
1. PR body mentions `ChatWithContentBlocks`, which doesn't exist in this repo — worth correcting the description.
2. The emitter stamps `created_by_session` unconditionally (empty session would write `""`), while the tool/decomposer guard on non-empty — a guard would make the four sites consistent.
3. The decomposed path double-stamps the key (decomposer materialization + post-process `UpdateTask`) — harmless, same value.

PR #255, opened 2026-07-14
feat(observability): add backend-agnostic OTLP tracing support

Vasu2803k:
## Summary

- Adds `OTelTracer` that exports Loom spans to any OTLP HTTP endpoint (Opik, Jaeger, Tempo, Grafana) via a single config change — no code changes required to switch backends
- Follows OTel GenAI semantic conventions v1.34+ throughout: `gen_ai.system`, `gen_ai.operation.name`, `gen_ai.usage.*`, `gen_ai.prompt`, `gen_ai.completion`
- Tested end-to-end with self-hosted Opik + AWS Bedrock (`us.anthropic.claude-sonnet-4-5`)

## What's included

**New packages / files**
- `pkg/observability/otel.go` — `OTelTracer` with `sync.Map` span pairing, `filteringSpanProcessor` to suppress infrastructure noise
- `pkg/observability/otel_attrs.go` — full Loom → `gen_ai.*` attribute translation + `gen_ai.system` normalization (`bedrock`→`aws.bedrock`, `gemini`→`google.generative_ai`, etc.)
- `pkg/observability/otel_config.go` — `OTelConfig` with env-var resolution (`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, `OTEL_SERVICE_NAME`, etc.)
- `pkg/observability/privacy.go` — PII/credential redaction shared between Hawk and OTel tracers
- `pkg/observability/otel_test.go` — 22+ test cases (attribute mapping, normalization, redaction, filtering, span context)
- `examples/config/otlp-opik.yaml` — ready-to-use config for Bedrock + self-hosted Opik
- `docs/architecture/otlp-integration.md` — architecture and configuration reference

**Modified**
- `pkg/llm/instrumented_provider.go` — adds `gen_ai.operation.name="chat"`, full prompt/completion text (`message.preview`, `response.preview`) on LLM spans; no truncation
- `pkg/agent/agent.go` — removes 100-char truncation from `message.preview`/`response.preview`
- `pkg/observability/auto_select.go` — adds `TracerModeOTel`, auto-activates when `otlp_endpoint` is set
- `pkg/observability/hawk.go` — extracts PII patterns to shared `privacy.go`
- `cmd/looms/config.go` — adds `otlp_endpoint`, `otlp_headers`, `otlp_insecure`, `otlp_include_spans`
- `cmd/looms/cmd_serve.go` — wires OTel mode, `service.version`, graceful `Flush`+`Shutdown` on exit
- `go.mod` — adds `go.opentelemetry.io/otel` v1.43.0 SDK dependencies

## Configuration

```yaml
observability:
  enabled: true
  mode: otel                   # or leave out — auto-activates when otlp_endpoint is set
  otlp_endpoint: "http://localhost:5173/api/v1/private/otel/v1/traces"
  otlp_insecure: true          # for local / self-hosted without TLS
  otlp_headers:
    Authorization: "Bearer <key>"   # optional
  otlp_include_spans:          # optional — suppress startup/storage noise
    - "llm."
    - "agent."
    - "tool."
    - "backend."
    - "mcp."
```

Or via environment:
```bash
export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
export OTEL_SERVICE_NAME=my-loom-service
```

## Test plan

- [ ] `go test -tags fts5 -race ./pkg/observability/...` passes
- [ ] `go test -tags fts5 -race ./pkg/llm/...` passes
- [ ] `go test -tags fts5 -race ./pkg/agent/...` passes
- [ ] Start server with `examples/config/otlp-opik.yaml`, send a chat message, verify in Opik:
  - `agent.chat` span has `gen_ai.prompt` (user message) and `gen_ai.completion` (answer)
  - `llm.completion` span has `gen_ai.prompt` (full prompt sent to model) and `gen_ai.completion` (model response)
  - Token counts and cost visible on both spans
  - No startup/migration spans in Opik trace list
- [ ] Stop server, verify no span loss (graceful flush log line appears)

### DIFF
--- a/ENV_VARS.md
+++ b/ENV_VARS.md
@@ -16,6 +16,7 @@ This document describes all environment variables used in the Loom project.
   - [HuggingFace](#huggingface)
 - [AWS Credentials](#aws-credentials)
 - [Observability (Hawk)](#observability-hawk)
+- [Observability (OTLP)](#observability-otlp)
 - [Web Search Tools](#web-search-tools)
 - [Docker Configuration](#docker-configuration)
 - [TLS/ACME Configuration](#tlsacme-configuration)
@@ -554,13 +555,14 @@ looms config set-key hawk_api_key
 
 ### LOOM_TRACER_MODE
 **Default:** `auto`
-**Values:** `auto`, `service`, `embedded`, `none`
-**Used in:** `pkg/builder/builder.go`
+**Values:** `auto`, `service`, `embedded`, `otel`, `none`
+**Used in:** `pkg/builder/builder.go`, `pkg/observability/auto_select.go`
 
 Controls which tracer implementation to use:
-- `auto`: Automatically select based on environment
+- `auto`: Automatically select based on environment (OTLP endpoint takes priority, then Hawk service, then embedded)
 - `service`: Use Hawk service (gRPC)
 - `embedded`: Use embedded storage (memory or SQLite)
+- `otel`: Export spans via OTLP HTTP to any compatible backend (Jaeger, Grafana Tempo, Opik, Honeycomb, Datadog, etc.)
 - `none`: Disable tracing
 
 ```bash
@@ -597,6 +599,116 @@ Path to SQLite database for embedded tracer.
 
 ```bash
 export LOOM_EMBEDDED_SQLITE_PATH=/path/to/traces.db
+```
+
+---
+
+## Observability (OTLP)
+
+Loom supports exporting traces to any **OpenTelemetry Protocol (OTLP) HTTP**-compatible backend (Jaeger, Grafana Tempo, Opik, Honeycomb, Datadog, OTLP Collector, etc.).
+
+Set `LOOM_TRACER_MODE=otel` (or leave `auto` and set the endpoint — OTLP takes priority in `auto` mode).
+
+### OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+**Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
+
+Full OTLP HTTP URL including path. Standard OTel env var — checked first before `LOOM_OTLP_ENDPOINT`.
+
+```bash
+# Jaeger
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
+# Grafana Tempo
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=https://tempo.example.com/otlp/v1/traces
+# Opik (local)
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:5173/api/v1/private/otel/v1/traces
+```
+
+### LOOM_OTLP_ENDPOINT
+**Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
+
+Loom-specific fallback for the OTLP HTTP endpoint. Used when `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is not set.
+
+```bash
+export LOOM_OTLP_ENDPOINT=http://localhost:4318/v1/traces
+```
+
+### OTEL_EXPORTER_OTLP_TRACES_HEADERS
+**Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
+
+HTTP headers sent with every OTLP export request. Standard OTel env var — checked first before `LOOM_OTLP_HEADERS`.
+
+**Format:** Comma-separated `key=value` pairs.
+
+```bash
+# Bearer token auth (Honeycomb, Grafana Cloud, etc.)
+export OTEL_EXPORTER_OTLP_TRACES_HEADERS="Authorization=Bearer my-api-key"
+# Multiple headers
+export OTEL_EXPORTER_OTLP_TRACES_HEADERS="Authorization=Bearer my-key,x-project-id=my-project"
+```
+
+### LOOM_OTLP_HEADERS
+**Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
+
+Loom-specific fallback for OTLP headers. Used when `OTEL_EXPORTER_OTLP_TRACES_HEADERS` is not set. Same `key=value,key2=value2` format.
+
+```bash
+export LOOM_OTLP_HEADERS="Authorization=Bearer my-api-key"
+```
+
+### LOOM_OTLP_INSECURE
+**Default:** `false`
+**Values:** `true`, `false`
+**Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
+
+Disables TLS certificate verification for the OTLP HTTP connection. Use only for local development against self-signed or plain HTTP endpoints.
+
+```bash
+export LOOM_OTLP_INSECURE=true
+```
+
+⚠️ **Warning:** Never set this in production — it disables certificate validation entirely.
+
+### OTEL_SERVICE_NAME
+**Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
+
+Sets the `service.name` resource attribute on all exported spans. Standard OTel env var.
+
+```bash
+export OTEL_SERVICE_NAME=my-loom-agent
+```
+
+### OTEL_SERVICE_VERSION
+**Used in:** `pkg/observability/otel_config.go`
+
+Sets the `service.version` resource attribute on all exported spans. Standard OTel env var.
+
+```bash
+export OTEL_SERVICE_VERSION=1.3.0
+```
+
+### Quick-Start Examples
+
+**Jaeger (local, no auth):**
+```bash
+export LOOM_TRACER_MODE=otel
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
+export LOOM_OTLP_INSECURE=true
+export OTEL_SERVICE_NAME=my-agent
+```
+
+**Grafana Cloud (with bearer token):**
+```bash
+export LOOM_TRACER_MODE=otel
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=https://otlp-gateway-prod-us-central-0.grafana.net/otlp/v1/traces
+export OTEL_EXPORTER_OTLP_TRACES_HEADERS="Authorization=Bearer <instance-id>:<api-key>"
+export OTEL_SERVICE_NAME=my-agent
+```
+
+**Auto-select OTLP (no explicit mode needed):**
+```bash
+# Setting the endpoint is enough — auto mode selects OTLP first
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
+export OTEL_SERVICE_NAME=my-agent
 ```
 
 ---
--- a/ENV_VARS.md
+++ b/ENV_VARS.md
@@ -660,13 +660,15 @@ export LOOM_OTLP_HEADERS="Authorization=Bearer my-api-key"
 **Values:** `true`, `false`
 **Used in:** `pkg/observability/otel_config.go`, `pkg/observability/auto_select.go`
 
-Disables TLS certificate verification for the OTLP HTTP connection. Use only for local development against self-signed or plain HTTP endpoints.
+Forces plain HTTP transport for the OTLP exporter (uses `http://` even if the endpoint URL specifies `https://`). This is implemented via `otlptracehttp.WithInsecure()` which **forces plaintext** — it does **not** disable TLS certificate verification for existing `https://` connections. Use only for local development against collectors that listen on plain HTTP (e.g. Jaeger or Opik on localhost).
+
+For `https://` endpoints with self-signed certificates, configure the collector to use a trusted certificate instead.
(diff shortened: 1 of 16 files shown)


InfinityOverflow (changes requested):
@Vasu2803k please fix the failing CI checks and test coverage

ilsiepotamus (changes requested):
Thanks for this — the overall shape is good: clean `Tracer` interface compliance, graceful flush/shutdown, no-op fallback consistent with existing patterns, and proper `go.mod` hygiene. I built and ran the PR branch: `go build -tags fts5 ./...`, `gofmt`, and `go test -tags fts5 -race` on `pkg/observability`, `pkg/llm`, and `pkg/agent` all pass.

However, there are **two blocking correctness bugs**, one of which breaks the PR's central feature.

---

## 🔴 Blocking

### 1. Parent-child trace linkage is broken — every child span lands in a different trace than its parent (verified empirically)

In `StartSpan` (`pkg/observability/otel.go`), when a Loom parent exists, the code injects a **UUID-derived** remote span context via `oteltrace.ContextWithRemoteSpanContext`. But the parent's actual exported OTel span has **SDK-generated random IDs** (from `t.tracer.Start`), not UUID-derived ones. So the child exports into the UUID-derived trace, referencing a parent span ID that is never exported, while the parent exports into a different, SDK-generated trace.

Verified with a `tracetest.InMemoryExporter` test against this branch:

```
parent: trace=c4656c22110e85cc07d97a565196f02a span=bb0ebc001f2c2c7f
child : trace=0004dd80709f4fa5b618d9dec7310b76 parentSpan=29115359c07d49c9  ← orphaned, different trace
```

The context returned by `StartSpan` already carries the real OTel parent span — if the injection were simply removed, SDK-native propagation would link everything correctly. The injection actively breaks it. This also falsifies the invariants stated in `docs/architecture/otlp-integration.md` ("OTel span.TraceID = Loom span.TraceID", "OTel span.ParentSpanID = Loom span.ParentID").

**Fix**: look up the parent's live OTel span from `activeSpans` (`t.activeSpans.Load(parent.SpanID)`) and use `oteltrace.ContextWithSpan`; fall back to the UUID-derived remote context only when no live OTel parent exists (cross-process case). Please also add a `tracetest`-based export assertion test — the current tests only check the Loom-side span fields, which is why this wasn't caught. (Opik testing wouldn't catch it either: spans still *appear*, just as disconnected traces.)

### 2. `mode` derivation in `cmd_serve.go` was not updated — validation and runtime now disagree

`Config.Validate()` (`cmd/looms/config.go`) was changed to derive `mode = "otel"` when `otlp_endpoint` is set — explicitly to "keep Validate() in step with what the server actually runs." But the server's own derivation at `cmd/looms/cmd_serve.go:976-984` still only knows Hawk/embedded. Result: `enabled: true` + `otlp_endpoint: ...` **without** explicit `mode: otel` validates as otel but silently runs **embedded** (or Hawk service). The PR description's "auto-activates when otlp_endpoint is set" only holds for the `NewAutoSelectTracer` / `pkg/builder` path, not `looms serve`.

**Fix**: add `if config.Observability.OTLPEndpoint != "" { mode = "otel" }` first in the cmd_serve derivation.

---

## 🟡 Should fix

- **Inconsistent privacy defaults.** `cmd_serve.go` hardcodes `RedactCredentials/RedactPII: true` for otel mode, but the auto-select and env paths (`LOOM_TRACER_MODE=otel`, or auto + endpoint via `pkg/builder`) resolve `safePrivacy(nil)` → **no redaction at all**, while now exporting full prompt/completion text off-box. Redaction should default on in every OTel construction path.
- **Unbounded prompt/response attributes hit ALL tracer backends, not just OTel.** `instrumented_provider.go` now sets full `message.preview`/`response.preview` (no truncation) on llm spans — these flow to Hawk and embedded SQLite too, changing storage growth and privacy posture for existing non-OTel users. It also contradicts the rationale in this same PR's `agent.go` comment ("Capping prevents unbounded trace payload sizes and avoids leaking full conversation content"), where previews are capped at 200 runes. Pick one policy; at minimum apply `sdktrace.WithSpanLimits` / a configurable cap. (The PR description says "no truncation" while `agent.go` adds a 200-rune cap — the description is stale too.)
- **`otlp_insecure` semantics are wrong.** `otlptracehttp.WithInsecure()` forces plain HTTP — it does **not** "disable TLS certificate verification" as `ENV_VARS.md` and the config comments claim. A user with a self-signed **https** endpoint who follows the docs will silently downgrade to unencrypted HTTP (the option is appended after `WithEndpointURL`, so it wins). Skipping cert verification requires `WithTLSClientConfig(&tls.Config{InsecureSkipVerify: true})`. Decide which behavior is intended and make docs match.
- **Redaction is duplicated, not shared** (the PR description claims "shared between Hawk and OTel"). Only the regexes moved to `privacy.go`; `redactOTelSpan` re-implements `HawkTracer.redact` minus the `strings.Contains(key, "token")` check — so `auth_token`-style keys are deleted by Hawk but exported by OTel. Omitting it was arguably deliberate (Hawk's check nukes `llm.tokens.*` count attributes — a pre-existing Hawk quirk), but that deserves one shared function with an explicit exemption, not silent divergence.
- **Standard `OTEL_EXPORTER_OTLP_ENDPOINT` (the most common OTel env var) is not honored** — only the `_TRACES_` variant. Users following generic OTel docs get "otlp_endpoint is required." Ditto `OTEL_EXPORTER_OTLP_TRACES_INSECURE`. Also `parseHeadersEnv` doesn't URL-decode values (the OTel spec uses W3C baggage encoding).
- **`RecordMetric` is a no-op in otel mode** — all the LLM call/latency/token/cost metrics `instrumented_provider.go` emits are dropped. It's a documented decision, but it's a real functional gap vs Hawk mode worth stating in the PR description and docs, not just a code comment.

## 🟢 Minor / nits

- `filteringSpanProcessor` prefix-filtering drops intermediate spans, orphaning their children in the backend — worth documenting. Also `SpanFilter` isn't wired into `AutoSelectConfig`, so filtering is unavailable via env/auto paths.
- Empty `ServiceName` in the auto-select path yields `service.name=""`; default it to `"loom"`.
- `agent.id` span attributes (`pkg/collaboration/debate.go:352`, `pkg/patterns/orchestrator.go:461`) aren't mapped (only `agent_id` is), so they export as `loom.agent.id` while `session_store.go` spans get `gen_ai.agent.id`.
- `activeSpans` entries leak (and OTel spans never end) for spans abandoned on panic/error paths.
- Doc/config drift: `otlp-integration.md` still marks `otel_attrs.go` as "📋 to be created"; the example config's usage comment references `looms-opik-test.yaml` but the file is `otlp-opik.yaml`; the suggested `LOOM_OBSERVABILITY_OTLP_HEADERS` env override likely doesn't work (viper `AutomaticEnv` can't decode a string into `map[string]string`); `OTEL_SERVICE_VERSION` is described as a "standard OTel env var" but isn't one (that's `OTEL_RESOURCE_ATTRIBUTES`).
- `gen_ai.prompt`/`gen_ai.completion` are deprecated in recent semconv (replaced by `gen_ai.input.messages`/events); fine for Opik compatibility, but "follows v1.34+" overstates it, and `gen_ai.usage.cost` is not a spec attribute.
- Tests: no assertions on actually-exported spans (an `InMemoryExporter` test would have caught blocking issue #1), no tests for `filteringSpanProcessor` or `parseHeadersEnv`, `TestLoomToGenAIMappings` just re-states the map, and plain `t.Error` is used where the project prefers testify.
- `firstEnv("OTEL_SERVICE_NAME", "")` — the empty second key is odd; plain `os.Getenv` suffices.

## What's done well

- Loom span/context propagation faithfully mirrors `NoOpTracer` (trace-ID override, resource-attribute inheritance), so downstream Loom behavior is preserved.
- Graceful `Flush` + `Shutdown` on server exit, ordered after gRPC stop, with timeout.
- Constructor error → no-op fallback matches the existing Hawk/embedded pattern; config validation additions are consistent.
- `ENV_VARS.md` and the architecture doc are thorough, with no marketing-speak violations.
- All race tests pass; build and gofmt are clean.

Both blocking fixes are small and localized — happy to re-review once they're in.

Vasu2803k:
@ilsiepotamus - Thanks for the comments. Resolved all of it. Please check.

ilsiepotamus (changes requested):
Re-reviewed at f57b78f5. Both blocking bugs are fixed — I verified them empirically (rebuilt the branch, re-ran the linkage check with an `InMemoryExporter`, full `-race` suites on `pkg/observability`, `pkg/llm`, `cmd/looms` all pass). Nice work on the turnaround, and the two regression tests (`TestParentChildSameTrace`, `TestThreeGenerationChain`) are exactly the right shape.

## ✅ Verified fixed

- **Parent-child linkage**: live-parent lookup from `activeSpans` with UUID-derived remote context as cross-process fallback — children now share the parent's SDK trace ID and reference its real span ID.
- **`cmd_serve.go` mode derivation**: `otlp_endpoint` now activates otel mode without explicit `mode: otel`, matching `Validate()`; config tests added.
- **Privacy defaults**: `safeOTelPrivacy` defaults redaction on in every OTel construction path.
- **Preview truncation**: llm spans now capped at 200 runes, consistent with `agent.go`.
- **`otlp_insecure` semantics**: corrected in the config doc comment and `ENV_VARS.md`.
- **`service.name`** defaults to `"loom"`.

## 🔴 Remaining — small, but the first one is a real bug in the new fix

1. **`OTEL_EXPORTER_OTLP_ENDPOINT` is used as-is, but the OTel spec defines it as a *base* URL** — exporters must append `/v1/traces` to it (only the `_TRACES_` variant is used verbatim). As written, the most standard OTel setup (`export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318`) makes the exporter POST to `/`, which 404s on every standard collector — silently, since export errors are async. Fix in `resolveOTelConfig`: when the endpoint comes from the base var (or has an empty/`/` path), append `/v1/traces`.
2. **`NewAutoSelectTracerFromEnv` (`auto_select.go:315`) wasn't updated to match** — it still checks only `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`/`LOOM_OTLP_ENDPOINT`, so auto mode won't select OTel when only the base var is set, even though `resolveOTelConfig` now accepts it. Same `firstEnv` list (plus the path handling from item 1).
3. **The docs commit had no effect.** Commit 6ae6bf0a says "update otlp-integration.md to reflect implemented state," but the file is byte-identical to the previous head — the change was likely lost in the rebase. It still marks `otel.go`/`otel_attrs.go`/`otel_config.go` as "📋 to be created" (lines 211/272/297) and states the invariant `OTel span.ParentSpanID = Loom span.ParentID` (line 264), which after the linkage fix only describes the cross-process fallback, not in-process behavior. Also: `examples/config/otlp-opik.yaml:7` still references `looms-opik-test.yaml` (file is `otlp-opik.yaml`), and the PR description still says "no truncation" — it's 200 runes now. Doc accuracy is a hard rule in this repo (see CLAUDE.md), so these need to land with the PR.

## 🟡 Carried over, won't block

Redaction logic still duplicated between `HawkTracer.redact` and `redactOTelSpan` (with the `token`-substring divergence), `RecordMetric` no-op gap not mentioned in docs, `agent.id` vs `agent_id` mapping inconsistency, `parseHeadersEnv` not URL-decoding values, `activeSpans` leak for abandoned spans. Fine as follow-ups — a tracking issue would be enough.

Items 1–2 are ~15 lines plus a test; item 3 is doc-only. Should be one quick round.

ilsiepotamus (approved):
Re-reviewed at c2368561 (rebased onto main with #265). All three remaining items are addressed — I fetched the branch and re-ran `go test -tags fts5 -race` on `pkg/observability`, `pkg/llm`, and `cmd/looms`: all pass, including the parent-linkage regression tests post-rebase. CI is green.

## ✅ Verified fixed

- **`OTEL_EXPORTER_OTLP_ENDPOINT` semantics**: `resolveOTLPEndpointEnv()` implements the spec correctly — traces-specific var verbatim, base var with `/v1/traces` appended (trailing slash handled), Loom var verbatim — with 6 subtests covering priority and path-append cases.
- **`NewAutoSelectTracerFromEnv`** now routes through the same resolver, so auto mode activates on the base var too.
- **Docs**: "📋 to be created" markers removed, ParentSpanID invariant corrected to describe the activeSpans lookup with cross-process fallback, and the `otlp-opik.yaml` usage comment now references the right filename.

## 🟡 Two cosmetic asks before merge (no new push needed for the second)

1. One invariant line in `docs/architecture/otlp-integration.md` (line 263) is still inaccurate: `OTel span.TraceID = Loom span.TraceID (W3C hex format)` — with the linkage fix, in-process root spans get SDK-generated trace IDs, so this only holds on the cross-process fallback path. Suggested replacement:
   `OTel span.TraceID: SDK-generated; equals hex(Loom TraceID) only on the cross-process fallback path`
2. The PR description still says "full prompt/completion text … no truncation" and "removes 100-char truncation" — the code now truncates previews at 200 runes. Please edit the description to match.

Approving — the implementation is in good shape and the turnaround across rounds has been fast and precise. The carried-over minors from the first review (shared redaction helper, `RecordMetric` gap in docs, `agent.id` mapping, header URL-decoding, activeSpans leak) are fine as a follow-up issue.

InfinityOverflow (approved):

PR #267, opened 2026-07-21
fix(storage): sanitize SQL result table names at every use

ilsiepotamus:
## Summary

Triage of an ArmorCode/Snyk SQL injection finding on `pkg/storage/sql_result_store.go:434` (tainted `table_name` from `sql_result_metadata` flowing into `Query`). The flagged sink was already sanitized via `sanitizeIdentifier()`, but the same tainted value reached other sinks in the file **unsanitized**:

- `CREATE TABLE` in `Store`
- `DROP TABLE` in `Store`'s error cleanup path
- `DROP TABLE` in `Delete` and `cleanupExpired` (both read `table_name` back from the metadata table — the exact taint flow Snyk tracks)

## Changes

- **`Store`**: sanitize the table name once at creation, so `CREATE TABLE`, `INSERT`, the error-path `DROP TABLE`, and the value persisted in `sql_result_metadata` are all guaranteed-safe identifiers. Removed the now-redundant second sanitization in the insert path.
- **`GetMetadata`**: re-sanitize `meta.TableName` after reading it back (defense-in-depth against tampered metadata rows). This also covers the builtin `query_tool_result` tool (`pkg/agent/builtin_tools.go`), which interpolates `meta.TableName` into SQL.
- **`Delete` / `cleanupExpired`**: sanitize the table name read back from metadata before `DROP TABLE`.

## Bug fixed

MCP result IDs embed the server name (`mcp_<serverName>_<ts>`). A hyphenated MCP server name (e.g. `my-server`) previously produced an invalid unquoted SQLite identifier, so `CREATE TABLE` failed and result storage broke for that server. Sanitizing at creation fixes this.

## Codebase sweep

Audited every `fmt.Sprintf`-built SQL statement in the repo for the same pattern:

| Location | Verdict |
|---|---|
| `pkg/storage/postgres/result_store.go` | Already safe — `pgx.Identifier{}.Sanitize()` + regex validation |
| `pkg/storage/postgres/task_store.go`, `pkg/storage/sqlite/task_store.go` | Safe — only `$N`/`?` placeholders interpolated |
| `pkg/storage/{sqlite,postgres}/graph_memory_store.go` | Safe — placeholder lists only |
| `pkg/agent/session_store.go` | Safe — identifiers regex-validated |
| `pkg/agent/builtin_tools.go` | Was exposed to tainted `meta.TableName`; now covered via `GetMetadata` sanitization |
| `pkg/shuttle/builtin/document_parse_enhanced.go` | Not a sink — generates DDL as text output, never executes it |

The companion XSS finding on `pkg/server/http.go:481` was triaged as a false positive: the URL-derived app name is allowlist-validated, used only as a registry lookup key, and never echoed into the response.

## Tests

New `pkg/storage/sql_result_store_test.go`:
- Hyphenated MCP server name regression (store → metadata → query → delete round-trip)
- Crafted ID with SQL metacharacters (metadata table survives)
- Tampered `table_name` metadata row (no injection through `GetMetadata`/`Delete`)
- Table-driven `sanitizeIdentifier` cases

Verified: `gofmt` clean, `golangci-lint` 0 issues, `go test -tags fts5 -race` passes for `pkg/storage/...`, `pkg/agent/...`, `pkg/mcp/...`, `pkg/shuttle/...`.

### DIFF
--- a/pkg/storage/sql_result_store.go
+++ b/pkg/storage/sql_result_store.go
@@ -212,8 +212,11 @@ func (s *SQLResultStore) Store(_ context.Context, id string, data interface{}) (
 		return nil, fmt.Errorf("no columns found in SQL result")
 	}
 
-	// Create table name
-	tableName := fmt.Sprintf("tool_result_%s", id)
+	// Create table name. Sanitize once here so every downstream use (CREATE,
+	// INSERT, DROP, and the value stored in sql_result_metadata) is a safe SQL
+	// identifier, even when the id embeds external input such as an MCP server
+	// name containing hyphens.
+	tableName := sanitizeIdentifier(fmt.Sprintf("tool_result_%s", id))
 
 	// Create table (REGULAR table, not TEMP - critical fix!)
 	columnDefs := make([]string, len(columns))
@@ -223,7 +226,7 @@ func (s *SQLResultStore) Store(_ context.Context, id string, data interface{}) (
 		columnDefs[i] = fmt.Sprintf("%s TEXT", safeName)
 	}
 
-	createSQL := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s (%s)", tableName, strings.Join(columnDefs, ", "))
+	createSQL := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s (%s)", tableName, strings.Join(columnDefs, ", ")) // #nosec G201 -- tableName and columns sanitized via sanitizeIdentifier
 	if _, err := s.db.Exec(createSQL); err != nil {
 		return nil, fmt.Errorf("failed to create result table: %w", err)
 	}
@@ -235,8 +238,7 @@ func (s *SQLResultStore) Store(_ context.Context, id string, data interface{}) (
 			placeholders[i] = "?"
 		}
 
-		safeTable := sanitizeIdentifier(tableName)
-		insertSQL := fmt.Sprintf("INSERT INTO %s VALUES (%s)", safeTable, strings.Join(placeholders, ", ")) // #nosec G201 -- tableName sanitized
+		insertSQL := fmt.Sprintf("INSERT INTO %s VALUES (%s)", tableName, strings.Join(placeholders, ", ")) // #nosec G201 -- tableName sanitized via sanitizeIdentifier
 		stmt, err := s.db.Prepare(insertSQL)
 		if err != nil {
 			return nil, fmt.Errorf("failed to prepare insert: %w", err)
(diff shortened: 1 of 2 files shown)


bsahitya (approved):
# Code Review — PR #267

**Title:** fix(storage): sanitize SQL result table names at every use  
**Branch:** `fix/sql-result-store-identifier-sanitization` → `main`  
**Files:** `pkg/storage/sql_result_store.go` (+19/-11), `pkg/storage/sql_result_store_test.go` (+178)

---

## Verdict: ✅ Approve

Clean, well-scoped security fix. The core insight is right: sanitize the table name once at creation so the physical table, the `INSERT`, the error-path `DROP`, and the value persisted to `sql_result_metadata` are all the same guaranteed-safe identifier, then re-sanitize on read as defense-in-depth. Tests build and pass under `-race`.

---

## What's correct

- **Store (`sql_result_store.go:219`)** — sanitizing at creation is the right layer. Because `sanitizeIdentifier` is idempotent, removing the redundant second call in the insert path (old `safeTable := sanitizeIdentifier(tableName)`) is safe, and the `CREATE` / `INSERT` / `DROP` trio now all agree on the same name.

- **GetMetadata (line 405)** — re-sanitizing `meta.TableName` on read genuinely closes the consumer path:
  - `pkg/agent/builtin_tools.go:577-599` interpolates `meta.TableName` into `FROM %s` and pagination SQL.
  - `generatePreview` (line 408) does too.

  Both now receive a sanitized value.

- **Delete / cleanupExpired** — sanitizing the value read back before `DROP TABLE` is consistent with the defense-in-depth story.

- **The real bug fix is legitimate** — `pkg/mcp/adapter/shuttle.go:344` builds IDs as `mcp_<serverName>_<ts>`, and a hyphenated server name (`my-server`) produced an invalid unquoted SQLite identifier, causing `CREATE TABLE` to fail. The regression test covers the full round-trip.

- **Tests are thorough** — hyphen regression, crafted-metacharacter ID, tampered metadata row, and table-driven `sanitizeIdentifier` cases. The `#nosec G201` annotations are accurate and point at the sanitization.

---

## Worth noting (low severity, not blocking)

### Identifier collision between distinct IDs

Metadata is keyed by the raw `id` (PK), but the physical table is keyed by the sanitized name. Two different IDs that sanitize to the same identifier, for example:

- `mcp_my-server_<ts>`
- `mcp_my_server_<ts>`

would map to one physical table via `CREATE TABLE IF NOT EXISTS`, so the second `Store` appends its rows into the first's table and a subsequent `Query` returns the union.

In practice this is effectively unreachable:

- `GenerateID()` is random.
- MCP IDs carry `time.Now().UnixNano()`.

A collision would require identical timestamps and names differing only by hyphen-versus-underscore.

Not something to block on. If you want it airtight, derive the table name from a hash of the raw ID (or store a 1:1 sanitized-name ↔ id mapping). Fine to leave as-is with a comment.

PR #268, opened 2026-07-21
ci: add Snyk dependency and Snyk Code scanning job

ilsiepotamus:
## Summary

Adds a `snyk` job to `ci.yml` (gated into `ci-success` like the other jobs):

- **Dependency scan** (`snyk test`) — merge gate; fails on high/critical vulnerabilities in Go dependencies. Runs after `buf generate` + `generate-weaver` because Snyk resolves the Go dependency graph via `go list`, which needs the module to compile.
- **Snyk Code scan** (`snyk code test`) — advisory (`continue-on-error: true`) while the existing findings are triaged in the Snyk UI; remove that line to make it a merge gate. Pushes to `main` report results to the existing Snyk project (`--report --project-id=0da40ba3-…`) so the UI tracks main, never transient PR states.
- **SARIF upload** to GitHub code scanning under stable categories `snyk-deps` and `snyk-code`, mirroring the existing gosec pattern.
- **`snyk monitor`** on main pushes refreshes the Snyk UI dependency snapshot, even when the scan found vulnerabilities.

## Token handling

The job checks for `SNYK_TOKEN` up front and skips the scan (with a `::notice::`) when absent. This covers two cases:

1. Fork PRs (e.g. external contributions) never receive repository secrets — CI stays green for them.
2. The job is a safe no-op until the secret is configured.

## Required follow-up before this does anything

```bash
gh secret set SNYK_TOKEN --repo teradata-labs/loom
```

Use a Snyk **service-account token** from the org that owns the project (Org Settings → Service accounts) rather than a personal token, so scans land in the right org without an `--org` flag.

## Verification

- `actionlint` v1.7.12: clean
- YAML parse: clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,7 +409,104 @@ jobs:
           category: gosec
 
   # ============================================================================
-  # Job 7: Coverage Check
+  # Job 7: Snyk Security Scan (dependencies + Snyk Code)
+  # Skips gracefully when SNYK_TOKEN is unavailable: secrets are never exposed
+  # to pull requests from forks, and the scan is a no-op until the SNYK_TOKEN
+  # repository secret is configured.
+  # ============================================================================
+  snyk:
+    name: Snyk Security Scan
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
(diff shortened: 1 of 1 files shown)


Vasu2803k (approved):
LGTM!

PR #269, opened 2026-07-21
feat(orchestration): HITL approval gates with durable checkpoint/resume

ilsiepotamus:
## Summary

Declarative human-in-the-loop **approval gates** for pipeline/iterative workflows, with **durable checkpoint/resume** — the loom half of the Tera Cloud workflow HITL design (cloud side lands separately in tera-backend).

- `PipelineStage.hitl_gate` (proto): fires deterministically after the stage's output validates, before the next stage. No LLM decides whether to ask — this is control flow, not a tool call.
- With no inline handler the run **suspends** into a versioned `WorkflowCheckpoint` proto and returns `*WorkflowSuspended` (detect via `errors.As`). No blocked goroutine; any process holding the checkpoint resumes via `Orchestrator.ResumeWorkflow(ctx, pattern, checkpoint, decision)`.
- Decisions: **APPROVE** (continue), **REVISE** (jump back to `revise_target_stage_id` with feedback threaded into the prompt via `{{revision_feedback}}`, bounded by `max_revisions`), **REJECT** (`*GateRejected`, nothing executes).
- **Fingerprint safety**: checkpoints record a SHA-256 of the deterministically-marshaled pattern; resume refuses an edited definition — a workflow change can never bypass a pending review.
- Checkpoints capture inter-stage state only (gates fire at stage boundaries): stage outputs, cost history, iteration/revision counters, pending gate, and workflow-namespace SharedMemory (restored on resume).
- Bonus fix: `publish_restart_request` reason/parameters now reach the restarted stage's prompt (previously logged only). Task-tracked execution leaves boards open on suspension instead of failing tasks.
- CLI: `looms workflow run` prompts inline on a TTY or writes a checkpoint (`--suspend-to`); new `looms workflow resume <wf> <ckpt> --approve|--revise "…"|--reject`.

## Testing

- `go test -race ./pkg/orchestration/` — full suite green, including 15 new gate tests (suspend/resume round-trip, revise feedback threading + re-suspension, fingerprint mismatch refusal, max_revisions, last-stage gate, iterative + restart-disabled fallback round-trips, inline handler paths, YAML parse/validation).
- `golangci-lint run` on changed packages — 0 issues. `buf lint` + `buf generate` clean.
- Example validated end-to-end with the real CLI: `looms workflow validate examples/reference/workflows/orchestration-patterns/hitl-gated-pipeline.yaml`.

Docs: `docs/guides/human-in-the-loop.md` (new "Workflow HITL Gates" section) + runnable example YAML.

### DIFF
--- a/cmd/looms/cmd_workflow.go
+++ b/cmd/looms/cmd_workflow.go
@@ -14,7 +14,9 @@
 package main
 
 import (
+	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -40,14 +42,19 @@ import (
 	"github.com/teradata-labs/loom/pkg/shuttle/builtin"
 	"github.com/teradata-labs/loom/pkg/visualization"
 	"go.uber.org/zap"
+	"google.golang.org/protobuf/proto"
 	"gopkg.in/yaml.v3"
 )
 
 var (
-	workflowDir     string
-	workflowAgents  []string
-	workflowDryRun  bool
-	workflowTimeout int
+	workflowDir       string
+	workflowAgents    []string
+	workflowDryRun    bool
+	workflowTimeout   int
+	workflowSuspendTo string
+	resumeApprove     bool
+	resumeReject      bool
+	resumeRevise      string
 )
 
 // workflowCmd represents the workflow command
@@ -126,6 +133,32 @@ Examples:
 	Run:  runWorkflow,
 }
 
+// resumeCmd resumes a workflow suspended at a HITL gate
+var resumeCmd = &cobra.Command{
+	Use:   "resume <workflow-file> <checkpoint-file>",
+	Short: "Resume a workflow suspended at a HITL gate",
+	Long: `Resume a workflow that suspended at a human-in-the-loop gate.
+
+When a stage declares a hitl_gate and no interactive prompt is available,
+'looms workflow run' writes a durable checkpoint file and exits. Review the
+gated output, then resume with exactly one decision flag:
+
+  --approve            continue to the next stage
+  --revise "feedback"  restart the gate's revise target stage with feedback
+  --reject             end the workflow without executing later stages
+
+The workflow file must be IDENTICAL to the one the run started with — the
+checkpoint records a fingerprint of the definition and refuses to resume a
+changed workflow.
+
+Examples:
+  looms workflow resume create-table.yaml create-table.checkpoint.pb --approve
+  looms workflow resume create-table.yaml create-table.checkpoint.pb --revise "Use MULTISET and add COMPRESS"
+  looms workflow resume create-table.yaml create-table.checkpoint.pb --reject`,
+	Args: cobra.ExactArgs(2),
+	Run:  runResume,
+}
+
 // listCmd lists available workflow files
 var listCmd = &cobra.Command{
 	Use:   "list [directory]",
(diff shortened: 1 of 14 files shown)


bsahitya (approved):
# Review: PR #269 — HITL approval gates with durable checkpoint/resume

**Verdict:** Solid, well-structured work. No blocking bugs found.

The design is clean: gates fire deterministically at stage boundaries, suspension is durable via proto checkpointing, and the SHA-256 fingerprint guard successfully prevents an edited workflow from bypassing a pending review. Live-handler and durable-suspension paths share consistent revise/reject/approve semantics, and the `max_revisions` counter is correctly propagated through the checkpoint (`GateRevisionCounts`), ensuring revision limits remain enforced across suspend→resume cycles.

## Items Verified

- Fingerprint round-trip is consistent across all four paths (pipeline, iterative-enabled, iterative restart-disabled fallback, and the iterative→pipeline `fingerprintOverride`). Resume correctly refuses execution on mismatch. ✅
- Corrupt-checkpoint guards (`next_stage_index` vs snapshots vs stages) exist in both executors. ✅
- Cost accounting is append-only across revises with no double-counting on `APPROVE` resume. ✅
- Memory snapshot/restore correctly excludes `stage-N-output` keys from the generic snapshot (via regex) and re-seeds them from index-aligned snapshots, avoiding duplicate storage. ✅
- Last-stage gate `APPROVE` returns the gated output without executing beyond the final stage. ✅
- Task-tracked executor correctly treats `*WorkflowSuspended` as non-terminal via `errors.As()` and leaves the board open. ✅

## Non-Blocking Observations

### 1. `on_timeout` / `timeout_seconds` are ineffective in durable-suspension mode

In `evaluateHITLGate()` (`hitl_gate.go:148-151`), a nil handler immediately triggers suspension. As a result, `on_timeout` (`approve`/`reject`) is only honored when using inline handlers.

Since non-TTY/backend executions typically use a nil handler, a workflow author who sets:

```yaml
on_timeout: approve
```

does not receive an automatic decision. The workflow simply suspends.

This appears to be an intentional design choice rather than a defect. The documentation (`docs/guides/human-in-the-loop`) states that backend services are responsible for enforcing expiry behavior in suspend mode. It is worth confirming that the tera-backend implementation actually enforces expiration and timeout actions; otherwise, timeouts will never be applied.

### 2. Terminal timeout does not interrupt blocking stdin reads

`terminalGateHandler.RequestDecision()` (`cmd_workflow.go`) uses:

```go
reader.ReadString('\n')
```

which blocks on stdin and does not unblock when the `context.WithTimeout()` created from `gate.TimeoutSeconds` expires.

As a result, in an interactive TTY session the configured gate timeout is effectively ignored. This may be acceptable given that a human is actively present at the terminal, but the silent mismatch between configured timeout behavior and actual runtime behavior could be surprising to users.

### 3. Iterative `REVISE` resume replays superseded snapshots into structured context

In `executeWithRestarts()` (`iterative_pipeline_executor.go:265-273`), the resume replay loop restores all `resume.snapshots` from `0..gatedIndex` into structured context before execution resumes.

For a `REVISE` decision targeting:

```text
target < gatedIndex
```

stages in the range:

```text
target..gatedIndex
```

receive structured-context entries generated from snapshots that are about to be invalidated and recomputed.

The impact is low because:

- Structured-context replay is best-effort.
- Parse failures are non-fatal.
- Re-executed stages eventually overwrite the stale entries.

However, the restored data is logically stale until those stages run again, so the replay behavior is not entirely consistent with the revision semantics.

PR #272, opened 2026-07-22
ci(release): fix Homebrew tap publishing that silently broke after v1.0.1

ilsiepotamus:
## Problem

The Homebrew tap (`teradata-labs/homebrew-tap`) has been stuck at **v1.0.1 since January** while v1.0.2, v1.1.0, v1.2.0, and v1.3.0 all shipped. Two independent breaks:

1. **The trigger can never fire.** `publish-homebrew.yml` (and `publish-winget.yml`) trigger on `release: types: [published]`, but `release.yml` publishes releases with the default `GITHUB_TOKEN` — and GitHub suppresses workflow triggers for events created by `GITHUB_TOKEN` (anti-recursion rule). `publish-homebrew.yml` has had **zero** automatic runs; every run in history was a manual `workflow_dispatch`.

2. **Even manual runs never landed.** The workflow only opens a PR on the tap, and the tap's `default` ruleset requires an approving review that no bot PR ever gets. All six v1.0.2 PRs from January were closed unmerged.

## Fix

- **`release.yml`**: new final `publish-packages` job (after `create-attestations`, so all assets exist) dispatches `publish-homebrew.yml` and `publish-winget.yml` via `workflow_dispatch`, which is exempt from the GITHUB_TOKEN suppression. Skipped for alpha/beta/rc tags.
- **`publish-homebrew.yml`**: after opening the tap PR, merges it with `gh pr merge --squash --delete-branch --admin`, failing the workflow loudly if the merge is rejected. The tap ruleset now has a Repository-admin bypass entry (configured 2026-07-22) that this relies on; `HOMEBREW_TAP_TOKEN` belongs to a tap admin.
- **`docs/reference/release-packaging.md`**: documents the publish chain, the GITHUB_TOKEN trigger-suppression gotcha, token/ruleset requirements, and manual recovery commands.

## Already remediated out-of-band

- Tap updated to v1.3.0 via manual dispatch ([run 29962697840](https://github.com/teradata-labs/loom/actions/runs/29962697840)) + admin merge of [homebrew-tap#8](https://github.com/teradata-labs/homebrew-tap/pull/8). All four formula SHA256s verified against the release's signed `.sha256` assets.
- Tap ruleset `default` (id 11822409) given a `RepositoryRole: admin` bypass actor so automation merges are possible.

## Verification

- YAML validated.
[...]

PR #274, opened 2026-07-23
fix(packaging): repo templates are source of truth for Homebrew formulas

ilsiepotamus:
## Problem

`brew install`/`brew upgrade` of both `loom` and `loom-server` 1.3.0 failed:

```
==> Downloading https://github.com/teradata-labs/loom/archive/refs/tags/v.tar.gz
curl: (56) The requested URL returned error: 404
Error: ... Failed to download resource "loom-server--loom-patterns"
```

Three stacked bugs:

1. **`#{version}` is nil inside Homebrew `resource` blocks** — it resolves to the resource's own (unset) version, not the formula's, so the patterns URL rendered as `v.tar.gz`. PR #272's workflow design assumed interpolation works there (comment cited homebrew-tap#9).
2. **Resource staging strips the tarball's top-level directory** — the `Pathname.glob("loom-*")` in the stage block never matched, so even with the URL fixed, install died with "Could not find Loom source directory".
3. **`Dir.home` during install is the sandboxed build home** (`/private/tmp/loom-server-*/.brew_home`) — the `~/.loom` patterns/config seeding has *never* worked from brew; it wrote into a throwaway directory since 1.0.x.

The tap was hotfixed directly (teradata-labs/homebrew-tap#10, teradata-labs/homebrew-tap#11) — installs work now. This PR fixes the machinery that produced the broken formulas.

## Changes

- **`publish-homebrew.yml`**: copies `packaging/macos/homebrew/*.rb` into the tap each release (single source of truth — hand-edits to the tap no longer survive), then stamps version, binary URLs, resource URL, and all three sha256s. All URLs literal; no `#{version}`. Verify step gains `ruby -c`, placeholder-hash, stale-URL, and interpolation guards.
- **Templates** (`packaging/macos/homebrew/`): synced to the fixed structure — patterns install into `pkgshare` with copy instructions in caveats; config creation delegated to `looms config init` (dogfooding); staging handles both archive layouts.
- **`pkg/versionmgr`**: bumps/extracts `refs/tags/vX.Y.Z.tar.gz` archive URLs so the resource URL stays version-consistent; new table-driven test case.

## Verification

- Sed-pipeline dry-run on the templates reproduces the merged tap formulas **byte-for-byte** (modulo an intentional template comment)
- `brew install teradata-labs/tap/loom-server` and `.../loom` both succeed on arm64 macOS: `looms version` = v1.3.0, 158 pattern YAMLs staged in pkgshare
- `go test -tags fts5 -race ./pkg/versionmgr/` passes; `gofmt -l` clean
- `just version-verify`: 14/14 files in sync

### DIFF
--- a/.github/workflows/publish-homebrew.yml
+++ b/.github/workflows/publish-homebrew.yml
@@ -55,6 +55,7 @@ jobs:
             "https://github.com/${{ github.repository }}/releases/download/v${VERSION}/loom-darwin-arm64.tar.gz"
             "https://github.com/${{ github.repository }}/releases/download/v${VERSION}/looms-darwin-amd64.tar.gz"
             "https://github.com/${{ github.repository }}/releases/download/v${VERSION}/looms-darwin-arm64.tar.gz"
+            "https://github.com/${{ github.repository }}/archive/refs/tags/v${VERSION}.tar.gz"
           )
 
           for url in "${URLS[@]}"; do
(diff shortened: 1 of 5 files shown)


bsahitya (approved):
# PR #274 Review — fix(packaging): repo templates are source of truth for Homebrew formulas

**Verdict:** Solid, well-reasoned fix. Approve with a couple of minor questions.

The root-cause analysis in the description matches what's in the diff, and the fixes are correct. I verified the key claims.

## What I verified

- **Sed pipeline correctness** — ran the exact workflow sed sequence against the real templates with a fake version/hashes. All three sha256 values land in the correct blocks (resource → aaaa, arm64 → bbbb, amd64 → cccc), version + all URLs stamp correctly, ruby -c passes, and the placeholder/#{version}/count guards all pass. ✅
- **looms config init exists** (cmd/looms/cmd_config.go:34) and writes to $LOOM_DATA_DIR/ — so delegating config creation to the binary is real, not aspirational. ✅
- **versionmgr** — go test -tags fts5 -race ./pkg/versionmgr/ passes. The new re3 in both updateHomebrewFormula and extractHomebrewFormula keeps the refs/tags/vX.Y.Z.tar.gz resource URL version-consistent, and because extraction now sees that URL, version-verify will actually catch drift in the resource URL too. ✅

## Strengths

- Correctly identifies the three stacked bugs; the #{version}-is-nil-in-resource-blocks insight is the crux and is handled properly (literal URLs stamped by CI).
- Defensive staging: Pathname.pwd with a loom-* glob fallback handles both archive layouts.
- Verify step is meaningfully hardened (ruby -c, placeholder-hash, stale-URL, interpolation guards) — good regression protection.
- Templates are now genuinely the single source of truth (cp before sed), which is the actual fix for "hand-edits to the tap don't survive."

## Minor points / questions (non-blocking)

1. Patterns installed into both loom (TUI) and loom-server kegs. That's ~158 YAMLs staged twice. The server/weaver consumes patterns — does the TUI client (loom.rb) actually need its own copy, or is it carrying an unnecessary download + resource block? If not needed, dropping the resource from loom.rb would simplify it. If it is intentional (so loom works standalone), fine — just confirm.
2. loom.rb caveats duplicate the copy instructions even though a user installing both formulas gets two pkgshare/patterns. Harmless, just slight redundancy.
3. test do in loom-server.rb:89 still calls looms config list (pre-existing, not introduced here) — worth confirming that subcommand still exists since the config surface was touched conceptually. Not part of this diff.

Nothing here blocks merge. The machinery is correct and the verification guards will prevent this class of breakage from recurring.

PR #275, opened 2026-07-23
feat(release): publish homepage to gh-pages as part of the release process

ilsiepotamus:
## Problem

The GitHub Pages site (https://teradata-labs.github.io/loom/) is a hand-maintained `index.html` on the `gh-pages` branch. Nothing in the release process touches it, so its version strings (`v1.3.0` badge, "Everything below is in v1.3.0") go stale the moment the next release ships — the same split-brain that let the Homebrew tap rot (#274).

## Changes

- **`web/` on `main` is now the single source of truth** for the site (`index.html`, `og.png`, `.nojekyll` imported from `gh-pages`). Hand-edits made directly on `gh-pages` are overwritten at the next release.
- **Version strings are marked**: bumpable versions in `web/index.html` are wrapped in `<span data-loom-version>vX.Y.Z</span>`. The version manager bumps only those spans — historical mentions like "since v1.1.0" are provably untouched (covered by a new table-driven test). `web/index.html` becomes the 15th version-tracked file, so `just version-verify` and the PR version-drift check now cover the homepage.
- **`release.yml` gains a `publish-homepage` job** (after `create-attestations`, skipped for alpha/beta/rc): verifies the tagged tree is stamped with the release version (fails the release if the version manager wasn't run), rsyncs `web/` onto `gh-pages`, and requests a Pages build explicitly via `POST /pages/builds` — GITHUB_TOKEN pushes can be exempt from triggering the managed pages-build-deployment, the same suppression class that silently broke the tap publishers for six months.
- Docs: `docs/reference/release-packaging.md` gains the homepage chain + manual recovery steps; CLAUDE.md's auto-updated files list includes the homepage.

## Verification

- `go test -tags fts5 -race ./pkg/versionmgr/` passes (forced `-count=1`)
- Round-trip: `version-manager set 9.9.9` bumps exactly the two marked spans; `since v1.1.0` untouched; `set` back restores
- `just version-verify`: **15/15 files in sync**
- Publish dry-run against real `gh-pages` content: resulting diff is exactly the two span-markup lines — no deletions, `og.png`/`.nojekyll` byte-identical
- `release.yml` YAML validated


bsahitya (approved):
## Code Review — PR #275: `feat(release): publish homepage to gh-pages as part of the release process`

Overall this is a clean, well-scoped, well-tested change. The design (single source of truth in `web/` on `main`, version-manager-stamped spans, explicit Pages build trigger) is sound and directly addresses the same trigger-suppression class that rotted the Homebrew tap. Tests pass (`TestHomepageHTML`, `TestGetAllTargets`) and the version-verify count is consistent (15/15). Comments below are mostly risk-flags and hardening, not blockers.

### 🔴 Should address

**1. `rsync --delete` will silently remove any gh-pages file not present in `web/`**
In `release.yml`:
```bash
rsync -a --delete --exclude .git web/ /tmp/gh-pages/
```
`--delete` makes `web/` authoritative for the *entire* `gh-pages` tree. Anything currently on `gh-pages` that isn't in `web/` gets wiped on the first release — most importantly a `CNAME` (custom domain) file or any historically published assets. The project currently uses the `github.io/loom` project path (no CNAME), so this is safe *today*, but it's a latent footgun the moment a custom domain is ever configured directly on the branch.
- Suggest either excluding `CNAME` (`--exclude .git --exclude CNAME`) or documenting explicitly that `web/` must contain every file that should exist on `gh-pages`.

### 🟡 Worth considering

**2. No `concurrency` group on `publish-homepage`**
If two release tags land close together (or a release is re-run), two jobs can race on `git push origin gh-pages` and one will fail on non-fast-forward. A `concurrency: { group: gh-pages-publish, cancel-in-progress: false }` would serialize them. Same applies to the manual-recovery path colliding with an in-flight release.

**3. `needs: create-attestations` chains the homepage behind the full binary build**
A homepage-only fix can't publish unless the entire binary/attestation chain succeeds. That's a reasonable default (homepage version must match a real release), and the manual recovery steps in `release-packaging.md` cover the gap — just calling it out as an intentional coupling.

**4. Extract/update regex is strict on the exact tag shape**
```go
regexp.MustCompile(`data-loom-version>v([0-9]+\.[0-9]+\.[0-9]+)<`)
```
This only matches `<span data-loom-version>vX.Y.Z</span>` with no attributes and no whitespace. If someone later writes `<span data-loom-version class="badge">` or `data-loom-version >`, the span is silently skipped — `extractHomepageHTML` could then return zero spans and error, or (worse) miss one span while updating others, producing intra-file drift. Low risk given controlled content, but a comment in `index.html` near the spans warning "do not add attributes to these spans" would help future editors.

### 🟢 Nits / positives

- Good call using `grep -Fq` (fixed-string) for the stamp verification — immune to regex metachars in the version.
- Idempotency handled correctly (`git diff --cached --quiet` short-circuit).
- Map-based dedupe in `extractHomepageHTML` yields nondeterministic slice order, but `verify.go` compares each element against canonical independently, so order is irrelevant — no bug.
- `#nosec` annotation on `os.WriteFile` is justified and consistent with the other `updateXxx` funcs in this file.
- Test coverage is solid: covers the bump-marked-spans-but-leave-historical-mentions case and the no-markers error case. Consider adding one more case asserting behavior when two spans carry *different* versions (drift detection), since that's the exact failure mode the feature is meant to prevent.
- Docs (`release-packaging.md`, `CLAUDE.md`) are accurate and include a manual recovery runbook — nice.

### Verdict
Approve with minor changes. The `rsync --delete` behavior (#1) is the only item I'd want acknowledged/guarded before merge; the rest are optional hardening.

PR #278, opened 2026-07-27
TER-633: Loom context and skill management

mrg-ray:
## What this changes

Skills and patterns move from per-turn injection to on-demand pull, and the context head becomes stable across a session.

- New `manage_skills` builtin (`list` / `load`). A load activates the skill for that session, registers its required tools so the same turn can use them, and delivers the skill body into the conversation as a user-role message alongside a short tool-result confirmation.
- The bound skills' names and descriptions are rendered once into the system prompt at session creation. The per-turn discovery-and-activation push is removed.
- Patterns load on demand through `load_pattern` and return as tool data. The pattern injection channel is removed.
- The findings channel, its cache and its extractor are removed, along with the `record_finding` tool.
- Compaction uses the LLM compressor when a prompt registry is configured, with a prompt that preserves decisions and approvals with their scope, open commitments and reload pointers; the heuristic summariser remains the fallback. The trigger is budget-based — the fixed `maxL1Tokens` trigger is gone, and the value survives only as a stat.
- Large tool results use one 64 KiB threshold at both offload sites, with one exempt set (`manage_skills`, `get_tool_result`, `query_tool_result`). Results render as a verbatim string or as JSON, never Go map syntax.
- Advertised tools are per-session: a session advertises what its own events registered. The error, SQL-result and shared-memory stores are keyed by session, and handle resolution is scoped to the caller's partition.
- A context dump and per-mutation debug logs sit behind one switch, off by default. The dump is written to a local per-run file, never to the tracer or logger; the per-mutation logs are emitted at zap Debug level.
- Token accounting counts what is dispatched to the provider.
- Deleted `fabric.SharedBackendWrapper`, an unreferenced auto-store path carrying its own 100 KiB threshold.

## Behaviour changes worth explicit sign-off

- **Skills no longer auto-activate.** Trigger modes and stickiness in skill YAML are inert; a skill enters the conversation only when the model calls `manage_skills(load)`.
- **`record_finding` and automatic finding extraction are gone.** `EnableFindingExtraction`, `ExtractionCadence` and `MaxFindings` no longer drive behaviour. The proto fields are untouched.
- **`PatternConfig.Enabled` and `WithPatternInjection` no longer drive behaviour.** Patterns arrive only through `load_pattern`.
- **The skill body arrives as a user-role message, by design.** A loaded skill states how the user's request is to be processed, so the model receives it as instruction rather than as tool data. Consequence worth naming: compaction's most-recent-user-message pin can therefore hold the body rather than the preceding user turn.
- **Token accounting no longer counts the raw result record attached to tool messages.** That record is never sent to a provider, and `Content` already holds the rendered form the model receives. Compaction therefore fires later than before at the same real load.
- **`storage.SharedMemoryStore.GetMetadata` takes the caller's session**, matching `Get` — the metadata carries schema, a sample item and a content preview. All in-repo call sites are updated.

## What deliberately does not change

The five memory segments and their roles; the compaction mechanism itself (oldest batch into a cumulative L2 summary, swap eviction at its cap), the recency floor and the tool-pair boundary adjustment; `conversation_memory` promotion; graph memory; the recovery paths; the tool circuit breaker; the MCP skill-activation surface.

## Known bounds

Pre-existing behaviour this change does not touch. Each was diffed against `main` at the base commit (`5b7d75d3`) and is unchanged by this PR; they are listed so the claims above are read with the right scope, not as work deferred.

- **Lazy and dynamic tool registration stay global.** Per-session advertisement covers tools a session's own load or first-need disclosure registered. Tools promoted by `RegisterLazyTools` (the server's UI toolset) or registered on the fly by the executor go into the shared registry, so they reach every session. `evaluateLazyTools` and `tryDynamicRegistration` are byte-identical to `main`.
- **The MCP bridge activates skills without the risk gate.** `manage_skills(load)` returns `approval_required` for a HIGH/RESTRICTED skill and does not activate it. The bridge's `loom_activate_skill` handler has never performed that check and still does not; the gate is new on the agent path, the bridge path is untouched.
- **A second compaction path runs above a fixed threshold.** `enforceTokenBudget` fires each turn above 70% and compresses everything before the last user message in one pass — no recency floor, no zone batches. It is unchanged from `main` and is not part of the budget-driven trigger described above, which governs the admission and replay paths.
- **Store partitioning covers handle resolution** in the three stores. It does not constrain the SQL text a caller passes to `query_tool_result`.

Code this change leaves standing on purpose, rather than sweeping:

- **Kernel result and schema caches** (`AddToolResult`, `CacheSchema`) — no non-test callers, 8 and 13 test references, still wired into token accounting. Removing them means reworking the accounting and deleting a suite, for no behavioural gain.
- **The discovery/router subsystem** — its consumer was the per-turn push this change removes, so `Discover` now has no caller. It keeps ~24 references across `pkg` and `cmd` and its binding resolver still feeds the system-prompt menu, so removal is its own change, not a tidy-up.
- **The skill task emitter** (`EmitForActivation`) — no call path; the design is intact and documented as not wired in `docs/architecture/skill-hygiene.md` and `skills-system.md`.
- **Findings and pattern config fields** — inert but proto-backed; deleting them breaks compatibility, so they are documented as no longer driving behaviour instead.

## How to review this

74 code files and 21 documentation files. The docs carry no new features: they were corrected to describe what the code now does — skills as a pull, no injection channels, the real compaction trigger and threshold — and `manage_skills` / `load_pattern` are documented for the first time.

Questions this will raise, answered here so they need not be asked:

- **Why no `unload`?** Loading appends; a session sheds a skill by ending. Removing a skill mid-session would drop its tool wiring underneath work already in flight.
- **Why no cap on the active set?** The old cap existed to bound injected-body cost, which no longer exists — a load costs one message. `MaxConcurrentSkills` still bounds the search path it always bounded.
- **Why is the skill body a user-role message?** A loaded skill states how the user's request is to be handled, so it belongs in the instruction slot, not the tool-data slot.
- **Why does discovery still exist if nothing calls it?** See Known bounds — its binding resolver still feeds the menu; retiring the rest is a separate change.
- **Why does compaction fire later than before?** Token accounting stopped counting a record that is never sent. The budget now reflects what the provider actually receives.

## Verification

`gofmt` clean, `go build -tags fts5 ./...` clean, `go test -tags fts5 -race ./...` green across the module (92 packages).

Behaviour is asserted from the context dump — what the model actually received — in `pkg/agent`: `TestEpic_EAC3`, `TestEpic_EAC5`, `TestEpic_EAC8`, the skills cutover suite, restore re-fire, offload at both sites, and the store partition tests.

Not verified here, and gated on a live provider: same-turn tool re-projection against a real model, the compressor's fidelity on a real compaction, and a real-model consumer verdict over the dumps.


ilsiepotamus (changes requested):
# Review: architecture agreed, changes requested on the restore path

The shape of this change is right and I want it to land: pull over push is the design the field converged on, and it earns its keep here concretely — stable cacheable head, no per-turn router spend, one message per load, and every context mutation visible in a dump. The TER-633 "no hidden injection channels" principle is executed consistently across skills, patterns, and findings, and the PR description is the most accurate one I've reviewed on this repo: I checked its claims skeptically and they held, including the "byte-identical" statements (verified by extracting and hashing both function bodies at `5b7d75d` vs HEAD — equal for `evaluateLazyTools` and `tryDynamicRegistration`).

Independent verification on this branch: `go build -tags fts5 ./...` clean; `go test -tags fts5 -race -count=1` green for `pkg/agent`, `pkg/storage`, `pkg/shuttle`, `pkg/skills`, and all of `pkg/llm/...`.

The blocking findings below are both in the restore path — exactly where the PR itself says live verification is missing — and they don't indict the design. Fix those two and I'm comfortable merging, with the rest as tracked follow-ups.

## Blocking

**B1 — Restore re-fire runs before `captureBaseTools`, permanently hiding base tools from every other session.**
`captureBaseTools` (`pkg/agent/agent.go:733`) runs only at the top of `runConversationLoop` (`agent.go:2184`), but restore replay runs earlier in the same call chain: `chat()` → `GetOrCreateSessionWithAgent` → `reFireOnRestore` (`pkg/agent/memory.go:287`) → `reFireSkillActivation` → `enforceRequiredSkillTools` → `registerSessionTool` (`agent.go:757`). On a fresh process whose **first** chat targets a restored session — the normal post-restart flow — `a.baseToolNames` is still nil, so every re-fired required tool is marked session-scoped, **including base tools declared in the agent's YAML** (e.g. `execute_query`). `advertisedTools` then hides that tool from every other session of the shared agent for the process lifetime, violating the invariant stated at `agent.go:838-840` ("requiring one never hides it elsewhere"). The restore test doesn't catch this because its fixture skill requires `web_search`, which isn't a base tool in the rig. Suggested fix: snapshot base tools at the end of `NewAgent`, after all construction-time registration, instead of lazily in the loop.

**B2 — Restore replay makes sequential LLM compressor calls while holding the global `Memory.mu` write lock.**
`GetOrCreateSessionWithAgent` holds `m.mu` across `segMem.ReplayMessages(ctx, session.Messages)` (`pkg/agent/memory.go:273`, unlock at `:281`). `ReplayMessages` loops compaction batches, and each batch now routes through the default-on LLM compressor with a 5s timeout (`pkg/agent/segmented_memory.go:814`). Restoring a large over-budget session means dozens of sequential network calls — worst case ~5s each on timeout — while **every other session's `GetOrCreateSession` blocks on the global lock**, and the restoring caller's request deadline burns before their turn starts. This is new behavior: at base, no compressor was ever wired, so replay was heuristic-only. Options: run replay compression outside `m.mu`, use the heuristic on the replay path, or make the compressor async.

## Should-fix

1. **Risk gate is skipped on restore replay.** `reFireSkillActivation` (`agent.go:445-463`) calls `ActivatePinned` with no `IsHighRisk` re-check. A HIGH/RESTRICTED skill loaded under YOLO re-activates after restart even if YOLO is now off or the skill's `risk_level` was raised since. (Blocked loads are correctly not re-fired via the `activated:false` marker — this is only about previously-permitted loads.)
2. **The MCP bridge bypass undermines the new gate.** `handleActivateSkill` (`pkg/mcp/server/bridge_handlers.go:884-912`) activates any skill, any risk, into any caller-supplied `session_id`, and its activations land in the **same** per-session active set the agent loop honors — exclusions get applied and required tools get wired on the next `enforceRequiredSkillTools` pass. The PR documents the bridge as untouched pre-existing surface, which is true, but the gate is new on one path while a side door opens into the same room. Also divergent semantics: bridge uses evicting `ActivateSkill`, agent path uses non-evicting `ActivatePinned`. Related: the gate itself is refuse-only, and its error text coaches the model toward suggesting YOLO mode (`manage_skills_tool.go:162-177`). Now that HITL approval gates exist on main (#269), routing `approval_required` into that flow is the natural completion — happy to help wire it.
3. **The 5s compression timeout defeats the compressor exactly where it matters.** `enforceTokenBudget` → `CompactMemory` compresses everything before the last user message in one LLM call (`segmented_memory.go:1230-1243`) — at the 70% trigger on a 200K window that's ~140K tokens through one 5s-capped call. It will nearly always time out and silently fall back to heuristic stub summaries, defeating the decision-preserving prompt in precisely the big-context sessions it was built for. Suggest chunking `CompactMemory` batches, scaling/configuring the timeout, and emitting a metric that distinguishes LLM compaction from heuristic fallback.
4. **The skill-body sidecar hijacks the compaction pin.** `evictL1Prefix` pins the most recent user message as "the active objective" (`segmented_memory.go:419-425`, the #262 protection). After a load, that's the skill body, and the user's actual question becomes evictable mid-task — #262's failure mode through a new door. Cheap discriminator available: sidecars carry `AgentID: a.id` (`agent.go:2833-2838`), genuine user turns don't. Related divergence: once a later user turn arrives, the body is ordinary compressible history — compaction can silently delete the SOP while the orchestrator still reports the skill active and its tools stay wired (restore re-fires activation but never re-delivers the body).
5. **No size bound anywhere on the skill body path.** `manage_skills` is exempt from the 64 KiB offload by design, and the sidecar append has no threshold. A large skill YAML enters L1 whole, gets pinned (see 4), and if it alone approaches the budget, `enforceTokenBudget` fails the conversation. A cap at load or lint time on `prompt.instructions` size would close this.
6. **Sidecar coalescing covers only Anthropic, Bedrock, and Ollama.** The `appendUserOrCoalesce` treatment (nice design, correct wire shape) wasn't applied to Gemini/Mistral/OpenAI converters. OpenAI tolerates consecutive user messages; Gemini and Mistral have historically rejected them. A skill load on those providers is an untested wire shape — either extend the coalescing or state the support boundary.
7. **SQL-store partitioning is cosmetic against a hostile session — please don't let loom-cloud read it as a tenant boundary.** The partition check gates only handle resolution (`sql_result_store.go:316`); `queryToolResult` then runs caller SQL verbatim against the process-global SQLite file (`sql_result_store.go:334`). A session owning any one handle can `SELECT * FROM sql_result_metadata` to enumerate every session's table names (and session IDs), then read — or `UPDATE`/`DROP`, since SQLite executes writes through `Query()` — any other session's tables. Pre-existing SQL surface, as the PR says, but the partitioning story invites over-trust. Fix direction: authorizer callback / read-only connection + table allowlist derived from the caller's own metadata rows. At minimum a documented warning.
8. **Disk-overflow path bypasses partition enforcement, and eviction strands the owner.** `get()` with `Location == DISK` never consults `enforceSession` (`shared_memory.go:235-252`); session ownership isn't persisted by `evictLRU` or the full-memory overflow branch. Today unreachable from model input (fail-open landmine rather than live exploit), but worse functionally: an LRU-evicted record becomes unreachable **to its own session** — tools build MEMORY refs, the in-memory miss returns "data not found" with no disk fallback, so sessions lose offloaded results under memory pressure.
9. **`SetSharedMemoryThreshold` doesn't propagate to the executor**, so "one threshold at both sites" breaks for any configured agent: `registry.go:889-892` calls it after `NewAgent` already wired the executor default, and the `!= 0` check makes the documented `0 = always reference` unreachable from YAML. (`MultiAgentServer` happens to rescue it later; registry-only embedders drift.)
10. **Nil-library panic in `manage_skills`.** `checkAndRegisterManageSkillsTool` (`agent.go:1368`) registers whenever `skillOrchestrator != nil`, passing `GetLibrary()` unchecked; `load`/`list` then nil-deref. The `load_pattern` registration and `reFireSkillActivation` both guard nil library — this one doesn't.
11. **Small/misconfigured context windows silently lose the primary compaction trigger.** `reservedOutputTokens` defaults to a flat 20000 when 0 (`segmented_memory.go:165-167`, contradicting `SetContextLimits`'s "10% of max" comment). With `maxContextTokens=8000`, `maxAvailable` is negative, `UsagePercentage()` goes negative, and the budget trigger never fires; at base the independent `maxL1Tokens` cap still compacted. Validate `reserved < max` at construction.
12. **`sessionToolLedger` is never cleaned up** — unbounded per-session map growth in long-running multi-session servers.

## Design follow-ups (filed as asks, not merge gates)

- **Pattern reachability**: with the per-turn recommendation channel deleted, the only way the model learns a pattern reference is a loaded skill's `pattern_refs` — a pattern library without skills is now unreachable (`load_pattern` has no list/search action, and its own description says to get references from `manage_skills`). Suggest a `search` action on `load_pattern` reusing the existing keyword machinery, model-driven — pure pull, no injection.
- **User-invoked skill loads**: the design this PR mirrors (Claude Code's Skill tool) has two entry points — model-chosen *and* user-invoked by name. The PR carried over the first and dropped the second: the shipped Teradata skills' `/td-*` slash commands now do nothing. A pre-turn forced load through the same `manage_skills` path is deterministic, dump-auditable, and preserves every invariant.
- **Inert config should say so**: skill `trigger:` blocks (all four shipped Teradata skills carry them), `PatternConfig.Enabled`/`MinConfidence`, `WithPatternInjection`, and proto fields 11-13 (`enable_finding_extraction` etc., whose comments still describe deleted behavior) all validate and silently drive nothing. A loader warning plus proto `deprecated` markers keeps the config surface honest.
- **Skill body authority**: the body is delivered as a bare user-role message with no provenance framing — filesystem YAML content speaks with user authority, strictly above the tool-data channel patterns get. Deliberate Claude-Code mirroring, but worth one framing line ("content of skill X, loaded as reference material") given skills can arrive via import paths.

## Nits

- Empty session IDs silently share one `""` partition in the stores rather than being rejected (`pkg/session/context.go:27-37`), inconsistent with `registerSessionTool`'s fallback to `a.id`.
- `docs/architecture/12-factor-architecture.md:468` still shows a `SharedBackendWrapper` example for the type this PR deletes.
- The skill body is persisted twice (as `Metadata["text_body"]` on the tool message and as the sidecar user message) — stripping it from the tool message halves durable storage per load.
- Restore re-fire keys on `strings.Contains(msg.Content, "get_error_details")` / `"stored in memory"` (`memory.go:405-414`) — a renderer rewording silently drops disclosure tools from restored sessions.
- Boundary inconsistency: params offload strictly-above threshold (`executor.go:517`), results at-or-above (`executor.go:337`).
- `enforceRequiredSkillTools` warn-and-skip on unknown tools while the load confirmation still tells the model "its required tools become available this turn".
- PR body says the debug switch writes "never to the tracer or the logger" — the per-mutation debug logs do go to zap at Debug level (`context_debug.go:16-20`); only the dump goes to the per-run file.
- `NormalBatchSize` is now dead config: `shouldCompress` requires budget > warning, so the normal-batch branch is unreachable.

## Genuinely well done

- The sidecar buffering/drain solution to tool_use↔tool_result adjacency under parallel calls — documented in-code and black-box tested.
- `TokenBudget.Set` fixes a real bug: the old `Reset()+Use()` recorded 0 when over budget, disabling compaction exactly when needed.
- Oracle-resistant partition errors (identical "data not found" for miss vs foreign owner), one enforcing `get()` for both data and describe paths.
- Test discipline throughout: black-box suites drive the real `Chat` pipeline and assert against context dumps rather than internals; the restore scenario builds a genuinely fresh agent over the same durable store; exact-threshold boundary tests at both offload sites; everything race-clean.
- The `%v`-on-map rendering fix, and the compaction prompt itself (decision+scope preservation, anti-hallucination, no role prompting).
- The PR description's honesty. "Known bounds" with diff-verified claims is the standard every large PR here should be held to.

mrg-ray:
Thanks — genuinely useful review, and the verify-first approach (hashing the byte-identical claims especially) is the right bar. Dispositions below.

## Blockers — fixed (`60b69b91`)
- **B1** — `captureBaseTools` now runs at the restore re-fire entry points too, so a restore-first process can't scope base tools to a session. Not snapshotted at end of `NewAgent` as suggested — embedders register tools after the constructor, so that would exclude them and reproduce the bug; the idempotent freeze covers every ordering. Test fails without the fix.
- **B2** — replay moved outside `Memory.mu`; the global lock no longer spans compressor calls.
- Both are point fixes. The root cause is broader: per-session behaviour is simulated over one shared Agent (base-tool snapshot, tool ledger, the global session lock), so this class can recur. Redoing that as a proper agent-definition / per-session-runtime split is out of scope here and will come as a separate PR.

## Should-fix — fixed (`ababe451`)
- **#9** threshold propagation to the executor — test added. **#10** nil-library guard on `manage_skills`. **#11** reserve capped below the window so small windows keep a positive budget — test added. **#12** session ledger dropped on delete.

## Should-fix — pre-existing, verified against base `5b7d75d3`, will be fixed in separate PRs
- **#7** — the SQL passthrough is verbatim at base; this PR only added handle-lookup partitioning. Agreed it must not read as a tenant boundary; a separate PR will address it.
- **#8** — the disk-branch bypass and the evict-to-disk stranding both predate this PR. A separate PR will address it.

## Should-fix — context compaction and restore
- **#3, #4, #5** — context compaction and restore needs a broader fix than this review scopes; these three will be covered in a separate PR dedicated to it, to keep this one manageable.

## Should-fix — skill gate
- **#2 / #1** — agreed the refuse-only gate is incomplete and the bridge is a side door. Since the HITL suspend/resume in #269 is yours, we'd prefer you route `approval_required` into it — happy to hand off the gate points on both the agent and bridge paths.
- **#6** — will be provided as a separate fix, to keep this PR manageable.

## Nits
- Fixed: the 12-factor doc no longer shows the deleted `SharedBackendWrapper`; the description's debug-logging line corrected. The remaining nits ride the follow-up work above.

## Design asks
Acknowledged, planned as follow-ups.

ilsiepotamus (approved):
# Approved — blockers verified fixed

Re-verified at `ababe451`, same bar as the first round.

## Blockers

**B1** — Fixed, and your argument against my suggested fix location is correct: capturing at the end of `NewAgent` would have excluded embedder-registered tools and reproduced the bug for exactly the loom-cloud pattern your test pins. The idempotent `sync.Once` freeze at both re-fire entry points covers every ordering. I verified the "fails without the fix" claim directly: reverted the two `captureBaseTools()` calls, ran `TestRestore_BaseToolStaysVisibleToOtherSessions` — it fails on the right assertion — restored.

**B2** — Fixed. Replay runs from the under-lock snapshot after `Memory.mu` is released, and since `ReplayMessages` holds `sm.mu` for its full duration, concurrent same-session use blocks on the session's own lock rather than observing partial state. One residual for the record, not a gate: the session is published to `m.sessions` before replay, so there is a micro-window between the global unlock and replay acquiring `sm.mu` where a concurrent same-session `AddMessage` could land in L1 ahead of the replayed history, and re-fired tools appear one turn late for a concurrent first caller. Narrow, transient, and strictly better than the global stall — it belongs to the per-session-runtime split you already named as the follow-up.

## Should-fixes in `ababe451`

All four verified: the nil-library guard mirrors `load_pattern`; the budget floor (10% of window, capped at half) closes the negative-denominator trigger loss with a test; the ledger cleanup is complete (`Memory.DeleteSession`/`ClearAll` have no callers outside the Agent methods that now clean it); the threshold re-push reaches the executor with correct `-1`/default mapping, accessor, and test. One nit survives: the registry still gates on `SharedMemoryThresholdBytes != 0`, so `0 = always reference` remains unreachable from YAML — proto3 zero-value semantics; a truly settable zero probably needs an `optional` field. Fine as a follow-up note.

## Deferrals

The pre-existing claims for #7/#8 match what my storage review established independently before your reply asserted them. Deferring #3–5 to a dedicated compaction PR is reasonable scoping. Checking in the TER-633 design docs resolves the rationale gap.

## Gate handoff — accepted

#1/#2 and the gate design ask are mine: **#282** tracks routing `approval_required` into the HITL suspend/resume flow from #269 on both the agent and bridge paths, plus the restore re-check. It slots after the checkpoint/resume branch currently in flight.

## Verification

`gofmt` clean; `go build -tags fts5 ./...` clean; `go test -tags fts5 -race -count=1` green for `pkg/agent`, `pkg/shuttle`, `pkg/storage`, `pkg/skills` at `ababe451`; B1 regression test confirmed to fail with the fix reverted.

PR #288, opened 2026-08-05
Context compilation: compile-per-call context, one threshold, proactive relief

mrg-ray:
# Context Management and Tool-Result Handling — HLD

---

## 1. The design in one statement

**Data structures hold no logic. All logic lives in one algorithm — ContextCompilation —
run at every LLM call.** The core idea is just-in-time context compilation: deciding
everything at the moment of the call gives the best possible leverage for context
precision — for outcome quality and for cost.

- **Arrival** appends. A tool result, a user message, an assistant message, a skill
  body — each is appended to its segment in full natural form. Nothing is examined,
  sized, flagged, or transformed at arrival.
- **Persist** writes each message's durable row, once, by fixed write rules (§4).
  Rows are never updated by the context machinery afterward. One named
  exception, which is not context machinery: cloud finalizes the streamed
  assistant row (content, thinking, fired skills) within its own turn — such
  an update may never touch a row of a completed turn.
- **Tool execution** runs by each tool's own contract (§7) when the model calls it.
- **Compilation** (§5) makes every decision in the system: what renders full, what
  renders as a stub (large-result offload and eviction), when conversation is folded
  into the summary, and what happens at the model's context limit.

The turn is one cycle:

```
user message arrives
└─ TURN LOOP (repeats until final answer):
   1. ARRIVAL   — something is appended (L1; KERNEL too for skill loads)
   2. PERSIST   — its row is written, once
   3. COMPILE + SEND (+ RELIEF)  — the algorithm, §5
   4. model answers, or calls tools → results arrive → back to 1
└─ TURN END — in-memory full payloads are gone; DB rows are all that remains
```

## 2. The segments (data, no logic)

- **KERNEL** — the advertised tool schemas. Rides the provider `tools` parameter.
  Skill loads append schemas here. The schemas' serialized bytes are part of the
  compiled artifact and therefore of the size estimate (§5.1); the dead
  `toolResults`/`schemaCache` fields and their accessors are deleted.
- **ROM** — the system prompt. Byte-stable for the life of the session.
- **L2** — the session summary: **one** cumulative text. Empty until the first fold.
  Persisted as version rows; the newest version is the summary (§5.4).
- **L1** — the message list, in order. In memory during a turn (full natural forms);
  rebuilt from rows on the next turn or on reload.

Three row kinds: **assistant rows** (text plus every tool call's name and complete
arguments in `tool_calls_json` — the Anthropic API rejects a replayed `tool_result`
without its paired `tool_use`, so signatures persist always and replay always),
**tool rows** (result text, linked by `tool_use_id`), **user rows** (user messages
and user-role sidecars such as skill bodies).

## 3. Today's defects (why each rule below exists)

- Results ≥ threshold are replaced at arrival by a summary carrying a per-turn ref;
  the ref dies with the turn but the text replays forever. Observed: models calling
  dead refs, fabricating near-miss refs, and the sandbox silently writing a dead ref
  into a staged file as content — analysis then runs on junk with a success status.
- Retrieval traffic (`query_tool_result` calls and their uncapped results) persists
  and replays — dead worked-examples plus duplicates of paged data.
- Compaction triggers inside `AddMessage`/`ReplayMessages`; removals are recorded
  nowhere; restored sessions diverge from live; the prompt cache is invalidated
  unpredictably; at `maxL2Tokens` the summary is archived and wiped mid-session.
- Old-context retrieval is three disjoint surfaces (`query_tool_result`,
  `conversation_memory` offset guessing, `session_memory` at 3+ sessions).
- The tabular Postgres branch has never worked in cloud and has no delete path.
- Tool schemas — which grow with every skill load — are counted nowhere.

## 4. Write rules (persist time — fixed mechanics, no decisions)

1. **Truncation.** Every tool result row (retrieval results excepted — rule 3
   persists none) is bounded at **16384 bytes stored, tail included**
   (`query_tool_result_threshold_bytes`, reseeded from 0 by migration; the same
   value is the compile-time offload bound and the retrieval page bound). If the
   content exceeds the bound: keep the first `16384 − len(tail)` bytes of
   content (cut backward to a UTF-8 rune boundary) and append the tail, so
   `stored = core + tail ≤ 16384` — a truncated row must never itself exceed
   the threshold, or §5.2's legacy case would stub every large result's row
   from its second turn onward. Tail (normative; token figures are
   `ceil(bytes/3)`):

   ```
   [truncated: %d of %d bytes shown (~%d of ~%d tokens). Re-run the call above if this data is needed again.]
   ```

   An error result is a result that failed — same rule, no separate error store.

2. **Signatures whole.** Calls persist with complete arguments, never truncated —
   the unbounded side is exactly what regenerates the bounded side. "Re-run the
   call above" is literally actionable wherever a result row is visible.

3. **Retrieval pairs are not persisted.** `query_tool_result` entries are filtered
   from `tool_calls_json` at persist; tool rows whose `tool_use_id` matches a
   filtered entry are not persisted. Both sides always — an orphaned pair breaks
   replay at the API. Surgical: other calls in the same assistant message persist
   normally; an assistant message left with no text and no calls is not persisted.
   In memory the pair stays intact for the rest of the producing turn. Why: the
   call's `message_id` argument names in-turn memory; replayed, it is a worked
   example of a pattern that is dead outside its turn, and its result duplicates
   data that has its own row.

4. **Sandbox results carry artifact names.** Harvest recording (S3 upload +
   artifact row) completes synchronously **before** the tool row persists, so the
   artifact id is in the row. Tail (normative; one line per harvested file):

   ```
   harvested:       [harvested: %s → artifact %s (canonical copy, downloadable). Do not regenerate.]
   harvest failed:  [harvest failed for %s: %s — the files were not exported.]
   not harvested:   the truncation tail of rule 1 (when cut), else no tail
   ```

   The pipeline is forward-only: once an outcome is an artifact, the artifact is
   the thing; later turns reference it, never rebuild it.

5. **Columns.** Cloud `session_messages`:

   ```sql
   ALTER TABLE session_messages
       ADD COLUMN IF NOT EXISTS seq     BIGINT  NOT NULL DEFAULT 0,
       ADD COLUMN IF NOT EXISTS evicted BOOLEAN NOT NULL DEFAULT FALSE,
       ADD COLUMN IF NOT EXISTS folded  BOOLEAN NOT NULL DEFAULT FALSE,
       ADD COLUMN IF NOT EXISTS turn    BIGINT  NOT NULL DEFAULT 0;
   CREATE SEQUENCE IF NOT EXISTS session_messages_seq_seq;
   CREATE INDEX IF NOT EXISTS idx_session_messages_session_seq
       ON session_messages(session_id, seq);
   ```

   `seq` is assigned at insert (globally monotonic; reads always carry the session
   filter). **Legacy backfill** (normative): `seq = ROW_NUMBER() OVER (ORDER BY
   created_at, message_id)` per session — replay's own order plus a stable
   tiebreak — then `setval` the sequence past `MAX(seq)`; `turn = 0` for all
   legacy rows (uniformly old, which is correct). Replay switches to
   `ORDER BY seq ASC`, which also removes today's equal-timestamp
   nondeterminism of `created_at` ordering. Loom sqlite: `seq` is the existing
   `messages.id`; add `evicted`, `folded`, `turn` columns.

   `turn` is an **integer, derived at insert exactly like seq — no counter, no
   session state.** Two INSERT variants: the `Chat()`-entry persist site — the
   only turn-incrementing event — uses
   `turn = (SELECT COALESCE(MAX(turn),0) + 1 FROM session_messages WHERE
   session_id = $1)`; every other persist site (assistant, tool result, skill
   sidecar, hygiene fixup) uses the same subquery without the `+1`. The only
   signal threaded through code is one boolean that only the `Chat()` site
   passes. `RETURNING seq, turn` stamps both onto the in-memory message —
   wiring needed anyway so the offload stub can print its `message_id`. No
   concurrency risk: a session is single-writer, messages persist
   sequentially. On reload, `T = max(turn)` over the read rows (§5.1). Loom
   sqlite uses the identical subquery pattern.

   Sidecars never advance the turn and hold no special status beyond that:
   they evict nothing (they are conversation) and they **fold like any
   message**. **When fold flags a region, any skill whose `manage_skills`
   load pair lies inside it is deactivated — its tools leave KERNEL at the
   next compile; re-loading the skill is the door back.** This is what keeps
   live and restored sessions identical: a restored session rebuilds KERNEL
   by walking `folded = false` rows, so a folded load pair is invisible to
   it — the live session must reach the same state by rule, not keep tools a
   restore would lose. Cache cost is zero: fold happens only inside relief,
   which rewrites the whole context anyway.
   `evicted`/`folded` are write-once true, set only inside relief (§5.2); no
   code path clears them.

6. **Summary rows.** The summary persists in the snapshot table
   (`session_memory_snapshots` in cloud, `memory_snapshots` in loom),
   `snapshot_type='fold'`, `content` = JSON `{"n": <int>, "text": "<text>"}`; the
   renderer reads max(n). Cloud's table carries **two** CHECK constraints that do
   not admit `'fold'` — the named one from migration 017 and the auto-named
   inline CHECK from migration 014 — and in Postgres a second CHECK does not
   replace the first, it ANDs with it, so the surviving one silently keeps
   rejecting the new value (observed live: three folds, zero rows persisted).
   The migration must drop **both deployed names**. Normative (migration 199 as
   fixed; migration 200 repairs envs that ran the one-name version):

   ```sql
   ALTER TABLE session_memory_snapshots
       DROP CONSTRAINT IF EXISTS chk_snapshot_type;                              -- migration 017's name
   ALTER TABLE session_memory_snapshots
       DROP CONSTRAINT IF EXISTS session_memory_snapshots_snapshot_type_check;   -- migration 014's inline CHECK
   ALTER TABLE session_memory_snapshots
       ADD CONSTRAINT chk_snapshot_type
       CHECK (snapshot_type IN ('fold', 'fold_entry', 'l2_summary', 'system', 'user', 'agent'));
   ```

   Legacy values stay listed — rows under them may exist and `ADD CONSTRAINT`
   validates existing rows. Loom's sqlite table has no CHECK.

### 5.1 Definitions

- **Window** — the model's context window, as bound via `SetContextLimits`. The
  provider's limit is the only hard boundary; no sub-window budget exists.
  (Deployments that want a smaller working context raise `max_tokens`, which
  shrinks the room the provider accepts — the same lever it enforces itself.)
- **Threshold** — cloud `query_tool_result_threshold_bytes` (system_config,
  Admin-UI settable), bound onto loom's `shared_memory_threshold_bytes`:
  **16384** bytes. One value, three roles: compile-time offload bound,
  persist-time row bound (§4.1), retrieval page bound (§7.1).
- **T** — the session's current turn number: `max(turn)` over the session's
  messages (derived at insert, §4.5). Every turn boundary in this design
  is arithmetic on `turn` against `T` — no rank computations, no marker rows.
- **K** — protected newest user turns. Config `ProtectedRecentTurns`, default **16**.
  The K newest turns are the messages with `turn > T − K`; "older than the K newest
  turns" is `turn ≤ T − K`; the **current turn** is `turn = T`. Flags never enter
  these comparisons.
- **Eviction floor** — a tool row is evicted only if its stored content ≥ 2× its
  rendered stub in bytes. Below that, stubbing saves nothing or goes negative.
- **Estimate** — two tiers over the compiled context, exactly as it would be
  sent (serialized KERNEL + ROM + summary + L1 view). **Cheap tier**: UTF-8
  bytes ÷ 2.7 — a deliberate over-count (JSON and tabular data run ≈3 bytes per
  token; 2.7 biases high) that is free to compute. If the cheap figure is under
  the start mark, that is the estimate — no tokenization on the easy path.
  **Accurate tier**: only when the cheap figure reaches the start mark, count
  real tokens (the tokenizer), memoised per rendered message content so
  unchanged messages are never re-tokenized; tool-call bytes are small and
  counted each time; KERNEL bytes at ÷3. The cheap tier keeps the normal path
  free; the accurate tier keeps relief decisions honest exactly where they
  matter — at the boundary.
- **Reserve** — `max_tokens` as **the value the request actually puts on the
  wire** — cloud resolves it as `lookupMaxOutputTokens(provider, model)`
  (fallback 64000) and passes the pair (window, reserve) to
  `SetContextLimits`; it is **not** a 10%-of-window default from another config
  path. The provider accepts exactly when `prompt + max_tokens ≤ window`, so
  the reserve is part of the boundary, not a safety margin.
- **Water marks** — both percentages of **usable = window − output
  reservation** (the prompt's real ceiling: the provider refuses when
  `prompt + max_tokens > window`, so a mark measured against the full window
  could sit beyond the refusal line and never fire), one base so start and
  release are directly comparable. The values come from the compression
  profile — HWM ← `CriticalThresholdPercent`, LWM ← `WarningThresholdPercent`
  — so they are per-agent configurable through the existing profile channel; a
  missing or inverted pair falls back to the 90/60 defaults:
  - **Start (HWM) = 90%** — relief begins when the estimate reaches it. The
    10% above it is headroom for the estimate's error.
  - **Release (LWM) = 60%** — relief sheds down to it. The 30-point gap is the
    hysteresis band: relief fires about half as often as a 15-point band
    would, each pass shedding more — shedding just enough to fit would re-run
    relief every turn and rewrite (and cache-bust) the context every time.
  - **Recovery penalty = 20 points** — the one pass after a provider refusal
    runs with both marks lowered (start 70% / release 40%). A refusal means
    the estimate under-counted; the lower bar forces the deeper shed the
    normal gate would not trigger. This is the only place the provider's
    refusal touches relief.

### 5.2 The algorithm

1. Emit KERNEL (advertised tool schemas).
2. Emit ROM.
3. If a summary exists, emit its newest version as one system message.
4. L1 = the session's messages with `folded = false`, seq ascending. Folded rows
   are filtered at the **database read** (turn start, reload) — they are never
   loaded into memory at all; the summary stands in for them. Messages appended
   during the current turn are in L1 by arrival, unfolded by definition.
5. `T` = the session's current turn number (§5.1).
6. For each message `m` in L1, in seq order — first matching case wins:
   - `m.role` is user or assistant → emit full. Conversation is never stubbed.
     Stored thinking content renders per the provider's replay rules and
     counts in the estimate like any other bytes; it folds with its row.
   - `m.role` is tool and `m.evicted` → emit the **evicted stub** (§5.5).
     Relief marked it in an earlier pressure event; the flag is write-once, so
     the render is permanent.
   - `m.role` is tool and `bytes(m.content) > threshold` and `m.turn = T` →
     emit the **offload stub** (§5.5). This **is** large-result offload: the
     full payload is in this turn's memory, and the stub's `query_tool_result`
     door (§7.1) works. Strictly over the threshold, not at it: retrieval pages
     and recall returns are bounded **at** the threshold, so an at-threshold
     result renders full — a page must never stub itself.
   - `m.role` is tool and `bytes(m.content) > threshold` and `m.turn < T` →
     emit the **evicted stub**. A legacy unbounded row (persist truncation makes
     over-threshold impossible on new rows): no payload exists in memory, a
     query promise would dangle, the only honest door is re-run.
   - otherwise → emit full.
7. Never separate a tool call from its result — every emitted `tool_use` has its
   `tool_result` and vice versa. Mechanism for the crash window (an assistant
   row persisted, the process dying before its result rows): a call with no
   matching result row gets a **synthetic failed result** emitted in its place,
   normative text: `[no result recorded — the call did not complete. Re-run it
   if needed.]` Synthetic, never strip: stripping the call would delete the
   signature, and the signature is the re-run door.
8. Run **releasePressure** (below). It self-gates: under the start mark it is
   a no-op; at or over it, it sheds to the release mark. Relief runs at
   compile, before every send — loom's own estimate is the trigger, not the
   provider's refusal.
9. Set cache breakpoints — three, each marking the end of a byte-stable
   prefix:
   - **ROM** (always) and the **summary** (when one exists): frozen between
     folds by construction.
   - The **last stable message before any current-turn ephemeral content** —
     where ephemeral means a `turn = T` offload stub (it re-renders as an
     evicted-or-prefix form next turn) or a `turn = T` `query_tool_result`
     call/result pair (never persisted, §4.3 — gone next turn). Everything
     before the marker is turn-stable and cache-safe; everything after is this
     turn's scratch. Each provider client translates the marker to its wire
     form (`cache_control` blocks) and the streaming path requests usage so
     cache reads and writes are metered.
10. Send the request. Provider accepts → done.
11. Provider error other than "context too long" → handle as today.
12. Provider error "context too long" → run **releasePressure once more with
    the recovery penalty** (both marks 20 points lower — the deeper shed the
    under-counting estimate would not trigger), then recompile from step 1 and
    send **once** more. A second "context too long" ends the turn with the
    existing recoverable error carrying `{estimate, target}`. Identification
    is positive, per provider:
    anthropic — status 400, `error.type="invalid_request_error"`, message
    containing `prompt is too long`; OpenAI-shaped (LiteLLM) — status 400,
    `error.code="context_length_exceeded"`. An error not positively identified
    is **not** "context too long" and propagates as today. Anthropic emits two
    distinct messages for this refusal and **both must match**: `prompt is too
    long` when the prompt alone exceeds the window, and `exceed context limit`
    when prompt plus `max_tokens` does. The second is the common one — with any
    real output reservation it fires first, while the prompt is still under the
    window — so a matcher listening only for the first never triggers relief in
    normal operation, and fails silently: `releasePressure` never runs, so §5.6
    prints nothing.

**releasePressure** — one complete pass, no sends inside it, entered from step
8 (penalty 0) and step 12 (recovery penalty):

1. Gate: if the estimate is under the start mark, return — no pressure.
2. Build the boundary ladder by **halving K**: boundaries `T − K, T − K/2,
   T − K/4, … T − 1` (deduplicated). Run **all evictions across the ladder
   first, then all folds** — everything re-runnable is shed before anything
   lossy is compressed. After each operation that changed anything, estimate
   the recompiled context (§5.1): if ≤ the release mark, return — enough is
   shed. An operation that changed nothing (its rows were already flagged by
   an earlier pressure event — the flags are the progress record) passes to
   the next. After the last operation, return regardless. The ladder sheds the
   minimum needed and preserves as much recent context as possible, instead of
   an all-or-nothing K-then-1 cliff.

   ```
   evict(T−K) · evict(T−K/2) · … · evict(T−1) · fold(T−K) · fold(T−K/2) · … · fold(T−1)
   ```

   - **evict(b)** — mark `evicted = true` on every tool row with `turn ≤ b`
     whose stored content ≥ 2× its stub. One transaction. Log (§5.6).
   - **fold(b)** — §5.4 over every `folded = false` row with `turn ≤ b`: write
     summary version n+1 and set the region's `folded` flags — one
     transaction. Log.

2. Properties:
   - Every boundary is `< T`: rows of the current turn are never touched. A
     turn that fills the window by itself is therefore refused again at step
     12's resend and ends in the recoverable error — deliberately, never in
     silent trimming of live work.
   - All eviction runs before any fold. Operations 1–2 shed only what a re-run
     recovers; operations 3–4 are lossy compression of what nothing can
     regenerate. The order **is** the design's first principle, executed.
   - Termination is structural: one gated pass per compile, one recovery pass,
     then the caller's single resend. Repeat pressure later in the same turn
     finds the pass a no-op (flags already set, `turn = T` untouchable), and
     the resend's refusal ends the turn.
   - Flags are write-once; a crash mid-pass is safe — committed flags simply
     render on the next attempt.
   - Each operation updates the in-memory L1 copies of the rows it flags, and
     fold installs the new summary version in memory, in the same step as the
     DB transaction — the recompile that follows must see post-stage state
     without a re-read.
   - The resend after an early stop is safe by construction: near the marks
     the estimate is real tokenizer counts (§5.1 accurate tier), so
     under-counting is bounded to tokenizer-vocabulary drift. If a resend is
     still refused, the recovery pass's 20-point-lower marks force the deeper
     shed, and the flags are durable — a later pass resumes exactly past the
     point this one stopped (completed operations no-op). One recovery pass
     and one resend, at most — do not "fix" the single-resend rule with a
     retry loop.

### 5.3 What the steps mean

Step 6's third case **is** large-result offload; releasePressure's `evict` **is**
eviction; its `fold` **is** fold. None of them exist anywhere else. Offload is a pure render
condition — recomputed every compile, never flagged, cache-stable because a big
result stubs from birth and stays a stub. Eviction and fold are events — they
happen under circumstances that will not recur (pressure), so they are recorded
as the write-once flags and the summary version. Between relief events the
compiled context is append-only: the stub set is frozen by the flags, the summary
is frozen at its newest version, conversation is immutable. Bytes before the tip
change only when relief runs — which is rare, because the hysteresis band
(shed to 60%, re-trigger at 90%) makes each pass buy many quiet turns.

### 5.4 Fold

1. Region: every `folded = false` message with `turn ≤ b`, where `b` is the
   boundary releasePressure passed (`T − K`, then `T − 1`). Pairs fold whole —
   a call and its result are never split across the boundary.
2. Compressor input = the current summary text (empty at first fold) + the
   region **as rendered** — stubs as stubs, so it summarizes conversation, never
   payloads. Input fits the compressor by construction: the region is a subset
   of a context that was just refused at the same-or-larger model's window, and
   it always excludes KERNEL, ROM, and every protected turn — thousands of
   tokens, far more than the compressor's fixed prompt.
3. Compressor = the same model as the agent, or better. The summary can never
   grow unboundedly, by construction: each fold supersedes, and the entire
   summary is the output of a single compressor call — its size is physically
   capped by that call's `max_tokens`, every time, regardless of session
   length. No budget knob exists or is needed.
4. Output = summary version n+1: the cumulative state of the work — what was
   done, what came of it, what is settled, what was ruled out, what is open —
   whose **first line states the covered span** (`covers msg:%d-%d`), citing
   `msg:` addresses for anything worth exact retrieval. The prompt asks for
   state-of-work, never transcript narration. The previous version is
   superseded, not merged — resolved commitments disappear, standing constraints
   carry forward.
5. Compressor failure (normative): version n+1 = the previous text unchanged plus
   one line — `also covers msg:%d-%d (unsummarized): %s` (first user line of the
   region). Coverage is never silently claimed and never lost; the next fold's
   compressor pass absorbs it.
6. The version insert and the region's `folded` flags commit in **one
   transaction**. A fold never stands in memory without its row. A failed insert
   fails the step, which surfaces as the relief error — never a fold that
   evaporates on reload.

### 5.5 Stubs (normative bytes)

Offload stub — a `turn = T` result strictly over the threshold (step 6, third
case); the payload is in memory and queryable:

```
[%s result, ~%d tokens, held in memory this turn — query_tool_result(message_id=%d, offset=0, limit=100) to read it; sql="SELECT ... FROM results" for tabular data.%s
 preview: %s]
```

Evicted stub — `evicted = true` or legacy oversize (step 6, second and fourth
cases); the only door is re-run:

```
[%s result, ~%d tokens, evicted from context — re-run the call above if this data is needed again.%s
 preview: %s]
```

Fields, both forms: `%s` tool name · `~%d` = `ceil(stored_bytes/3)` ·
(offload) `%d` = the message's id (its seq) · trailing `%s` (both forms) = the
row's `[harvested: …]` line(s) if any, prefixed by one space, else empty — the
tails sit at the end of stored content, outside the preview, so the stub must
carry them itself · `preview:` =
**the first 160 bytes of the content with every whitespace run collapsed to a
single space**, cut backward to a rune boundary. Never "the first line":
structured payloads open with a bracket on its own line, and a first-line preview
renders as `[` — a receipt that describes nothing (observed live: 24 of 66
receipts blind). The 160-byte cap is load-bearing: uncapped, a one-line JSON row
yields a stub nearly the row's size and the 2× floor then exempts exactly the
worst rows from eviction.

Eviction is one-sided by design: the result is stubbed; the call above it —
conversation, full signature — stays rendered. At maximum pressure the model
still sees exactly what it needs to regenerate what was shed.

### 5.6 Logging (invariant, not aspiration)

Every releasePressure operation logs, at info: which operation and boundary ·
rows marked or folded (count and seq range) · the estimate and target when
computed · on fold, version n, output bytes, fallback used or not. The
triggering refusal, the resend outcome, and the terminal error log the same
numbers. Every number the system acts on is printed at the moment it acts — a
relief that cannot be reconstructed from its log is a defect.

## 6. Recall

`recall("msg:%d-%d")` — returns the span's **user and assistant rows as
stored** — call signatures included, since they live inside the assistant row
and are what makes "re-run" literal — and **omits `role='tool'` rows
entirely**; a result's only door is re-running its call. Returning stored tool
rows would make recall a stale-data door with durable addresses — the exact
failure mode this design kills — while drowning the conversation the model
asked for. `recall("msg:%d")` is the single-message form, equivalent to a span
of one (session filter built into the read). The addresses come from **the
summary's citations**: everything recallable displays its address at the place
the model learns it exists. Stubs do not advertise recall — a stubbed result's
door is re-run; a summary's cited span is conversation, which no tool can
regenerate, so recall is its only door. Description (normative):

```
Retrieve conversation that is no longer shown, by an address cited in the
session summary: recall("msg:41-78") returns those messages as stored;
recall("msg:41") is the single-message form. Tool results are not included —
re-run the producing call for data. Returns are bounded; a cut range prints
the exact next call.
```

- Returns arrive as an ordinary tool result at the tail — current turn now,
  evictable later, persisted under the same write rules.
- Returns are bounded at the threshold; a range that exceeds it is cut at a row
  boundary and ends with (normative):
  `[range truncated — continue: recall("msg:%d-%d")]`.
- What recall cannot return — data beyond a truncated row's kept prefix — is
  exactly what re-run is for; the row's own tail says so.

Deleted surfaces: `conversation_memory` (tool, promoted slot, offset paging, its
3+-session disclosure), `session_memory` as a read surface, `get_tool_result`,
`get_error_details` and the error store. `LoadMemorySnapshots`/`SaveMemorySnapshot`
are kept and back the summary versions.

### 7.1 `query_tool_result` — this turn's memory, nothing else

Serves pages (`offset`,`limit`) or SQL over one L1 message's full in-memory
payload, addressed by `message_id` — the id printed on the offload stub.
**Every return is bounded at the threshold — pages and SQL alike.** A SQL
result that serializes over the threshold is cut at a result-row boundary and
ends with (normative): `[result truncated at %d bytes — narrow with LIMIT /
aggregation and re-query]` — truncate-and-advise, not an error: the model sees
the shape of what it asked for and self-corrects. With this bound, no
retrieval return can ever trigger the offload render case. Valid only within the producing turn: memory is per turn, large
data is not persisted, and there is no cross-turn data door except re-run. A
`message_id` that does not resolve to a current-turn in-memory payload fails the
call with an error naming the id and saying to re-run the producing call. SQL:
the first `sql` call against a message lazily parses the payload (array of
uniform objects, or `{columns, rows}`) into in-turn SQLite, dropped at turn end;
non-rectangular payloads are paging-only. Call/result pairs are not persisted
(§4.3). Description (normative):

```
Read the full data of a large tool result FROM THIS TURN, by the message_id shown
on the result. Data is held in memory only within the turn that produced it —
never use a message_id from conversation history; re-run the producing tool
instead. (message_id, offset, limit) returns bounded pages; (message_id, sql)
queries tabular data as table `results`.
```

### 7.2 Sandbox — `inputs` and `harvest`

```
inputs: files to create in the sandbox before the command runs.
  source — the file's content (inline), or the message_id of a tool result
           produced EARLIER IN THIS SAME TURN. Data does not survive the turn:
           never use a message_id from earlier in the conversation — re-run the
           producing tool for fresh data.
  dest   — path to create, relative to the working directory.

harvest: files to keep after this turn. Harvested files become durable artifacts,
  and the artifact is thereafter the only copy of that output. To reference,
  share, or build on a harvested result in any later turn, use the artifact —
  never re-run the pipeline to regenerate it.
```

A `message_id` in `inputs.source` that does not resolve to a current-turn
in-memory payload **fails the call** (error `STALE_REFERENCE`, message names the
id) — the silent fallback that wrote the id into the staged file as content is
removed. The loud error is the tripwire that measures the guidance.

### 7.3 Turn end

In-memory full payloads and the in-turn SQLite are dropped. Nothing else happens:
no store to close beyond them, no swaps, no strips (persist already applied its
rules per message). Rows and summary versions are all that survive.

## 8. Reload

Read rows (with `evicted`, `folded`, `seq`, `turn`) and the newest summary
version. Render by §5.2.
That is the entire reload — live and restored are the same read of the same rows;
there is no reload algorithm to diverge. `ReplayMessages`' compaction loop is
deleted; compaction triggers in `AddMessage` are deleted — no state mutates
anywhere but relief.

Skill re-fire on a durable key: an assistant row's `tool_calls_json` entry with
`name=manage_skills`, `input.action=load`, paired by `tool_use_id` with a tool row
whose content starts with `Skill loaded: ` — the confirmation manage_skills
already emits. (Today's marker lives only in loom's `tool_result_json`, which
cloud does not reconstruct — restored cloud sessions currently come back without
their skills' tools advertised.) The restore walk's other branch — disclosure
re-fire, re-registering `query_tool_result` on "stored in memory" content
matches and `get_error_details` on stored-error records — is **deleted** with
those mechanisms: refs never survive a turn, so there is nothing to
re-advertise on restore, and the error store is gone.

Pre-migration sessions (backfilled `seq`/`turn`, no summary, possibly oversize
rows, rows containing old 💡 hints) load as-is: the legacy guard stubs oversize
rows, dead hints age out, their first relief proceeds normally. No content
migration.

## 9. Configuration

| knob | surface | value |
|---|---|---|
| threshold (offload / row bound / page bound) | cloud `query_tool_result_threshold_bytes` → loom shared-memory threshold | 16384 (migration reseeds the 0 seed; envs configured via Admin UI keep their value) |
| K | `ProtectedRecentTurns` | 16 (the halving ladder makes large K cheap: relief tries T−16 first and only walks toward T−1 as needed) |
| water marks | profile `CriticalThresholdPercent` (HWM) / `WarningThresholdPercent` (LWM) | start (HWM) 90% · release (LWM) 60% of usable (window − output reservation); invalid pair → 90/60 defaults |
| recovery penalty | fixed constant | 20 points off both marks, for the one post-refusal pass |
| eviction floor | fixed rule | stored ≥ 2× stub |
| retired | `maxL2Tokens` and its setter (no L2 budget — one summary) | — |

## 10. Rollout

1. Kernel counting — independent.
2. Write rules (§4): truncation + tails, retrieval-pair strip, artifact-before-
   persist, error-store removal, columns + backfill + constraint migration,
   threshold reseed, description rewrites, `STALE_REFERENCE`. Ships alone —
   with one stated dependency: while the old arrival-time offload still exists
   (until step 3), a large result's content is replaced by the ✓-summary
   before persist, so step 2's truncation **must persist from the original
   payload, captured before that replacement** — otherwise it truncates the
   summary text, not the data prefix.
3. Query tool re-addressing (`message_id`, in-turn serving from L1, lazy SQLite);
   delete the per-turn `SharedMemoryStore` and the Postgres tabular branch.
4. ContextCompilation (§5): render rules, relief, fold, summary versions; delete
   the `AddMessage`/`ReplayMessages` triggers, `evictL2ToSwap`, the pin,
   `CompactMemory` wholesale semantics.
5. Recall (§6); delete `conversation_memory`, `session_memory` reads, offset
   paging; unregister removed tools.
6. Reload as read (§8) — mostly deletions once 4–5 exist.

## 11. Invariants

- **All logic in one place.** No mechanism outside ContextCompilation decides
  anything about the context. Arrival appends; persist applies fixed write rules;
  tools serve by contract. Reviewed on every future change.
- **Relief is self-gated on loom's own estimate, at compile, before every
  send.** The provider's refusal triggers only the single deeper recovery
  pass. The normal path pays the cheap byte-bound estimate; real tokenization
  happens only at the boundary.
- **Cache markers sit where the bytes stop being stable.** ROM, summary, and
  the last stable message before any current-turn ephemeral content (offload
  stubs, retrieval pairs). Between lawful transitions — fold, eviction — the
  marked prefix is byte-identical call to call; a transition is the only event
  allowed to rewrite it.
- **Idempotency is the tool contract.** Recovery is re-run; a tool an agent
  cannot safely re-call was never fit to be an agent tool.
- **Pair atomicity.** A call renders and folds with its result. Load-bearing for
  replay; may never be relaxed.
- **Write-once flags, one transaction per operation, set only inside
  releasePressure.** Rendering is a pure function of rows + flags + summary;
  reload is a read.
- **Addresses shown are addresses that resolve.** `msg:` spans cited by the
  summary resolve forever; `message_id` on an offload stub resolves for exactly
  the turn the stub exists in, and its tools say so.
- **Every relief decision logs its numbers** (§5.6).
- **No count knob prices content.** K counts turns; every size decision uses the
  real artifact's bytes.


ilsiepotamus (changes requested):
# Deep-dive review

**Verdict: the core algorithm is genuinely well-built and matches the HLD closely, but this is not mergeable yet.** There is one outright breakage (postgres reads), a cluster of integration gaps where the new design's promises aren't delivered on the paths that matter most (MCP/Teradata results, native Anthropic caching), several silent behavioral regressions from the ROM rewrite, and a failing CI gate.

Methodology: built the PR head locally (`go build -tags fts5` ✓, full `-race` suite across 34 affected packages ✓, exit 0), then ran four parallel deep-review passes (core algorithm, persistence/write rules, LLM clients, tool surface/deletions) and re-verified every CRITICAL/HIGH finding directly against the code before including it here.

## Blockers

### 1. CRITICAL — Postgres session reads are completely broken
All seven message queries now `SELECT` 14 columns (`…, evicted, folded, turn` — `pkg/storage/postgres/session_store.go:188,429,523,605,649,658`) but `scanMessages` (:967) still scans **11** destinations. pgx v5 rejects the mismatch, so `LoadSession`, `LoadMessages`, search, and recall fail on every call — a postgres deployment cannot reload any session. Worse, even with the count fixed, `evicted/folded/turn` are never mapped onto `agent.Message`, so reload would silently lose turn numbers (T=0 for all rows) and all eviction/fold state. CI can't see this: every postgres test is gated behind `TEST_POSTGRES_URL`. This alone shows the postgres path was never exercised once.

### 2. HIGH — MCP adapter truncation starves `query_tool_result`; the flagship path is dead
The executor now passes results through whole ("arrival appends whole"), but the MCP adapter still truncates upstream: `pkg/mcp/adapter/shuttle.go:306-311` cuts to `MaxResultBytes` — 20,000 default, hardcoded **16384** for teradata/postgres/snowflake/bigquery/redshift (`pkg/agent/mcp_integration.go:225,297`) — and **discards the remainder**. A 500KB Teradata SELECT arrives already cut to ~16KB, so the full payload never reaches in-turn memory; the offload stub either never fires, or fires and serves a silently incomplete fragment while telling the model it's reading "the full data" — exactly the §3 failure class this design exists to kill. The adapter's `sharedMemory` wiring is now dead code too. Adapter truncation needs to be disabled (or bypassed) for MCP results to ride the new path.

### 3. HIGH — Tool-suppression policy gate regression
`recall` and `query_tool_result` are registered directly via `a.tools.Register(...)` (`pkg/agent/agent.go:251-252`), bypassing the `RegisterTool` path where `WithoutBuiltinTool` is enforced; on main, `query_tool_result` registration was gated on `isBuiltinToolSuppressed`. `builtinToolsToSuppress()` in `cmd/looms/cmd_serve.go:129-156` still lists three *deleted* tools, lists `query_tool_result` (now a dead entry), and omits `recall`. Net: `tools.none` agents still advertise both tools, and no test covers that path.

### 4. HIGH — Native Anthropic client ignores the compile's cache markers
`pkg/llm/anthropic/client.go` contains zero references to `CacheBreakpoint`; it re-derives breakpoints from a turn-based heuristic (`maxTurn−2` + tip, :238-262). The tip breakpoint can land on exactly the current-turn ephemeral content compile excludes, buying cache writes that are never read back. OpenAI (:320) and Bedrock (:376) honor the marker; the flagship provider doesn't — the "compile marks the perfect place, clients translate it" contract is not true on `provider: anthropic`, and this path has zero test coverage. Related: Anthropic and Bedrock merge ROM+summary into one system block with a single trailing `cache_control`, so a fold invalidates ROM's cache too — the property the separate breakpoints exist for survives only on the LiteLLM path.

### 5. HIGH — Multi-agent workflow instructions silently vanish
`getSystemPrompt` now returns `basePrompt` directly; `formatSystemPromptWithDatetime` → `formatWorkflowCommunicationInstructions` is an unreachable dead chain, yet `pkg/server/multi_agent.go:1291,1488` and `pkg/server/spawn_agent.go:159` still call `SetWorkflowCommunicationContext` expecting injection. Coordinators and sub-agents no longer receive their communication directives at all.

### 6. HIGH — Deleted tools still advertised by live code and shipped defaults
The workflow-coordinator generator still writes prompts teaching `session_memory(action=…)` and puts `"session_memory"` in generated `Tools.Builtin` (`pkg/agent/config_loader.go:1491-1503`); `pkg/validation/semantic.go:354-357,637-641` validates and alias-remaps `get_error_details`/`conversation_memory`/`session_memory`; `README.md:165-166` and `embedded/{weaver,guide}.yaml` + embedded skills still name them. The generator one produces configs that instruct the LLM to call a ghost tool.

### 7. CI is red
gofmt fails on `pkg/agent/agent.go` and `test/context-optimiser/_legacy/route_full_lifecycle_test.go`; "All CI Checks Passed" is failing and the merge is blocked.

## Medium

- **Bedrock has no context-too-long detection** (`pkg/llm/bedrock/client_sdk.go:271-281,695-697` wrap errors opaquely) — the relief backstop never fires there, despite this PR adding Bedrock cache-breakpoint support. Only proactive relief protects it.
- **OpenAI-shaped clients inject `cache_control` unconditionally** (`pkg/llm/openai/client.go:319-347`), and Mistral/HuggingFace delegate to that client. Only LiteLLM→Anthropic consumes it; native endpoints get a foreign field plus string→block content reshaping. Needs gating to Anthropic-backed gateways, or an integration test proving strict endpoints tolerate it.
- **Date grounding removed with no replacement.** The datetime header is gone for ROM byte-stability (deliberate), but there's no date tool and no per-turn injection — agents now have zero anchor for "today", which regresses temporal reasoning. `formatSystemPromptWithDatetime` is dead code.
- **Graph-memory prompt supplement removed, not relocated.** `pkg/agent/agent.go:1193` is now a dangling comment over blank lines; `graphMemoryPromptSupplement` (:1256) has no callers. The tool still registers, but the "remember proactively" instructions are gone, and `docs/architecture/graph-memory.md` still documents them as active.
- **`sm.mu` write lock held across fold compressor LLM calls** — 120s timeout per fold, up to ~5 folds per relief pass (~600s worst case). Concurrent readers (async graph extraction, `query_tool_result`, stats) block for the duration (`pkg/agent/context_compilation.go:413,605-608`).
- **Recovery resend skips the circuit-breaker filter**: `pkg/agent/agent.go:2324` omits `recovery.activeTools(tools)` (present at :2266 and :2302), so a tripped tool is re-offered on the resend.
- **Nothing binds the estimate's reserve to the wire `max_tokens`** — HLD §5.1's own definition demands it. Loom defaults (4096 wire / 10% reserve) are safe, but set `max_tokens: 64000` without `reserved_output_tokens` and the water marks sit *above* the provider's refusal line — proactive relief never fires.
- **Fold starvation**: a call whose result never persisted (crash window) permanently blocks `adjustCompressionBoundary` (`pkg/agent/segmented_memory.go:404-409`), so fold can never advance past it — even though compile would synthesize the missing result anyway.
- **Inter-agent sends >16KB always fail (latent)**: `pkg/agent/agent_communication.go:53,142,209` truncate marshaled JSON with the English tail → `Receive`'s `json.Unmarshal` (:103) fails every time. No production callers today, but it's a shipped API.
- **SQLite cross-agent reads diverge**: `LoadMessagesForAgent`/`LoadMessagesFromParentSession` don't select the new columns or filter `folded` (the postgres versions do — and hit blocker #1).
- **HLD §7.2 (sandbox `inputs`/`harvest`/`STALE_REFERENCE`) is not in this repo** — only the `harvestTails` parser landed; the producer presumably lives cloud-side. Fine if intentional, but the PR body presents it as part of this change; state the split explicitly.

## Low / cleanup (compressed)

- `query_tool_result` page emits a single oversized unit whole, exceeding the 16KB bound (`pkg/agent/builtin_tools.go:289-297`).
- Soft reminder appends to the ROM message (cache-bust at exactly the high-pressure moment) and is silently lost when relief recompiles `messages` (`pkg/agent/agent.go:2240-2244` vs :2300).
- Thresholds under ~100 bytes break the `stored ≤ threshold` invariant (`pkg/agent/write_rules.go:36-57`); no floor validation on `SetThreshold`.
- `covers msg:` span enforcement is prefix-presence only (`pkg/agent/context_compilation.go:613`), so a compressor echoing the old line keeps a stale span.
- `pkg/agent/types.go:289` says K default is 5, actual is 16; "K newest turns never touched" comments overclaim (the ladder reaches T−1).
- Cheap-tier overcount makes the shed loop target ~49% true utilization instead of 60% (over-shedding, extra folds); kernel counted at ÷2.7 in the cheap tier vs the HLD's ÷3.
- `ReleasePressure` returns `shed=true` when nothing shed → redundant recompile and a guaranteed-refusal resend.
- `max_tool_results` config parsed but inert (`config_loader.go:187,628`); `RegistryConfig.ErrorStore` removed — **check loom-cloud/avmo-tera-cloud compile before merging** (`BuildSkillsOptions` itself is unchanged).
- Storeless sessions print `message_id=0` on offload stubs (`in_turn_payload.go:56-62`).
- Migration 000019 leaves `agent_errors`/`sql_result_metadata` tables orphaned (follow-up cleanup migration).
- Stale comments: `SessionReferenceTracker` in `pkg/storage/shared_memory.go:307`, disclosure re-fire comments in `memory.go:325,411-413`, "rebuilt each turn" on the now-frozen task context (`agent.go:1196-1198`).
- Dead code: `formatSystemPromptWithDatetime`, `formatWorkflowCommunicationInstructions`, `graphMemoryPromptSupplement`, `min()` in `segmented_memory.go:337`, write-only `refStore`/`WithReferenceStore`, adapter `sharedMemory`.

**Test estate**: the new `test/context-optimiser` suite is env-gated (`LOOM_CONTEXT_OPTIMISER`) so it never runs in CI, and `_legacy/` (~4.5K lines) is invisible to the Go toolchain entirely (underscore dir) and references deleted APIs — it can never compile. Recommend pruning `_legacy` and deciding whether `PLAN.md` belongs in the repo. Also: no `docs/` updates for a change this size, while existing docs now describe deleted mechanisms.

## What's genuinely good

The SQLite write-rules implementation is verified correct in detail — truncation math with rune-boundary fixpoint, both-sides retrieval-pair strip, the turn subquery with `RETURNING` stamping, one-transaction fold/evict, write-once flags with no clearing path. The two-tier estimate keyed on *rendered* content is staleness-free by construction. `pkg/llm/errors.go` gets both Anthropic refusal wordings plus the LiteLLM passthrough right, with real wire-shape tests, and retry correctly never retries context-too-long. `recall` and `query_tool_result` match their normative contracts nearly byte-for-byte. Zero dangling Go references to any deleted surface. Race detector is clean across everything. The 26 commits are well-scoped and the HLD-as-PR-body is exemplary.

## Suggested sequencing

1. Fix postgres `scanMessages` (+ map the three fields) and add at least one postgres CI job so this class can't recur; fix gofmt.
2. Decide the MCP truncation story — the PR's value proposition depends on it.
3. Restore the suppression gate, the workflow-comm injection, and a date mechanism (per-turn, outside the cached prefix).
4. Make the Anthropic client honor `CacheBreakpoint`; add Bedrock refusal detection.
5. Sweep the ghost-tool references (generator, validation, README, embedded YAMLs) and prune `_legacy`.

mrg-ray:
Addressed in `501eaaf9..540dde87` — the mechanical findings; the design-decision items are deferred deliberately and listed at the end.

**Blocker 1 (postgres reads) — fixed, `501eaaf9`.** `scanMessages` scans all 14 columns and maps `evicted`/`folded`/`turn` onto `agent.Message`. Proof on live postgres: `TestRLS_UserIsolation_Messages` fails at the parent commit and passes now, plus a new gated roundtrip test (`session_messages_integration_test.go`) asserting turn stamping, the folded read-filter, and evicted-flag survival through save→load. The sqlite cross-agent readers (`LoadMessagesForAgent`, `LoadMessagesFromParentSession`) now select the new columns and filter `folded = 0`, matching postgres.

**Blocker 3 (suppression gate) — fixed, `dc2dfe7e`.** Both retrieval tools register through `RegisterTool`; `builtinToolsToSuppress()` drops the three deleted names and covers `query_tool_result` + `recall`; gate test added.

**Blocker 4 (anthropic client) — fixed, `a2710d6b`.** One system block per source message carrying the compile's marker — ROM and the summary cache independently, so a fold no longer invalidates ROM's prefix. Message breakpoints come only from `CacheBreakpoint`; the turn heuristic and tip mark are gone. Test rewritten to the marker contract (marked blocks carry `cache_control`, unmarked don't).

**Blocker 6 (ghost tools) — swept, `540dde87`.** Generator, validation lists and aliases, README, embedded weaver/guide YAMLs and skills. `recall` replaces the deleted retrieval aliases.

**Blocker 7 (CI) — was fixed before this batch; all checks green since `dfc0bd7f`.**

**Mediums:** Bedrock context-refusal detection added at all three error sites (`IsBedrockContextTooLong`, matching anthropic's wordings plus Bedrock's own; wire-string tests). `cache_control` emission gated to claude models on the OpenAI-shaped client. Recovery resend now applies `recovery.activeTools`. Soft reminder rides as a transient trailing system message — never on the ROM message, survives relief recompiles. Fold starvation: a prior-turn group missing results no longer pins `adjustCompressionBoundary` (compile synthesizes the missing result). SQLite cross-agent reads aligned (above).

**Lows:** `ReleasePressure` returns `shed=false` when nothing shed. `covers msg:` line must reach this fold's `hiSeq` — a stale echo no longer claims coverage. `SetThreshold` clamps to a 256-byte floor. Oversized page units are bounded with the §4.1 tail. Storeless rows render the evicted stub instead of `message_id=0`. K comment corrected. Dead code deleted (`formatSystemPromptWithDatetime`, `graphMemoryPromptSupplement`, local `min`); stale comments fixed; `_legacy/` (~4.5K lines) and `PLAN.md` pruned; graph-memory doc updated.

**Verified:** full build, `pkg` suites + race on agent/anthropic/openai/bedrock, gated e2e suite green, golangci-lint 2.9.0 clean, postgres integration suite on a live instance (remaining failures there are role-provisioning artifacts of the local environment, identical at the parent commit).

**Deferred — design decisions, not fixes:**
- **Blocker 2 (MCP adapter truncation):** where the 16 KB bound lives is an architecture call; the adapter bypass/disable needs its own change with the cloud path in view.
- **Blocker 5 (workflow-comm injection):** where those instructions belong in a byte-stable-ROM world. `formatWorkflowCommunicationInstructions` is kept as the seam.
- **Date grounding:** same class — per-turn injection outside the cached prefix vs. a tool.
- **`sm.mu` across fold compressor calls:** lock restructuring around a 120 s external call.
- **Reserve ↔ wire `max_tokens` binding, postgres CI job, un-gating the e2e suite, orphaned-table cleanup migration:** queued with the above.

ilsiepotamus:
Re-verified the fix batch (`501eaaf9..540dde87`) independently: postgres scan + field mapping, sqlite cross-agent alignment, suppression gate through `RegisterTool`, Anthropic honoring the compile's markers with ROM/summary as separate system blocks, Bedrock refusal detection, `cache_control` gated to claude models, `activeTools` on the resend, transient soft reminder, starvation guard, threshold floor, honest `shed` return, bounded page units, ghost-tool sweep, `_legacy` pruned. Full build + `-race` across the 34 affected packages green locally, CI green. Good batch — thank you for the fast turnaround.

**Two of the deferred items need to land in this PR, because they are regressions** — behavior that shipped on `main` and silently stopped working — not open design questions:

1. **Workflow-comm injection (blocker 5).** `pkg/server/multi_agent.go:1291,1488` and `pkg/server/spawn_agent.go:159` still call `SetWorkflowCommunicationContext`, but `formatWorkflowCommunicationInstructions` has no callers — coordinators, sub-agents, and spawned agents no longer receive their communication directives at all. The seam you kept just needs a caller; the transient trailing-message pattern you built for the soft reminder is a ready-made home outside the cached prefix.

2. **Date grounding.** The datetime header was the only date anchor agents had; deleting it for ROM byte-stability without a replacement leaves every agent date-blind for relative-date reasoning ("last month", "due Friday", temporal recall). Same fix shape: a one-line date (not time-of-day, to keep the bytes stable within a day if desired) riding the transient trailing message, or a datetime tool — either restores the behavior without touching the cached prefix.

The changes-requested review stays open pending these two.

The remaining deferrals are now tracked so they don't evaporate:
- #290 — MCP adapter truncation starving `query_tool_result` (blocker 2)
- #291 — `SegmentedMemory.mu` held across fold compressor calls
- #292 — reserve ↔ wire `max_tokens` binding
- #293 — postgres CI job
- #294 — un-gating a `test/context-optimiser` smoke subset in CI
- #295 — orphaned `agent_errors`/`sql_result_metadata` tables

One trivial leftover from the sweep: `cmd/looms/cmd_serve.go`'s `otherMechanismTools` map still describes `get_tool_result`/`get_error_details` as provided by mechanisms that no longer exist — fine to fold into either fix commit.

mrg-ray:
Thanks — this was a thorough read and it found real defects. Every finding is answered below. Where a fix has a test that fails without it, the row says so; the rest are verified by inspection only.

## Blockers

| # | Finding | Status | How |
|---|---|---|---|
| 1 | Postgres session reads broken | Fixed | `scanMessages` scans all 14 columns and maps `evicted`/`folded`/`turn`. Tested: live-postgres roundtrip, fails without the fix |
| 2 | MCP adapter truncation starves `query_tool_result` | Fixed | The adapter's byte and row bounds are deleted, so results reach arrival whole and one threshold governs. Tested: >40KB payload arrives byte-identical |
| 3 | Tool-suppression gate regression | Fixed | Both retrieval tools register through `RegisterTool`; the suppress list and its doc comment now name the current tools. Tested |
| 4 | Native clients ignore the compile's cache markers | Fixed | Anthropic and Bedrock each emit one system block per system message, carrying `cache_control` only where compile marked it. Tested on both |
| 5 | Multi-agent workflow instructions vanish | Fixed | The team block is reconnected into the ROM build. Tested, including placement and byte-stability |
| 6 | Deleted tools still advertised | Fixed | Swept from the generator, validation, README, embedded YAMLs, operator help and comments; the `autoInjectedOrSpecialTools` whitelist no longer accepts them. Tested |
| 7 | CI red | Fixed | gofmt and golangci-lint clean; all checks currently green |

## Medium

| # | Finding | Status | How |
|---|---|---|---|
| 1 | Bedrock has no context-too-long detection | Fixed | A positive matcher is wired at all three error seams. Caveat: it matches Anthropic's two wordings plus one Bedrock phrasing we have not observed on the wire, so it may be incomplete rather than wrong |
| 2 | `cache_control` injected unconditionally | Fixed | The OpenAI-shaped client emits it only for Claude models, identified by model name — so a Claude route named otherwise loses the breakpoint, which costs caching but breaks nothing |
| 3 | Date grounding removed with no replacement | Fixed | Time is written into the user turn at arrival, once, leaving ROM byte-stable. Tested |
| 4 | Graph-memory supplement removed, not relocated | Fixed | Dead function deleted, dangling comment resolved, architecture doc corrected |
| 5 | `sm.mu` held across the fold compressor call | Fixed | The lock is released around the LLM call and the snapshot re-validated on re-acquire. Tested |
| 6 | Recovery resend skips the circuit-breaker filter | Fixed | `recovery.activeTools` now applied at all three call sites. Verified by inspection |
| 7 | Reserve not bound to the wire `max_tokens` | Fixed | The effective reservation is carried in the agent config, and `SetContextLimits` treats 0 as unset so it cannot be clobbered. Tested at the construction seam |
| 8 | Fold starvation on a never-persisted call | Fixed | A prior-turn incomplete group no longer pins the boundary, since compile synthesizes the missing result. Verified by inspection |
| 9 | Inter-agent sends >16KB always fail | Fixed | The payload is bounded without corrupting the encoding, so `Receive` always unmarshals. Tested |
| 10 | SQLite cross-agent reads diverge | Fixed | SQLite aligned; while verifying we found the two postgres cross-agent readers had the same gap and fixed those too. Tested on live postgres |
| 11 | HLD §7.2 sandbox surface absent | Confirmed | Only the `harvestTails` consumer belongs in this repo; the producer is cloud-side. Stating the split rather than treating it as deferred work |

## Low / cleanup

| # | Finding | Status | How |
|---|---|---|---|
| 1 | Oversized page unit rides whole | Fixed | A unit larger than the page budget is bounded with the §4.1 tail. Inspection only |
| 2 | Soft reminder appended to the ROM message | Fixed | It rides as a transient trailing system message, never stored, and survives a relief recompile. Inspection only |
| 3 | Small thresholds break `stored ≤ threshold` | Fixed | All three threshold consumers clamp to a floor, not just `SegmentedMemory`. Tested |
| 4 | `covers msg:` enforcement is prefix-only | Fixed | The span's upper bound must now reach this fold's `hiSeq`. Inspection only |
| 5 | K comments overclaim | Fixed | Default corrected to 16; comments say the ladder walks toward T−1 |
| 6 | Cheap-tier overcount → ~49% shed target | Open | Confirmed. The loop terminates on the cheap tier because the release mark sits below the start mark. Left as a tuning change rather than folded into this PR |
| 7 | `ReleasePressure` returns `shed=true` when nothing shed | Fixed | It returns false, sparing the redundant recompile and guaranteed-refusal resend. Inspection only |
| 8 | `max_tool_results` parsed but inert | Open | Confirmed inert. Separately checked: the cloud repo has no `RegistryConfig` reference, so that removal breaks nothing there |
| 9 | Storeless sessions print `message_id=0` | Fixed | A row with no durable id renders the evicted stub, whose door is re-run. Inspection only |
| 10 | Migration 000019 leaves tables orphaned | Open | Agreed. A drop migration is destructive, so it is being handled as its own change |
| 11 | Stale comments | Fixed | All four corrected |
| 12 | Dead code | Partial | All deleted except `refStore`/`WithReferenceStore`, which still has live setters in `pkg/server` and so is not a free removal |
| 13 | Test estate | Partial | `_legacy/` and `PLAN.md` are gone. The env gate is kept deliberately: the suite is type-checked by the lint job — which is how CI caught issues in it during this review — but no job executes it. There is still no postgres CI job, and that is the gap that let blocker 1 through |

Two notes. On the refuted `harvestTails` item: agreed, and thanks for measuring it rather than filing it. On finding 3 in the medium list: the fix covers the text path; a caller supplying `ContentBlocks` still has its own content, since providers prefer blocks over `Content`.

ilsiepotamus (approved):
Approving — the changes-requested review is resolved.

Verification across three rounds: all seven blockers fixed and independently re-verified in code (postgres scan+mapping, MCP adapter bound deleted, suppression gate, Anthropic/Bedrock honoring compile's cache markers with per-message system blocks, workflow-comm reconnected, ghost-tool sweep, CI green). Both requested regression fixes landed — workflow-comm with tests, and the arrival-timestamp approach to date grounding is a genuinely better solution than what I suggested. Full `-race` suite green locally on `631503f2` across all 34 affected packages, including the restructured fold lock under load.

Remaining open items are honestly tracked: #293 (postgres CI job — the real gap), #294 (e2e smoke in CI), #295 (drop migration), plus the cheap-tier shed-target tuning and inert `max_tool_results` noted in your table. #290/#291/#292 will be closed with pointers to the fixing commits once this merges.

PR #298, opened 2026-08-06
fix(agent): load parallel workflow tasks from canonical schema

Ed-Jaramillo:
## Problem

Canonical parallel workflows list their agents under `spec.tasks[].agent_id`:

```yaml
spec:
  type: parallel
  tasks:
    - agent_id: agent-one
      prompt: "do task one"
    - agent_id: agent-two
      prompt: "do task two"
```

The Registry loader that turns workflow files into chattable coordinators grouped `parallel` with `fork_join` and looked for `spec.agent_ids` instead. Because canonical parallel workflows do not have that field, the loader extracted no agents and returned `no agents found in workflow` before creating the coordinator or any namespaced sub-agents.

The direct execution parser already reads `tasks[].agent_id` correctly. This change aligns the Registry loader with that canonical schema.

## Change

- Give `parallel` its own loader case and read `spec.tasks[].agent_id`.
- Require `tasks` to be an array and each task to be an object with a non-empty string `agent_id`.
- Return indexed errors for malformed tasks so the Registry cannot partially register an unrunnable workflow.
- Deduplicate repeated agent IDs while preserving first-seen order, creating one namespaced sub-agent per unique agent.
- Leave the fork-join extraction behavior, proto schema, producers, and `prompt` validation unchanged.

## Regression coverage

- Canonical parallel workflow loading.
- Three tasks assigned to two unique agents, including first-seen ordering and deduplication.
- Coordinator and executor workflow metadata.
- Missing or non-array `tasks`.
- Non-object tasks.
- Missing, empty, non-string, and whitespace-only `agent_id` values.

## Validation

- `go test -tags fts5 -race ./pkg/agent -run TestLoadWorkflowAgents_Parallel -count=1`
- `just build`
- `just security`

`just check` was also run. Proto generation, formatting, vet, lint, and the changed tests passed, but the repository-wide race suite encountered existing Windows file-lock, path, and permission failures outside this change. Build and security were therefore run separately and passed.

## Relationship to #289

This is independent from #289 and can be reviewed or merged separately. Both PRs touch the workflow-pattern switch, so whichever lands second may need to resolve a one-line context conflict.


ilsiepotamus (approved):
Verified the bug independently — it's real and hits our own shipped examples. The proto (`ParallelPattern.tasks[].agent_id` in `proto/loom/v1/orchestration.proto`), the converter (`pkg/orchestration/workflow_config.go:383-389`), and the validator (`pkg/validation/validator.go:432-435`) all agree on `tasks[].agent_id`, while the registry loader alone read `spec.agent_ids`. Result: canonical parallel workflows (including `examples/reference/workflows/orchestration-patterns/doc-generation.yaml`) passed `looms workflow validate` but were silently skipped at startup and hot-reload.

The fix aligns the loader with the proto-canonical schema, with order-preserving dedup matching the existing pipeline/conditional idiom. The behavior change for non-canonical `parallel` + `agent_ids` files (now hard-erroring at load instead of registering a coordinator that could never execute) is honest tightening — that shape never worked end-to-end since the converter rejects it. Tests are solid: positive path plus a 6-case table-driven negative test, 93.75% patch coverage.

One follow-up to track separately (not a blocker for this PR): the sibling `fork-join`/`fork_join` spelling mismatch from #289/#287 still exists on main — the loader accepts only `fork_join` while the validator, converter, and Weaver's `apply_template` emit `fork-join`, so shipped `type: fork-join` examples still silently fail to register. That fix should be re-raised.

I've updated the branch to current main (clean merge); approving — merge once the re-run checks are green.

PR #299, opened 2026-08-07
Workload profiles select the relief marks; cache-aware and gateway-correct cost

mrg-ray:
Four changes that share one theme: values a workload profile or a gateway already carries were not reaching the code that needs them.

## profiles: a workload profile now selects the relief marks

ContextCompilation's relief has two water marks — the high mark where relief begins and the low mark it sheds down to, both percentages of usable context (window minus reserved). They were hardcoded constants, so `CompressionProfile` — which already carries `CriticalThresholdPercent` and `WarningThresholdPercent` — described the behaviour without setting it. Cloud's Workload Profile selector wrote a value nothing read.

`marksLocked()` now sources both marks from the profile, falling back to the constants when the pair is unusable (release ≥ start, non-positive, or start > 100). The three presets get distinct bands, which is what makes the selector mean anything:

| Profile | high | low | Why |
|---|---|---|---|
| data_intensive | 80 | 45 | Growth arrives in bursts — one tool result can add tens of points in a call. Start early so a burst has headroom above the mark; shed deep so one pass buys many turns instead of thrashing per call |
| balanced | 90 | 60 | A 30-point band: relief roughly half as often as a 15-point band, each pass shedding more |
| conversational | 92 | 70 | Growth is small and steady and recent turns are the value. Start late, shed shallow, keep as much recency as the bound allows |

The remaining `CompressionProfile` fields (`MaxL1Tokens`, `MinL1Messages`, the batch sizes) are vestiges of batch compaction and drive nothing; the comment now says so rather than implying they are live.

Recovery penalties apply through `applyPenalty`, which floors at 1 — subtracting a penalty from an admin-set 5% must not produce a negative mark.

## roms: skill instructions outrank the general operating rules

A loaded Skill arrives once, as a conversation message. START_HERE and the system prompt are resident on every call. With no stated precedence the resident rule wins by default, which is the wrong way round — the Skill is the specific instruction.

Measured on the td-data-stats suite: "prefer doing the work over describing it" suppressed the Skill's "always return SQL" contract on 19 of 20 answers, including four turns that asked for SQL outright.

START_HERE now states the precedence, bounded to method and presentation. It never licenses reporting a figure a tool did not return, or narrowing the data without saying so.

## llm: cost must respect cache tiers, and defer to the gateway's own figure

A cache write bills at 1.25x the input rate and a cache read at 0.10x. `calculateCost` took only (input, output) and charged every input token at the full rate. On a ContextCompilation run where 91.7% of input was cache reads, that reported $2.49 against an actual $0.72 — the design's entire benefit is converting expensive tokens into 0.1x reads, and the cost function could not see it.

litellm already returns a correct, cache-aware cost in `x-litellm-response-cost`, verified against its own arithmetic to eight decimals; nothing read it. It is now authoritative, with a cache-aware local estimate as fallback (a zero or absent header means fall back).

The two providers differ in what they count, which is where the double-count risk sits: OpenAI-compatible `prompt_tokens` **includes** cached tokens, so the uncached remainder is derived by subtraction; Bedrock/Anthropic `input_tokens` **excludes** them, so the buckets are additive.

Also makes bedrock's legacy client cache-aware — its SDK client already was — and parses the cache fields it was discarding.

## llm/openai: price gateway-proxied Claude with Anthropic rates

Gateways proxy non-OpenAI models through the OpenAI-compatible client under ids like `coding-agent/claude-sonnet-4-6`. Absent from the catalog, they fell through to the GPT rate card and were billed at $2.50/$10 instead of $3/$15 — a 20% understatement on every Sonnet call, and worse for Opus.

The Anthropic family is now recognised by substring first, mirroring the bedrock client, with `claude-opus-4-1` checked before `claude-opus` since the substring test would otherwise swallow it.

The response-cost header cannot serve here, and the code records why: on streaming calls litellm flushes headers before the stream completes, so it reports a zero cost and the local estimate is the only correct source. loom streams every agent call.

## server/test: a test logger that survives its background workers

Unrelated to the above, but it was failing Unit Tests under `-race` and had to be fixed here. The coordinator notification handler spawned by `spawnWorkflowSubAgents` runs until its context is cancelled and logs once on the way out; the test cancels in `t.Cleanup` but nothing waits for that log, so it can land after `tRunner` has returned. A `zaptest` logger writing to a finished `*testing.T` is a race — and with the timing forced, a hard panic: *"Log in goroutine after ... has completed"*.

The worker is correct as written — a background goroutine that must outlive the request context, annotated as such. The harness was the wrong half. `newTestLogger` forwards to `t.Logf` under a mutex and drops writes once the test is over, registered early so its cleanup runs last, after every shutdown has been signalled; nothing that happens during the test is lost.

Verified by forcing the race: with a 300ms sleep before the shutdown log, the old harness panics every run and the new one is clean.

## Verification

`go build -tags fts5 ./...`, `go test -tags fts5 ./pkg/agent/ ./pkg/llm/...` — all green. `golangci-lint run` reports 0 issues; `gofmt -l` clean.

Cost arithmetic is pinned by `pkg/llm/openai/cost_cache_test.go`, including that a gateway-proxied Claude is not priced as a GPT.

### DIFF
--- a/pkg/agent/compression_integration_test.go
+++ b/pkg/agent/compression_integration_test.go
@@ -44,8 +44,8 @@ agent:
 	assert.Equal(t, "data_intensive", profile.Name)
 	assert.Equal(t, 4000, profile.MaxL1Tokens, "Should use data_intensive MaxL1Tokens (4000)")
 	assert.Equal(t, 3, profile.MinL1Messages)
-	assert.Equal(t, 60, profile.WarningThresholdPercent)
-	assert.Equal(t, 90, profile.CriticalThresholdPercent)
+	assert.Equal(t, 45, profile.WarningThresholdPercent)
+	assert.Equal(t, 80, profile.CriticalThresholdPercent)
 
 	// Step 3: Create agent with compression profile
 	agent := createTestAgentWithProfile(profile)
@@ -63,8 +63,8 @@ agent:
 	assert.Equal(t, 4000, segMem.maxL1Tokens, "Should use data_intensive maxL1Tokens (4000 tokens)")
 	assert.Equal(t, 3, segMem.minL1Messages, "Should use data_intensive minL1Messages")
 	assert.Equal(t, "data_intensive", segMem.compressionProfile.Name)
-	assert.Equal(t, 60, segMem.compressionProfile.WarningThresholdPercent)
-	assert.Equal(t, 90, segMem.compressionProfile.CriticalThresholdPercent)
+	assert.Equal(t, 45, segMem.compressionProfile.WarningThresholdPercent)
+	assert.Equal(t, 80, segMem.compressionProfile.CriticalThresholdPercent)
 }
 
 // TestCompressionProfile_EndToEnd_Conversational verifies conversational profile
@@ -90,8 +90,8 @@ agent:
 	assert.Equal(t, "conversational", profile.Name)
 	assert.Equal(t, 9600, profile.MaxL1Tokens, "Should use conversational MaxL1Tokens (9600)")
 	assert.Equal(t, 6, profile.MinL1Messages)
-	assert.Equal(t, 60, profile.WarningThresholdPercent)
-	assert.Equal(t, 90, profile.CriticalThresholdPercent)
+	assert.Equal(t, 70, profile.WarningThresholdPercent)
+	assert.Equal(t, 92, profile.CriticalThresholdPercent)
 
 	// Create agent and verify memory uses conversational profile
 	agent := createTestAgentWithProfile(profile)
--- a/pkg/agent/compression_profiles.go
+++ b/pkg/agent/compression_profiles.go
@@ -45,12 +45,22 @@ type CompressionProfile struct {
 }
 
 // ProfileDefaults provides preset profiles for common workload types.
-// These are static fallback values - use NewSegmentedMemoryWithDynamicAllocation for adaptive sizing.
+//
+// The two thresholds are the relief water marks (HLD §5.1), percentages of
+// usable context: CriticalThresholdPercent is the HWM — where relief begins —
+// and WarningThresholdPercent is the LWM it sheds down to. The gap between
+// them is hysteresis: a wide band means rare, deep passes; a narrow band means
+// frequent, shallow ones. That is what a workload profile actually selects.
+//
+// The remaining fields (MaxL1Tokens, MinL1Messages, the batch sizes) are
+// vestiges of batch compaction and drive nothing.
 var ProfileDefaults = map[loomv1.WorkloadProfile]CompressionProfile{
 	loomv1.WorkloadProfile_WORKLOAD_PROFILE_BALANCED: {
-		Name:                     "balanced",
-		MaxL1Tokens:              6400, // ~8 messages @ 800 tokens each (static fallback)
-		MinL1Messages:            4,
+		Name:          "balanced",
+		MaxL1Tokens:   6400, // ~8 messages @ 800 tokens each (static fallback)
+		MinL1Messages: 4,
+		// The default: a 30-point band, relief roughly half as often as a
+		// 15-point band would fire, each pass shedding more.
 		WarningThresholdPercent:  60,
 		CriticalThresholdPercent: 90,
 		NormalBatchSize:          3,
@@ -58,21 +68,27 @@ var ProfileDefaults = map[loomv1.WorkloadProfile]CompressionProfile{
 		CriticalBatchSize:        7,
 	},
 	loomv1.WorkloadProfile_WORKLOAD_PROFILE_DATA_INTENSIVE: {
-		Name:                     "data_intensive",
-		MaxL1Tokens:              4000, // ~5 messages @ 800 tokens each (static fallback)
-		MinL1Messages:            3,
-		WarningThresholdPercent:  60,
-		CriticalThresholdPercent: 90,
+		Name:          "data_intensive",
+		MaxL1Tokens:   4000, // ~5 messages @ 800 tokens each (static fallback)
+		MinL1Messages: 3,
+		// Growth arrives in bursts: one tool result can add tens of points in a
+		// single call. Start early so a burst has headroom above the mark, and
+		// shed deep so one pass buys many turns instead of thrashing per call.
+		WarningThresholdPercent:  45,
+		CriticalThresholdPercent: 80,
 		NormalBatchSize:          2,
 		WarningBatchSize:         4,
 		CriticalBatchSize:        6,
 	},
 	loomv1.WorkloadProfile_WORKLOAD_PROFILE_CONVERSATIONAL: {
-		Name:                     "conversational",
-		MaxL1Tokens:              9600, // ~12 messages @ 800 tokens each (static fallback)
-		MinL1Messages:            6,
-		WarningThresholdPercent:  60,
-		CriticalThresholdPercent: 90,
+		Name:          "conversational",
+		MaxL1Tokens:   9600, // ~12 messages @ 800 tokens each (static fallback)
+		MinL1Messages: 6,
+		// Growth is small and steady, so the risk of overshooting between turns
+		// is low and recent turns are the value. Start late and shed shallow to
+		// keep as much recency as the bound allows.
+		WarningThresholdPercent:  70,
+		CriticalThresholdPercent: 92,
 		NormalBatchSize:          4,
 		WarningBatchSize:         6,
 		CriticalBatchSize:        8,
--- a/pkg/agent/compression_profiles_test.go
+++ b/pkg/agent/compression_profiles_test.go
@@ -41,8 +41,8 @@ func TestProfileDefaults(t *testing.T) {
 			profile:               loomv1.WorkloadProfile_WORKLOAD_PROFILE_DATA_INTENSIVE,
 			expectedMaxL1:         4000, // 5 messages @ 800 tokens each
 			expectedMinL1:         3,
-			expectedWarning:       60,
-			expectedCritical:      90,
+			expectedWarning:       45,
+			expectedCritical:      80,
 			expectedNormalBatch:   2,
 			expectedWarningBatch:  4,
 			expectedCriticalBatch: 6,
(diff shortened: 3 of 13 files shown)


ilsiepotamus (changes requested):
Three independent review passes (pkg/agent relief marks; pkg/llm cost; pkg/server logger + START_HERE + PR-body audit). All touched packages independently re-verified with `go test -tags fts5 -race` (pkg/agent full, pkg/server `-count=5` on the affected tests, all 14 pkg/llm packages), `gofmt -l` clean.

**The code is close to approve-worthy.** The token-accounting split is correct per provider shape (OpenAI-compatible subtractive with clamp, Bedrock/Anthropic additive — AWS SDK Converse field names verified against bedrockruntime v1.56.0, litellm's inclusive `prompt_tokens` normalization verified), the litellm wrapper client inherits the fix, catalog precedence cannot shadow the family matcher, the logger fix targets the right half (harness, not the worker) and the LIFO-cleanup reasoning holds, and the new tests pin real behavior — the marks test exercises `startMarkLocked`/`releaseMarkLocked` rather than restating constants, and the cost test's cache-blind comparison is a genuine behavioral pin.

Requesting changes on four items — three are accuracy/honesty edits, one is a small code fix.

## Request changes

**1. The PR body credits this PR with #288's work.** `git diff main...HEAD -- pkg/agent/context_compilation.go` is empty. `marksLocked()` — profile-sourced, with the exact fallback condition (`release <= 0 || release >= start || start > 100`) — and `applyPenalty`'s floor at 1 exist verbatim at the merge base; the merge base *is* #288, which introduced them. "They were hardcoded constants" is also wrong about the pre-PR state: the presets were flat 60/90, so the *selector* was inert while explicit per-field overrides worked. The branch's own commit message (6a626eb9) states this correctly — the body should say what it says: #288 wired the marks to the profile; this PR gives the presets distinct values.

**2. "The response-cost header cannot serve here, and the code records why" — the code records nothing.** No comment in `pkg/llm/openai/` mentions litellm flushing headers before stream completion. The behavior is consistent with the claim (`ChatStream` reads the header; zero → fallback), so either add the comment or drop the claim from the body.

**3. Docs tabulate preset thresholds that match neither side of this PR.** Four pages still carry the pre-#288 values (60/75, 50/70, 70/85): `docs/reference/agent-configuration.md:485-489`, `docs/architecture/memory-systems.md:305-309`, `docs/guides/memory-management.md:79-135`, `docs/architecture/agent-system-design.md:921-923`. `docs/architecture/memory-systems.md:298` also claims "the only floor is `minL1Messages`" — directly contradicted by this PR's own vestigial-fields comment (the real floor is `protectedRecentTurns`). The drift is #288's omission, but a PR whose whole point is re-valuing these presets is the natural place to fix at least the two tables.

**4. Genuine-OpenAI cached reads are billed at Anthropic's 0.10x.** OpenAI bills cached input at 0.5x (gpt-4o: $1.25 cached vs $2.50) with no write premium. When a real gpt-4o reports `cached_tokens`, the new code undercharges the cached portion 5x. It's bounded — opposite sign and comparable magnitude to the cache-blind 1.0x it replaces, the write path is inert for OpenAI, and the litellm header masks it on non-streaming gateway calls — but direct-OpenAI and streaming paths always use the estimate. Fix is small and contained: family-dependent read multiplier (0.5x when the GPT rate card is used, 0.10x for the Anthropic family).

## Recommended, non-blocking

- **START_HERE Skills precedence**: "method and presentation" plausibly covers the circuit-breaker rule and "never poll or wait in a loop" — a sloppy Skill could claim precedence over backpressure discipline. One clause closes it, e.g. "Precedence never extends to the Quality rules or circuit-breaker handling."
- **`parseProviderCost` accepts `Inf`**: `strconv.ParseFloat("Inf", 64)` succeeds and `+Inf < 0` is false, so a garbage header yields `CostUSD = +Inf`. Add `math.IsInf`/`IsNaN` guards.
- **`chatStreamDisabled` cache args are always zero**: `bedrockStreamChunk.Usage` has no cache fields and `message_start` isn't parsed. No callers today, but if revived it will look cache-aware while staying cache-blind. Comment it or parse `message_start`.
- **Four `zaptest.NewLogger` registry sites remain in `multi_agent_broadcast_test.go`** (lines 120, 226, 349, 433). Not reachable from the background workers today (the leaked goroutine's only registry path, `agentDisplayName` → `GetAgentInfo`, doesn't log) — but switching them is cheaper than re-deriving that proof next time.

## Nits

- Comment in `anthropicFallbackPricing` says check "opus-4-1 before opus-4" but the code's case is `claude-opus`.
- Legacy ids (`claude-3-opus`, `claude-3-haiku`, `claude-3-5-haiku`, `claude-3-7-sonnet`) match no substring and still fall through to the GPT card. Retired models, no worse than before.
- The body's $2.49 figure only reproduces if the old code was *also* applying GPT-4o rates — cache-blind at Claude rates is $3.02. The two bugs are conflated across sections.
- The td-data-stats measurement (19/20) isn't reproducible from the repo; fine as motivation, but say it's external.
- The Verification section omits `-race` and omits `./pkg/server` — the very package whose `-race` failure motivated the logger fix.
- 1.25x is the 5-minute-TTL write rate (1-hour writes bill 2x). Fine today — loom only sends `cache_control: ephemeral` — worth knowing if that changes.

## Follow-ups worth ticketing (out of scope here)

- Workflow sub-agent workers are cancel-without-join everywhere, including production `Shutdown` (`multi_agent.go:1933-1936`). A `done` channel would give tests and server shutdown a deterministic wait, making harnesses like this logger unnecessary.
- `pkg/llm/azureopenai` is a copy of the openai client and remains cache-blind (2-arg `calculateCost`).
- Narrow config-compat note: a config with `workload_profile: data_intensive` plus an explicit `warning_threshold_percent` in [80,89] was valid before and now fails `Validate()` → Warn-log → silent fallback to balanced. No shipped YAML hits it; worth a release-note line.

ilsiepotamus (dismissed):
Four independently-correct fixes bundled under a thin unifying theme ("values that existed but weren't being read"). Verified each by hand rather than trusting the description:

- `marksLocked()`'s fallback correctly reverts to hardcoded defaults for all three edge cases (release≤0, release≥start, start>100) — read the guard clause directly.
- `applyPenalty` floors at 1; no path to a negative or zero mark downstream.
- Cache-tier cost math is correct for both provider shapes: OpenAI-compatible correctly subtracts cache buckets from `prompt_tokens` (which includes them), Bedrock/Anthropic correctly adds them (`input_tokens` excludes them) — checked against `cost_cache_test.go`'s real 20-turn benchmark numbers by hand.
- Substring pricing match order (`claude-opus-4-1` before `claude-opus`) is implemented correctly; no collision for the sonnet cases.
- The test-logger race fix: mutex correctly guards both the write-check and the `finished` flip, and `t.Cleanup` LIFO ordering is satisfied.

Build clean, gofmt clean, `-race` passes across every touched package.

Shape note (not blocking): this is genuinely four unrelated fixes in one PR — relief-mark wiring, ROM precedence, cache-aware cost, and a test-infra race fix. Each is small and independently correct, so I'm not asking for a split retroactively, but worth keeping in mind for future PRs since they'd have been easier to review/revert independently.

One coordination item: #329 (also yours) touches the same LLM cost-calculation and compression-profile territory. #329's cost-calc slice looks superseded by this PR — recommend landing this one and dropping the overlapping part of #329 rather than carrying both. Flagging that on #329's review too.

Approving.

mrg-ray:
Pushed `00847d02` and updated the description — this closes the four items from the first review pass.

**4 (code).** OpenAI bills a cached input token at 0.5x with no separate write bucket; Anthropic bills a read at 0.10x and a 5-minute write at 1.25x. The estimate applied Anthropic's multipliers to every family, so a genuine gpt-4o reporting `cached_tokens` undercharged the cached portion five-fold — masked by the gateway header on proxied non-streaming calls, but live on direct-OpenAI and on every streaming path. The multipliers are now named constants selected by family, decided by `anthropicFallbackPricing` as a pure id matcher so it still identifies a Claude priced from the catalog rather than the fallback card. `TestCalculateCost_CacheMultipliersFollowTheRateCard` pins both families and fails if gpt-4o is priced with Anthropic's numbers.

**3 (docs).** All four pages now carry this PR's marks — data_intensive 80/45, balanced 90/60, conversational 92/70 — and say what the two percentages are rather than leaving 'warning' and 'critical' unexplained. `memory-systems.md` also claimed `minL1Messages` was the only floor; corrected to `protectedRecentTurns`, with a note that `minL1Messages` gates nothing.

**1 (body).** The description now credits #288 with wiring `marksLocked()` to the profile and states what this PR adds: the presets were a flat 60/90, so the selector was inert while explicit overrides worked.

**2 (body).** Dropped the claim that the code records why the header can't serve; the sentence now just states the streaming behaviour, since no such comment exists.

`go build`, `go vet`, and `-race` on pkg/llm/openai and pkg/agent all clean. The non-blocking items from that review (Inf/NaN guard on `parseProviderCost`, the START_HERE precedence clause, the remaining `zaptest` sites, azureopenai's cache-blind copy) are untouched here.

mrg-ray:
Synced with `main` (`6004893c`) — it was 158 commits behind. @ilsiepotamus could you re-review? The push for the four review items dismissed your approval, and the merge adds a second commit on top.

Three conflicts, all in `pkg/llm/openai/client.go`, all places where main refactored around lines this branch also touched. Each resolution keeps both intents:

- `convertResponse` — main added finish_reason extraction, this branch added the `providerCostUSD` parameter. Both kept.
- `calculateCost` — main made the catalog namespace configurable (`c.catalogProvider`), this branch added the cache-tier parameters. Main's lookup now sits inside this branch's signature.
- `callAPI` — main moved request construction, headers and the rate-limited send into `sendRequest` with the EOF retry. Main's body kept in full; this branch's only change here was returning `http.Header` so the caller can read litellm's cost header, which is preserved.

Two of main's newer tests called the pre-branch signatures (`client_catalog_provider_test.go`, `reasoning_content_test.go`, `client_test.go`); I updated the call sites without changing what they assert.

`go build`, `go vet`, `gofmt` clean, and `-race` green on `pkg/llm/openai`, `pkg/agent` and `pkg/server` on the merge result.

PR #301, opened 2026-08-07
fix(bedrock): isolate SDK tool mappings per request

Ed-Jaramillo:
## Summary
- Remove the shared mutable tool-name map from the Bedrock Anthropic SDK client.
- Carry sanitized-to-original tool mappings within each `Chat` and `ChatStream` request.
- Cover colliding names, concurrent `Chat` calls, nil fallback behavior, and streaming `tool_use` events.

## Problem
Workflow agents can share one `SDKClient`. Concurrent requests currently reset, write, and read the same map around network calls, which can cause `fatal error: concurrent map writes` or restore a tool name using another request's mapping.

A mutex around individual map operations would prevent the runtime panic but would not preserve request identity. Request-local state fixes both failure modes without serializing LLM calls.

## Scope
This PR covers the Anthropic SDK-backed Bedrock client only. Canonical provider identity and the direct Converse client are separate concerns.

## Validation
- `go test -tags fts5 -race ./pkg/llm/bedrock -run 'TestSDKClient_(ToolNameMap_RequestLocal|ChatStream_ToolNameMapRequestLocal|ConcurrentChat_ToolNames)$' -count=10`
- `go test -tags fts5 -race ./pkg/llm/bedrock -count=1`
- `go vet -tags fts5 ./pkg/llm/bedrock`


ilsiepotamus (approved):
Verified both failure modes independently — the bug is real. `SDKClient` instances are genuinely shared across concurrent requests (provider pool / registry default via `Registry.createLLMProvider`, `pkg/agent/registry.go:1111-1126`), so the struct-level `toolNameMap` was (1) an unguarded concurrent map write — a crash under load and a guaranteed `-race` report — and (2) a silent cross-request contamination bug: the reassignment window spans the network call, so request B's map could replace A's before A's response was decoded, reverse-mapping A's tool_use names with B's mapping and potentially dispatching the wrong tool. Request-local state is the right design here; a mutex would have fixed the crash but not the contamination.

The fix is complete within its scope: the struct field is gone with zero remaining references, and no other shared mutable state remains on `SDKClient`. Tests are excellent — the deterministic interleaving test pins the wrong-name failure mode, the SSE test covers `ChatStream`, and the concurrent-Chat test (2 goroutines × 25 iterations) trips `-race` on the old code.

Follow-up to track (explicitly out of scope here, but zero-tolerance on races applies): the same shared-map pattern exists in six sibling providers — `bedrock/client.go:54` (Converse; `converse_stream.go:148` even writes it mid-stream), `anthropic/client.go:85`, `azureopenai/client.go:70`, `gemini/client.go:48`, `ollama/client.go:51`, `openai/client.go:52`. This PR is the template for fixing them; let's file an issue.

I've updated the branch to current main (clean merge); approving — merge once the re-run checks are green.

PR #302, opened 2026-08-07
fix(bedrock): use canonical provider identity

Ed-Jaramillo:
## Summary
- Return canonical `bedrock` identity from the Anthropic SDK-backed Bedrock client.
- Add a Registry regression proving a pipeline workflow registers its coordinator and colon-namespaced executor.
- Require model-switch responses to report the canonical provider name.

## Problem
`SDKClient.Name()` returns `bedrock-sdk`. The workflow loader copies that runtime name into generated agent configurations, but agent validation accepts the configured provider name `bedrock`, not `bedrock-sdk`. Startup and hot reload therefore discard every generated workflow agent before the TUI can discover namespaced sub-agents.

The regression uses a pipeline workflow, so it does not depend on the fork-join or parallel loader PRs.

## Scope
This PR changes provider identity only. Request-local tool-map concurrency is handled separately in #301.

## Validation
- `go test -tags fts5 -race ./pkg/agent -run '^TestRegistry_LoadWorkflows_BedrockSDKProviderUsesCanonicalIdentity$' -count=10`
- `go test -tags fts5 -race ./pkg/llm/bedrock -count=1`
- `go test -tags fts5 -race ./test/e2e -run '^$' -count=1`
- `go vet -tags fts5 ./pkg/agent ./pkg/llm/bedrock`


ilsiepotamus (approved):
Verified the full identity chain on main — the bug is real and user-visible on the default Bedrock/Claude path. `NewClientForModel` (`pkg/llm/bedrock/routing.go:46`) routes Claude models to the SDK client, which self-reported `"bedrock-sdk"`; the workflow loader copies `llmProvider.Name()` into generated coordinator/sub-agent configs (`pkg/agent/config_loader.go:1474,1601`); and the `ValidateAgentConfig` whitelist only accepts `"bedrock"` — so `LoadWorkflows` and the hot-reload watcher silently discarded every generated workflow agent. Even past validation, `"bedrock-sdk"` would have broken the provider factory, fallen to the 8192-token conservative context limit instead of 200K (`pkg/agent/model_context_limits.go`), and escaped OTel `gen_ai.system` normalization.

Fixing `Name()` at the source is the correct layer — an alias in the validation whitelist would have left all of the downstream consumers broken — and it aligns with the Converse client, which already returns `"bedrock"`. Grepped the repo: only the method, its test, and an e2e comment referenced the old string; nothing routes, prices, or configures on it, so no regression surface. The regression test pins the exact failure mode without needing AWS credentials.

Minor note, no action needed: SDK vs Converse can no longer be distinguished via `Name()`/span attributes. Nothing depends on that, and identity was the wrong channel for it — if operators ever need it, a construction-time log line would do.

Independent of #301 (non-overlapping hunks, merges cleanly in either order). I've updated the branch to current main (clean merge); approving — merge once the re-run checks are green.

PR #306, opened 2026-08-10
fix(agent): correct temporal grounding leak into user messages

bsahitya:
## Description

Relative phrases like "today", "yesterday", or "this month" need the model to know each user turn's arrival time. The original implementation supplied that by prepending a `[Mon 2006-01-02 15:04 MST] ` stamp directly onto the user-visible `Content` field — leaking a timestamp prefix into every persisted/displayed user message.

The first cut of this PR fixed the leak by moving the anchor into the system prompt/ROM slot. Review (round 1) showed that regressed two downstream consumers and froze the model's "now" at session creation. This PR now adopts the reviewer's suggested direction: keep stored `Content` verbatim, and render the arrival `Timestamp` into the compiled LLM view only — the same "pure render condition" seam the context compiler already uses for tool offload/eviction.

## Type of Change

- 🐛 Bug fix (non-breaking change which fixes an issue)

## Changes Made

- **`chat()` (`agent.go`)**: `Content` is stored verbatim; arrival time is captured in the existing `Timestamp` field. No stamp is prepended to the persisted body.
- **`renderLocked` (`context_compilation.go`)**: user turns render `"[Mon 2006-01-02 15:04 MST] " + Content` into the compiled view only, guarding a zero `Timestamp` for legacy rows. Rendered in UTC so a server timezone change across restart cannot alter the bytes; `Timestamp` is write-once and persisted as Unix seconds, so every compile — including after restart — is byte-identical and all cache breakpoints hold. The newest user turn always supplies current time, so inter-turn gaps stay visible.
- **Multimodal turns**: providers build the request exclusively from `ContentBlocks` when present, so the stamp is also applied to the leading text block (`stampLeadingTextBlock`, deep-copying so stored blocks stay verbatim). Without this, image turns would silently lose temporal grounding — a gap that existed even in the pre-bug body-prepend code.
- **`getSystemPrompt` (`agent.go`)**: dropped the per-session `CURRENT DATE AND TIME` ROM anchor. It froze the model's "now" at session creation (warm vs DB-restored asymmetry) and its minute-granularity value defeated cross-session ROM prompt-cache reuse. With per-turn stamps it is redundant.
- **Graph memory extractor (`graph_memory_extractor.go`)**: `buildConversationBlock` renders each turn's own arrival date (`[user @ 2026-08-10]`), and `currentDate` is derived from the newest turn's durable `Timestamp` (`newestMessageDate`), falling back to a content-embedded date, then wall clock. A session resumed after N days now anchors relative phrases to the conversation date, not extraction day — fixing a silent regression of the event-date anchoring feature (migration 000006).

## Review items addressed (round 1)

- **F1** (event-date anchoring lost its date source): fixed — extractor anchors on the durable `Timestamp` and renders per-turn dates.
- **F2** (model's "now" froze at session creation, warm-vs-restored asymmetry): fixed — per-turn stamp in the compiled view; ROM anchor removed.
- **F3** (minute-granularity ROM stamp defeated cross-session cache reuse): fixed — ROM anchor dropped.
- **F4** (midnight flake): fixed — the date assertion accepts either day in a `dateBefore`/`dateAfter` band around the call.
- **F5** (nit): the relevant helper usage was replaced with `storedTurns`.

## Testing

```
go test -tags fts5 -race ./pkg/agent/...
go build -tags fts5 ./...
go vet -tags fts5 ./pkg/agent/
```

All pass; 0 race conditions; gofmt clean.

### Test Coverage

- Compiled user turn is stamped while stored `Content` stays verbatim (`TestChat_CompiledUserTurnCarriesTimestamp`, `TestCompile_UserTurnRendersArrivalStampInViewOnly`).
- Multimodal turn stamps the leading text block, stored blocks verbatim (`TestCompile_MultimodalUserTurnStampsLeadingTextBlock`).
- Two consecutive compiles are byte-identical (`TestCompile_TwoCompilesAreByteIdentical`).
- Legacy timestamp-less rows render verbatim (`TestCompile_UserTurnWithoutTimestampRendersVerbatim`).
- Extractor anchors to backdated `Timestamp`s and renders per-turn dates (`TestNewestMessageDate_PrefersDurableTimestamp`, `TestBuildConversationBlock_RendersPerTurnDates`).
- Cross-session restart test updated to match the stamped LLM view.

## Proto Changes

- No proto changes in this PR

## Documentation

- Code comments added/updated

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -1194,15 +1194,14 @@ func (a *Agent) getSystemPrompt(ctx context.Context) string {
 		basePrompt = `Use available tools to help the user accomplish their goals. Never fabricate data - only report what tools actually return.`
 	}
 
-	// Anchor the model in wall-clock time so relative phrases ("today",
-	// "this month", "yesterday") resolve to real dates. This lives in the
-	// system prompt — NOT prepended to user message Content — so it never
-	// leaks into the user-visible, persisted message body (clients render
-	// Content verbatim). Like the rest of the ROM slot, this is rendered once
-	// at session creation and is byte-stable for the session, so it carries
-	// date granularity (sufficient for temporal grounding); a session that
-	// spans midnight keeps its creation-time anchor rather than a live clock.
-	basePrompt = "CURRENT DATE AND TIME: " + time.Now().Format("Monday, 2006-01-02 15:04 MST") + "\n\n" + basePrompt
+	// Temporal grounding does NOT live here. A wall-clock anchor baked into the
+	// ROM slot would freeze the model's "now" at session creation (warm sessions
+	// span days; a DB-restored session would re-anchor to a different instant),
+	// and — because ROM has its own cross-session cache breakpoint — a
+	// per-session timestamp would defeat prompt-cache reuse across an agent's
+	// sessions. Instead, each user turn carries its arrival time, rendered into
+	// the compiled view only (see renderLocked): the newest user turn always
+	// supplies current time, and inter-turn gaps stay visible.
 
 	// Inject task context (current tasks, ready front, board stats).
 	// Rendered once into ROM at session creation — the ROM slot is
@@ -1835,9 +1834,10 @@ func (a *Agent) chat(ctx context.Context, sessionID string, userMessage string, 
 	// Content is the canonical, user-visible message body: it is persisted and
 	// returned verbatim to clients.
 	// Do NOT prepend a timestamp here — that leaks a "[Mon 2006-01-02 15:04 MST]"
-	// prefix into every displayed user message. Arrival time is already captured
-	// durably in the Timestamp field; per-turn temporal grounding ("today",
-	// "this month") belongs in the system prompt / ROM, not baked into the body.
+	// prefix into every displayed user message. Arrival time is captured durably
+	// in the Timestamp field; per-turn temporal grounding ("today", "this month")
+	// is restored by rendering that Timestamp into the compiled view only
+	// (renderLocked), never into the stored body.
 	userMsg := a.appendMessage(ctx, session, Message{
 		Role:          "user",
 		Content:       userMessage,
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -1194,6 +1194,16 @@ func (a *Agent) getSystemPrompt(ctx context.Context) string {
 		basePrompt = `Use available tools to help the user accomplish their goals. Never fabricate data - only report what tools actually return.`
 	}
 
+	// Anchor the model in wall-clock time so relative phrases ("today",
+	// "this month", "yesterday") resolve to real dates. This lives in the
+	// system prompt — NOT prepended to user message Content — so it never
+	// leaks into the user-visible, persisted message body (clients render
+	// Content verbatim). Like the rest of the ROM slot, this is rendered once
+	// at session creation and is byte-stable for the session, so it carries
+	// date granularity (sufficient for temporal grounding); a session that
+	// spans midnight keeps its creation-time anchor rather than a live clock.
+	basePrompt = "CURRENT DATE AND TIME: " + time.Now().Format("Monday, 2006-01-02 15:04 MST") + "\n\n" + basePrompt
+
 	// Inject task context (current tasks, ready front, board stats).
 	// Rendered once into ROM at session creation — the ROM slot is
 	// byte-stable for the session, so this is a snapshot, not a live view.
@@ -1822,13 +1832,15 @@ func (a *Agent) chat(ctx context.Context, sessionID string, userMessage string, 
 	// This is the Chat()-entry persist site — the only turn-incrementing event
 	// (HLD §4.5) — hence turnStart=true.
 	//
-	// Time enters the session here, written into the turn at arrival: temporal
-	// words ("today", "this month") resolve at utterance time, and a value
-	// written once is durable content like any other row — the whole session
-	// stays byte-stable. Nothing renders time dynamically anywhere.
+	// Content is the canonical, user-visible message body: it is persisted and
+	// returned verbatim to clients.
+	// Do NOT prepend a timestamp here — that leaks a "[Mon 2006-01-02 15:04 MST]"
+	// prefix into every displayed user message. Arrival time is already captured
+	// durably in the Timestamp field; per-turn temporal grounding ("today",
+	// "this month") belongs in the system prompt / ROM, not baked into the body.
 	userMsg := a.appendMessage(ctx, session, Message{
 		Role:          "user",
-		Content:       time.Now().Format("[Mon 2006-01-02 15:04 MST] ") + userMessage,
+		Content:       userMessage,
 		ContentBlocks: p.contentBlocks,
 		AgentID:       a.id, // Track which agent received this message
 		Timestamp:     time.Now(),
--- a/pkg/agent/chat_content_blocks_test.go
+++ b/pkg/agent/chat_content_blocks_test.go
@@ -341,4 +341,3 @@ func TestSystemPrompt_CarriesCurrentDate(t *testing.T) {
 	require.Contains(t, prompt, "CURRENT DATE AND TIME:", "system prompt anchors the model in wall-clock time")
 	require.Contains(t, prompt, time.Now().Format("2006-01-02"), "the anchor carries today's date")
 }
-
--- a/pkg/agent/chat_content_blocks_test.go
+++ b/pkg/agent/chat_content_blocks_test.go
@@ -95,8 +95,7 @@ func sampleBlocks() []ContentBlock {
 }
 
 // findUserMessage returns the first user-role message whose content ends with
-// the given text — the turn carries its arrival timestamp as a prefix, so the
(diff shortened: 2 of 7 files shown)


ilsiepotamus (changes requested):
# Review — round 1 (head b03a3a38)

The diagnosis is right and half the fix is right: prepending the arrival stamp to the persisted, user-visible `Content` was a presentation leak, and storing `Content` verbatim (with `Timestamp` as the durable arrival record) is the correct persistence change — the updated tests pin that well. But the PR fixes the leak by deleting the model-visible temporal signal instead of relocating it, and two things break silently downstream. Requesting changes on F1 + F3; F2 needs at least a conscious decision.

Verified locally at head `b03a3a38` in an isolated worktree: `gofmt` clean, `go build -tags fts5 ./...` OK, the four touched tests pass with `-tags fts5 -race`; CI green. `getSystemPrompt` is indeed only rendered at session create/restore (`pkg/agent/memory.go:371,598`), and `Timestamp` is genuinely persisted (`pkg/agent/session_store.go:641`) — the PR's two central claims hold.

---

## F1 — major: event-date anchoring silently loses its date source

`extractDateFromMessages` (`pkg/agent/graph_memory_extractor.go:493`) scans message `Content` for `\d{4}[/-]\d{2}[/-]\d{2}` to anchor the extraction prompt's `Current date:` header (`graph_memory_extractor.go:300`). The removed `[Mon 2006-01-02 …]` prefix was its guaranteed match; with bare content it returns empty and falls back to `time.Now()` **at extraction time**. The extractor LLM also previously saw each turn's arrival stamp inline in the conversation block; `buildConversationBlock` now renders undated text under a single extraction-day header.

Failure case: a session resumed after N days — relative phrases in older turns ("yesterday", said three days ago) get anchored to extraction day, writing wrong `event_date` rows into graph memory. This regresses the event-date anchoring feature (migration 000006) without any test noticing, because same-day extraction is indistinguishable.

The fix is small and doesn't re-leak anything: the `pkg/types.Message` the extractor receives already carries `Timestamp` (`pkg/types/types.go:152`) through `GetRecentConversationTurns` — render it per message in `buildConversationBlock` (e.g. `%d. [%s @ 2026-08-10]: %s`) and derive `currentDate` from the newest message's `Timestamp` instead of regex-scanning `Content`. That's strictly more robust than the old prefix-scan.

## F2 — major (design): the model's "now" freezes at session creation, inconsistently

Pre-PR the model saw an arrival stamp on every user turn. Post-PR the only time signal is the creation-time ROM stamp:

- A **warm** session held in memory for days keeps its stale anchor — the comment's "a session that spans midnight" undersells this; Loom sessions are durable and long-lived.
- A **DB-restored** session gets a *fresh* anchor via `ensureSessionMemory` (`pkg/agent/memory.go:598`) — so warm and restored sessions answer "what day is it?" differently.
- Inter-turn gaps are invisible ("deploy is tomorrow" on Monday; "did it happen?" asked Thursday — the model can't tell three days passed).
- Fold/L2 summaries can no longer carry dates the model never saw.

See "Suggested direction" below for a fix that keeps `Content` verbatim *and* restores per-turn grounding. If that's deferred, the ROM comment should state the real staleness window and the warm-vs-restored asymmetry.

## F3 — medium: minute-granularity stamp contradicts its own comment and defeats cross-session ROM cache reuse

ROM is a dedicated system block with its own cache breakpoint (`pkg/llm/anthropic/client.go:238-265`) precisely so its prefix caches. Pre-PR, all sessions of an agent shared identical ROM bytes → cross-session prompt-cache hits. Now every session's ROM is unique to its creation **minute**.

The new comment says the anchor "carries date granularity (sufficient for temporal grounding)" — but the format is `"Monday, 2006-01-02 15:04 MST"`, which is minute granularity. Beyond the cache cost, embedding a wall-clock time in a static prompt teaches the model a confidently-wrong clock hours into a session.

If the anchor stays (see F2/suggested direction for whether it should), drop `15:04 MST`: `CURRENT DATE: Monday, 2006-01-02`. That matches the comment, restores same-day cross-session cache reuse, and removes the stale-clock hazard.

## F4 — minor: midnight flake in `TestSystemPrompt_CarriesCurrentDate`

`getSystemPrompt` and the assertion each call `time.Now()` independently; straddling midnight flakes the test. Capture the date before and after the render and accept either.

## F5 — nit

`findUserMessage` still matches by `HasSuffix`; with verbatim content it can assert equality. Fine to leave.

---

## Suggested direction: relocate the stamp to compile-time render, don't delete it

The codebase already has the exact seam. `compileLocked` (`pkg/agent/context_compilation.go:60`) builds the LLM view at every call and routes each row through `renderLocked` (`context_compilation.go:171`), whose stated principle is that offload is *"a pure render condition"* — stored rows never change, only their rendered view. The arrival stamp is the same kind of thing:

1. Keep this PR's `chat()` change — `Content` verbatim, `Timestamp` populated. ✔ (already done)
2. In `renderLocked`, add a user-row case (guarding `!m.Timestamp.IsZero()` for legacy rows): render `"[" + m.Timestamp.Format("Mon 2006-01-02 15:04 MST") + "] " + m.Content` into the **compiled view only**. `Timestamp` is write-once and persisted as Unix seconds, so every compile — including after restart — renders identical bytes: byte-stability and all cache breakpoints are preserved. The model sees exactly what it saw before this PR; clients see clean `Content`.
3. Fix the extractor from `Timestamp` (F1 above).
4. Drop the `CURRENT DATE AND TIME` ROM anchor, or keep it date-only — with per-turn stamps the newest user turn always carries current time, so the anchor is redundant and only costs cache reuse (F3).

Suggested tests: (a) capturing-LLM sees the stamped user turn while stored `Content` is verbatim; (b) two consecutive compiles are byte-identical; (c) extractor with backdated `Timestamp`s anchors to the conversation's date, not today.

Known trade-offs of the render approach, for the record: ~10 uncounted tokens per user turn in the L1 estimate (negligible), and `time.Unix` restores in the server's local zone, so a TZ config change across restart alters rendered bytes (cache is cold after restart anyway; render in UTC if strictness is wanted).

---

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

bsahitya:
Thanks for the thorough round-1 review — I've addressed all of the comments.

Adopted the suggested direction: stored `Content` stays verbatim, and the arrival `Timestamp` is rendered into the compiled LLM view only (`renderLocked`), the same "pure render condition" seam used for tool offload/eviction.

- **F1** — extractor now anchors on the durable `Timestamp` (`newestMessageDate`) and renders each turn's own date in `buildConversationBlock`, so a session resumed after N days anchors to the conversation date, not extraction day.
- **F2** — per-turn stamp restores grounding in the compiled view; the ROM anchor (which froze "now" at session creation and diverged warm vs DB-restored) is removed.
- **F3** — dropped the minute-granularity `CURRENT DATE AND TIME` ROM anchor, restoring cross-session ROM prompt-cache reuse.
- **F4** — the date assertion now accepts either day via a `dateBefore`/`dateAfter` band, so it can't flake across midnight.
- **F5** — replaced the `HasSuffix` helper usage with `storedTurns` for verbatim assertions.

Also handled multimodal turns: providers build the request from `ContentBlocks` when present, so the stamp is applied to the leading text block too (`stampLeadingTextBlock`, deep-copying so stored blocks stay verbatim) — closing a grounding gap that predated this PR.

Verification: `go build -tags fts5 ./...`, `go vet -tags fts5 ./pkg/agent/`, and `go test -tags fts5 -race ./pkg/agent/...` all pass; gofmt clean; 0 race conditions.

ilsiepotamus (approved):
Round 2 (234e6c17) resolves everything from round 1 — approving. This landed exactly where we discussed: stored `Content` verbatim, arrival `Timestamp` rendered into the compiled view only in `renderLocked`, the same pure-render seam the offload/eviction logic uses.

Checklist against the round-1 findings:

- **F1** ✅ `newestMessageDate` anchors extraction on the newest turn's durable `Timestamp`, and `buildConversationBlock` renders each turn's own arrival date — backdated-timestamp tests cover both.
- **F2** ✅ ROM anchor removed; the newest user turn always supplies current arrival time, so warm and DB-restored sessions ground identically.
- **F3** ✅ No per-session timestamp in the ROM slot; cross-session prompt-cache reuse restored, pinned by the `NotContains` assertion.
- **F4** ✅ dateBefore/dateAfter band — can't flake across midnight.
- **F5** ✅ `storedTurns` asserts against actual persisted rows.

The multimodal `stampLeadingTextBlock` is a nice catch beyond the review scope — providers do build the request from `ContentBlocks` when present, so the stamp would otherwise never reach the model on image turns; copying the slice keeps stored blocks verbatim.

Independently verified the byte-stability premise: `Timestamp` persists as Unix seconds and every restore path rehydrates it (session_store.go:633/757), so compiled stamps survive restarts, and the minute-granularity format is immune to the seconds truncation. Byte-stable, cache-safe.

One non-blocking note, no action needed here: `newestMessageDate` now takes precedence over `extractDateFromMessages`, so replayed histories whose content carries historical dates (benchmark harnesses) will anchor to ingestion time unless the replay writes backdated Timestamps. That's the harness's responsibility, and the new precedence is strictly more correct for live sessions.

PR #308, opened 2026-08-11
feat(server): occurred_at arrival-time override for replayed/imported conversations

ilsiepotamus:
## Description

Adds `WeaveRequest.occurred_at` — a proto-first arrival-time override so replayed or imported conversations anchor temporal grounding at their historical time instead of the ingestion wall clock.

### Motivation

PR #306 (approved) anchors graph-memory extraction on the durable `Message.Timestamp` and renders per-turn arrival dates. That is strictly more correct for live sessions, but replayed histories — e.g. the LongMemEval harness, which ingests haystack sessions over gRPC with "took place on YYYY/MM/DD" in the content — get `Timestamp = time.Now()` at ingestion, so post-#306 they would anchor to the benchmark run day and every turn would carry a run-day date contradicting the inline historical date. This PR makes historical arrival time a first-class capability of the public API instead of a content-scanning hack.

### What's included

**Proto (law first)**
- `google.protobuf.Timestamp occurred_at = 11` on `WeaveRequest` — `buf lint`, `buf breaking`, `buf generate` all clean

**Server gate** (`pkg/server/occurred_at.go`)
- `applyOccurredAt` wired into `Weave` + `StreamWeave` on both `Server` and `MultiAgentServer`
- Rejected with `FAILED_PRECONDITION` unless the operator sets `server.allow_time_override: true` (default **false** — client-supplied timestamps can poison temporal grounding, so honoring them is an explicit operator decision)
- Future-dated values beyond a 5-minute clock-skew allowance rejected with `INVALID_ARGUMENT`

**Agent** (`pkg/agent/occurred_at.go`)
- `WithOccurredAt` context override applied centrally in `appendMessage`: the user turn, assistant reply, and tool rows persisted during the call all carry the historical time, so extraction anchoring (`newestMessageDate`), per-turn dates, and compiled-view arrival stamps stay mutually consistent

**Config**
- `server.allow_time_override` (viper default false), wired in `cmd_serve` via `SetAllowTimeOverride`, documented in the example config

**Docs**
- `docs/reference/streaming.md`: field + gate semantics
- `docs/reference/CLAUDE.md`: corrected the `WeaveRequest`/`WeaveResponse` listings and client example, which documented fields that don't exist in the real proto (`message`, `tokens_used`, `resp.Message`)

### Tests

8 new tests, all passing with `-tags fts5 -race`:
- Gating default (reject with `FAILED_PRECONDITION`, agent never runs) — both servers
- Future timestamp rejection (`INVALID_ARGUMENT`)
- Accepted path anchors user + assistant rows at the override — both servers
- No-override path keeps wall-clock timestamps (gate enabled but field unset)
- Agent-level: `WithOccurredAt` anchors all rows; absence keeps wall clock

### Verification

- [x] `just check` passes ("All checks passed! (matches GitHub CI)")
- [x] `buf breaking --against origin/main` — no breaking changes
- [x] `gofmt -l` clean
- [x] New tests run with `-race`, 0 race conditions

### Follow-up (separate branch)

The LongMemEval harness (`feat/longmemeval-benchmark`) should pass each haystack session's date as `occurred_at` and enable `server.allow_time_override` in its benchmark deployment — that closes the post-#306 anchoring gap for replayed histories.

### DIFF
--- a/cmd/looms/cmd_serve.go
+++ b/cmd/looms/cmd_serve.go
@@ -2341,6 +2341,13 @@ func runServe(cmd *cobra.Command, args []string) {
 	// Set logger for server operations
 	loomService.SetLogger(logger)
 
+	// Honor WeaveRequest.occurred_at only when the operator opted in —
+	// replay/import arrival times are rejected otherwise (see applyOccurredAt).
+	loomService.SetAllowTimeOverride(config.Server.AllowTimeOverride)
+	if config.Server.AllowTimeOverride {
+		logger.Info("Arrival-time override enabled (server.allow_time_override): WeaveRequest.occurred_at will be honored")
+	}
+
 	// Set provider factory for dynamic model switching
 	loomService.SetProviderFactory(providerFactory)
 	logger.Info("Provider factory configured on server for model switching")
--- a/cmd/looms/config.go
+++ b/cmd/looms/config.go
@@ -197,15 +197,16 @@ type TUIConfig struct {
 
 // ServerConfig holds server-specific configuration.
 type ServerConfig struct {
-	Port             int                 `mapstructure:"port"`
-	Host             string              `mapstructure:"host"`
-	HTTPPort         int                 `mapstructure:"http_port"` // HTTP/REST+SSE port (default: 5006, 0=disabled)
-	EnableReflection bool                `mapstructure:"enable_reflection"`
-	InsecureAdmin    bool                `mapstructure:"insecure_admin"` // Allow admin endpoints without LOOM_ADMIN_TOKEN (default: false)
-	TLS              TLSConfig           `mapstructure:"tls"`
-	Clarification    ClarificationConfig `mapstructure:"clarification"` // Clarification question timeouts
-	CORS             CORSServerConfig    `mapstructure:"cors"`          // CORS configuration for HTTP endpoints
-	Auth             AuthConfig          `mapstructure:"auth"`          // Endpoint authentication (Supabase JWT)
+	Port              int                 `mapstructure:"port"`
+	Host              string              `mapstructure:"host"`
+	HTTPPort          int                 `mapstructure:"http_port"` // HTTP/REST+SSE port (default: 5006, 0=disabled)
+	EnableReflection  bool                `mapstructure:"enable_reflection"`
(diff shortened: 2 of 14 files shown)


bsahitya (commented):
Solid, proto-first change that fixes a real post-#306 gap for replayed/imported histories. The design is clean: a single context override (`WithOccurredAt`) applied centrally in `appendMessage`, so user/assistant/tool/sidecar rows persisted during the call all anchor consistently. The security posture is right too — opt-in gate defaulting to false, `FAILED_PRECONDITION` when disabled (agent never runs), `INVALID_ARGUMENT` for future-dated values, and the gate wired identically into `Weave`/`StreamWeave` on both `Server` and `MultiAgentServer`. Docs correcting the stale `WeaveRequest`/`WeaveResponse` listings in `docs/reference/CLAUDE.md` is a nice bonus.

A few non-blocking notes:

1. **Tool-row anchoring is claimed but not directly tested.** Both the PR description and the proto/doc comments state tool rows carry the override, and the code supports it since every `appendMessage` call site (agent.go:2503, :2573, :2609, :2630, :2816, :2864, :2875; hygiene.go:74) flows through the same seam. But the mock LLMs in the tests emit no tool calls, so all four accepted-path tests only assert `user`/`assistant` rows. A single test with a tool-calling mock would lock the strongest part of the claim against regressions.

2. **Uniform stamping collapses multi-turn replays.** Since the override applies to *every* row persisted during the call, replaying a multi-turn session in one `Weave` would flatten all turns to a single `occurred_at`. That's fine for the intended one-session-per-call LongMemEval usage, but worth a one-line doc note so callers don't assume per-turn fidelity within a single call.

3. **No lower bound on `occurred_at`.** Future values are rejected, but a zero/epoch or year-1900 timestamp is accepted. Probably acceptable for import, though a sanity floor could catch obviously-malformed clients.

4. **`occurredAtMaxSkew` (5m) is a fixed const.** Reasonable default and well-commented; flagging only in case operators with looser clock sync would want it configurable later.

Verification claims (`just check`, `buf breaking` clean, `-race`) and the `occurred_at = 11` field numbering look correct. None of the above blocks merge.

bsahitya (approved):

PR #309, opened 2026-08-11
fix(security): guard tainted path values directly with filepath.IsLocal (CodeQL #666–#670)

ilsiepotamus:
## Summary

Triage + fix for the five open CodeQL alerts on `main`:

| Alert | Rule | Location | Assessment |
|---|---|---|---|
| #666–#668 | `go/path-injection` | `pkg/artifacts/session_metadata.go` (`WriteSessionArtifactMetadata`: `MkdirAll`/`CreateTemp`/`Rename`) | Not exploitable — false positive |
| #669 | `go/path-injection` | `pkg/artifacts/session_metadata.go` (`ReadSessionArtifactMetadata`: `ReadFile`) | Not exploitable — false positive |
| #670 | `go/zipslip` | `pkg/server/skills_import.go` (`extractZipToTempDir`) | Not exploitable — false positive |

All flagged paths were already gated: `ValidateSessionID` enforces a strict `[A-Za-z0-9._-]` allowlist (no separators, no `..`) on every session-metadata path, and the zip extractor had `Clean` + `IsAbs` + `Rel`-prefix containment, a fresh temp dir target, no symlink materialization, and a 64MB decompression cap. Hostile inputs are covered by existing tests.

## Why the alerts stayed open (root cause)

The 2026-07 hardening placed the analyzer-visible check on the wrong variable. Verified against CodeQL's query sources ([TaintedPathCustomizations.qll](https://github.com/github/codeql/blob/main/go/ql/lib/semmle/go/security/TaintedPathCustomizations.qll), [ZipSlipCustomizations.qll](https://github.com/github/codeql/blob/main/go/ql/lib/semmle/go/security/ZipSlipCustomizations.qll)):

- `filepath.IsLocal` **is** modeled (`IsLocalCheck`), but a barrier guard sanitizes only the exact expression it checks. Both sites guarded a derived `rel` value while the tainted `root`/`dest` flowed to the sinks.
- The zip check `strings.HasPrefix(rel, ".."+string(filepath.Separator))` fails `DotDotCheck` (requires a string literal) and has the opposite polarity of `PrefixCheck` (which models the `HasPrefix(path, safeDir)` idiom).

## Fix

Guard the tainted value itself, before it is joined:

- `SessionArtifactsRoot`: `filepath.IsLocal(sessionID)` — one shared choke point closes #666–#669.
- `extractZipToTempDir`: `filepath.IsLocal(name)` right after `filepath.Clean`, replacing the `IsAbs` + `Rel`/`HasPrefix` block it strictly subsumes — closes #670.

Semantics are equal-or-stricter: the only newly rejected inputs are Windows drive-relative paths (`C:x`) and reserved device names (`NUL`).

## Tests & docs

- `TestAddSkill_RejectsHostileZip` is now table-driven: parent traversal, absolute path, nested traversal, bare `..`.
- Existing `TestSessionArtifactsRoot_rejectsBadPath` hostile-ID table unchanged and green.
- `just check` passes (lint + full `-tags fts5 -race` suite + build; gosec 0 issues).
- `docs/reference/security-scan-triage.md`: corrected the 2026-07-23 entry (it claimed #666–#668 were fixed; they weren't) and recorded this triage with the guard-placement lesson.

**Verification**: this PR's CodeQL check should report all five alerts as fixed; they auto-close on merge to `main`.

### DIFF
--- a/docs/reference/security-scan-triage.md
+++ b/docs/reference/security-scan-triage.md
@@ -1,9 +1,9 @@
-# Security Scan Triage — Snyk Findings (2026-07)
+# Security Scan Triage
 
-Triage record for the Snyk SAST/SCA findings reported against `teradata-labs/loom:main`
-via ArmorCode as of 2026-07-22. Each class of finding was verified against the code
-before being classified. This document exists so future scans don't re-litigate the
-same findings from scratch.
+Triage record for security-scanner findings against `teradata-labs/loom:main` —
+Snyk SAST/SCA via ArmorCode (2026-07-22) and GitHub CodeQL (2026-07/08). Each class
+of finding was verified against the code before being classified. This document
+exists so future scans don't re-litigate the same findings from scratch.
 
 ## Fixed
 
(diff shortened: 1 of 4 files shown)


bsahitya (approved):
Reviewed the diff and cross-checked against the CodeQL query sources cited in the description.

**Correctness of the fix**
- `SessionArtifactsRoot`: guarding `filepath.IsLocal(sessionID)` on the raw input (rather than a `filepath.Rel`-derived value) is the right call — CodeQL's `IsLocalCheck` barrier only credits the exact guarded expression, so the earlier `rel`-based check never sanitized the `root` value reaching the sink. The new guard is equal-or-stricter than the prior allowlist + Rel containment.
- `extractZipToTempDir`: replacing the `IsAbs` + `Rel`/`HasPrefix` block with `filepath.IsLocal(name)` right after `filepath.Clean` is a clean subsumption. Verified the cleaned-name cases: `..` -> stays `..` (rejected), `skill/../../etc` -> `../etc` (rejected), `/etc/escape.txt` -> absolute (rejected). The `name == "." || name == ""` early-continue still guards the local-but-empty case before the IsLocal check.

**Nits / non-blocking**
- Semantics change is accurately scoped in the description (newly rejects Windows drive-relative `C:x` and reserved device names like `NUL`) — acceptable for these paths.
- Confirmed `strings` remains imported/used elsewhere in `skills_import.go`, so removing the `HasPrefix` line doesn't leave a dangling import.

**Tests**
- Table-driven `TestAddSkill_RejectsHostileZip` now covers parent traversal, absolute, nested traversal, and bare `..` — good coverage of the escape vectors.

The triage doc correction (owning that the 2026-07 guards didn't actually close the alerts) is a good record. LGTM.

PR #310, opened 2026-08-11
fix(agent): load all canonical workflow pattern types in registry

ilsiepotamus:
## Description

Follow-up to #307, which fixed the fork-join instance of a broader defect: the registry loader (`loadOrchestrationWorkflow`) accepted only a subset of the canonical `spec.type` vocabulary, so workflows that `looms workflow validate` reported **valid** were **silently skipped** by the agent registry at startup and on reload (logged at ERROR, execution continued). After #307, three canonical types were still affected:

- **`iterative`** — no loader case at all
- **`swarm`** — no loader case at all (the shipped example `technology-swarm.yaml` validated but never loaded)
- **`conditional`** — the loader read `branches[*].agent_id`, but canonical branches are *nested workflow patterns* (`convertConditionalPattern` recurses into them); canonical files died with "no agents found in workflow". The loader also never collected `condition_agent_id`, so even under its old wrong schema the condition agent was never registered.

The loader also accepted **`iterative_pipeline`** — the exact mirror image of the `fork_join` bug: a loader-only spelling the validator and canonical converter reject, never usable end to end.

## Type of Change

- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation update (shipped example YAMLs corrected)
- [ ] 🔧 Refactoring
- [x] 🧪 Test update

### Loader (`pkg/agent/config_loader.go`)

- Replaced the per-type agent-ID switch with `workflowAgentIDs`, a recursive collector covering all 7 canonical types (`debate`, `fork-join`, `pipeline`, `parallel`, `conditional`, `iterative`, `swarm`).
- `conditional`: recurses into `branches` and `default_branch` (nested patterns of any type, including nested conditionals), collects `condition_agent_id`, iterates branches in sorted order so registration order is deterministic, and de-duplicates agents shared across branches.
- `iterative`: collects agents from the base pipeline under `spec.pipeline`.
- `swarm`: `agent_ids` plus optional `judge_agent_id`.
- `debate`: moderator is now de-duplicated when it also appears in `agent_ids` (previously produced a duplicate namespaced sub-agent config).
- Dropped `iterative_pipeline` from the case label. All remaining `iterative_pipeline` occurrences in the repo are internal executor/checkpoint identifiers, not YAML `spec.type` values — the same removal rationale as #307's `fork_join`.
- Exact error strings for the existing `parallel` and `fork_join` regression tests are preserved.

### Vocabulary contract (the part that keeps this class of bug from recurring)

- `validation.CanonicalWorkflowPatternTypes()` is now the single exported source of truth for the canonical vocabulary. The validator and `pkg/orchestration`'s converter guard (`validateSpec`) both consume it — previously three hand-maintained copies existed (`validator.go:385`, `workflow_config.go:167`, and the loader switch).
- `TestLoadWorkflowAgents_CanonicalPatternContract` pins the loader to it: for every canonical type, a minimal canonical spec must pass `pkg/validation` **and** load through `LoadWorkflowAgents` with the expected coordinator + namespaced sub-agents. The table is asserted with `ElementsMatch` against `CanonicalWorkflowPatternTypes()`, so adding a type to the vocabulary without teaching the loader (and the table) fails the test.

### Validator allowlists (`pkg/validation/validator.go`)

- `conditional` and `swarm` now allow `spec.retry_policy`, which their converters parse (`parseOutputRetryPolicy` calls at `workflow_config.go:512` and `:690`) and which `workflow-all-fields-reference.yaml` documents. Previously the reference file's own conditional/swarm sections failed schema validation. (Pipeline retry is per-stage and unaffected.)

### Shipped examples — all orchestration examples now validate AND load

- `complexity-routing.yaml`: rewritten to canonical conditional with nested branch patterns. Its "nested workflows (branches with sub-workflows) are planned for future releases" note was stale — `conditional_executor.go:137` executes the selected branch via `orchestrator.ExecutePattern` and the converter has recursed into branches all along. The old file (`type: conditional` + bare `agent_ids`) failed the validator, the converter, and the loader.
- `scheduled-workflows/{daily-report,frequent-monitor,hourly-sync,weekly-backup}.yaml`: flattened the non-canonical `spec.pipeline.{initial_prompt,stages}` nesting to canonical `spec.initial_prompt` / `spec.stages`, and renamed stage `prompt:` to `prompt_template:`. These four never validated or loaded. (`heartbeat-new.yaml` / `heartbeat-resume.yaml` were already canonical.)
- `orchestration-patterns/README.md`: conditional section now shows the canonical shape; stale limitation note removed.
- `pkg/validation/real_workflows_test.go`: all five example paths in this sweep were stale (files moved into `orchestration-patterns/`, two examples deleted), so **every subtest silently skipped** via `t.Skipf`. Paths fixed, missing files now `t.Fatalf` so path rot can't disable the sweep again, and the conditional + swarm examples were added to it.
- `TestLoadWorkflowAgents_ShippedOrchestrationExamples`: loads every YAML under `examples/reference/workflows/orchestration-patterns/` through the registry loader (13 files). This is the test that caught the four broken scheduled examples.

## Out of scope (verified, deliberately not fixed here)

**Weaver's `apply_template` emits `spec.schedule`, but the scheduler reads top-level `schedule`.** `buildWorkflowYAMLFromTemplate` (`agent_management_templates.go:437`) writes the suggested schedule under `spec`, while `pkg/scheduler/loader.go:120` reads `config.Schedule` (top-level, matching `WorkflowConfig.Schedule` and the shipped scheduled examples) and the validator allowlists reject `spec.schedule`. Three shipped templates set `Schedulable: true`, so template-emitted schedules are currently invisible to the scheduler. That's a Weaver emitter fix with its own regression test — separate PR.

## Related Issues

Follow-up to #307 (announced in its review). No tracking issue; root-cause analysis is inline above.

## Testing

- `TestLoadWorkflowAgents_CanonicalPatternContract` — all 7 canonical types validate + load (fails on main for `iterative`, `swarm`, `conditional`)
- `TestLoadWorkflowAgents_ConditionalNestedPatterns` — nested conditional recursion + cross-branch de-duplication
- `TestLoadWorkflowAgents_ConditionalRejectsMalformedSpecs` — strict errors for missing/malformed branches and default_branch
- `TestLoadWorkflowAgents_LoaderOnlySpellingsRejected` — `iterative_pipeline` rejected
- `TestLoadWorkflowAgents_ShippedOrchestrationExamples` — every shipped orchestration example loads (fails on main for 6 of 13 files)
- Existing regression tests preserved and passing: `TestLoadWorkflowAgents_ForkJoinPatternType` (#307), `TestLoadWorkflowAgents_ParallelTasks` / `_ParallelRejectsMalformedTasks` (#298), exact error strings unchanged
- Full `just check` (lint + `go test -tags fts5 -race ./...` + build) green locally

### Test Checklist

- [x] All new and existing tests pass
- [x] Race detector shows zero race conditions (`-race` on the full suite via `just check`)
- [x] All checks pass (`just check`)

### Test Coverage

## Proto Changes

## Documentation

- [ ] README.md updated (top-level: not needed)
- [x] Shipped example YAMLs + `orchestration-patterns/README.md` corrected to canonical shapes
- [x] No `docs/` changes needed — the hyphenated canonical vocabulary was already the documented behavior; this makes the loader match it

## Security

## Performance

- [x] No performance impact (loader runs once per file at startup/reload)

## Deployment Notes

Workflows using `spec.type: iterative_pipeline` are now rejected by the loader. Like #307's `fork_join`, that spelling was never accepted by the validator or the canonical converter, so no working configuration existed to migrate. Users with old-format conditional files (`branches[*].agent_id`) now get a clear error (`conditional branch "x": workflow pattern spec missing 'type' field`) instead of a silent partial load that dropped the condition agent.

---

### DIFF
--- a/pkg/agent/config_loader.go
+++ b/pkg/agent/config_loader.go
@@ -1602,8 +1602,15 @@ func (c *workflowAgentCollector) collect(spec map[string]interface{}) error {
 		c.addPipelineStages(pipelineSpec)
 
 	case "conditional":
-		if conditionAgent, ok := spec["condition_agent_id"].(string); ok {
-			c.add(conditionAgent)
+		// convertConditionalPattern requires both fields; rejecting them here
+		// keeps the registry from registering a workflow that can never run.
+		conditionAgent, ok := spec["condition_agent_id"].(string)
+		if !ok || strings.TrimSpace(conditionAgent) == "" {
+			return fmt.Errorf("conditional workflow requires non-empty 'condition_agent_id'")
+		}
+		c.add(conditionAgent)
+		if prompt, ok := spec["condition_prompt"].(string); !ok || strings.TrimSpace(prompt) == "" {
+			return fmt.Errorf("conditional workflow requires non-empty 'condition_prompt'")
 		}
 		branches, ok := spec["branches"].(map[string]interface{})
 		if !ok {
--- a/pkg/agent/config_loader.go
+++ b/pkg/agent/config_loader.go
@@ -11,6 +11,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	loomv1 "github.com/teradata-labs/loom/gen/go/loom/v1"
@@ -1410,83 +1411,11 @@ func loadOrchestrationWorkflow(path string, data map[string]interface{}, llmProv
 		return nil, fmt.Errorf("orchestration workflow missing 'spec.type' (or use spec.entrypoint for agent-reference workflows)")
 	}
 
-	// Extract agent IDs based on pattern type
-	var agentIDs []string
-	switch patternType {
-	case "debate":
-		if ids, ok := spec["agent_ids"].([]interface{}); ok {
-			for _, id := range ids {
-				if idStr, ok := id.(string); ok {
-					agentIDs = append(agentIDs, idStr)
-				}
-			}
-		}
-		// Include optional moderator
-		if moderator, ok := spec["moderator_agent_id"].(string); ok && moderator != "" {
-			agentIDs = append(agentIDs, moderator)
-		}
-
-	case "fork-join":
-		if ids, ok := spec["agent_ids"].([]interface{}); ok {
-			for _, id := range ids {
-				if idStr, ok := id.(string); ok {
-					agentIDs = append(agentIDs, idStr)
-				}
-			}
-		}
-
-	case "parallel":
-		tasks, ok := spec["tasks"].([]interface{})
-		if !ok {
-			return nil, fmt.Errorf("parallel workflow requires 'spec.tasks' array")
-		}
-
-		seen := make(map[string]bool)
-		for i, task := range tasks {
-			taskMap, ok := task.(map[string]interface{})
-			if !ok {
-				return nil, fmt.Errorf("parallel workflow task %d must be an object", i)
-			}
-
-			agentID, ok := taskMap["agent_id"].(string)
-			if !ok || strings.TrimSpace(agentID) == "" {
-				return nil, fmt.Errorf("parallel workflow task %d missing non-empty 'agent_id'", i)
-			}
-
-			if !seen[agentID] {
-				agentIDs = append(agentIDs, agentID)
-				seen[agentID] = true
-			}
-		}
-
-	case "pipeline", "iterative_pipeline":
-		if stages, ok := spec["stages"].([]interface{}); ok {
-			seen := make(map[string]bool)
-			for _, stage := range stages {
-				if stageMap, ok := stage.(map[string]interface{}); ok {
-					if agentID, ok := stageMap["agent_id"].(string); ok && !seen[agentID] {
-						agentIDs = append(agentIDs, agentID)
-						seen[agentID] = true
-					}
-				}
-			}
-		}
-
-	case "conditional":
-		if branches, ok := spec["branches"].(map[string]interface{}); ok {
-			seen := make(map[string]bool)
-			for _, branch := range branches {
-				if branchMap, ok := branch.(map[string]interface{}); ok {
-					if agentID, ok := branchMap["agent_id"].(string); ok && !seen[agentID] {
-						agentIDs = append(agentIDs, agentID)
-						seen[agentID] = true
-					}
-				}
-			}
-		}
-
-	default:
-		return nil, fmt.Errorf("unsupported workflow pattern type: %s", patternType)
+	// Collect every agent ID the workflow references, recursing into nested
+	// patterns (conditional branches, iterative base pipelines).
+	agentIDs, err := workflowAgentIDs(spec)
+	if err != nil {
+		return nil, err
 	}
 
 	if len(agentIDs) == 0 {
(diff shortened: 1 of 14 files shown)


pkg/agent/config_loader_workflow_test.go
```diff
@@ -361,3 +362,403 @@ spec:
...
+		}
+		return nil
+	})
+	require.NoError(t, err, "examples directory missing (update the path)")
+	require.NotEmpty(t, files)
+
+	for _, file := range files {
+		t.Run(filepath.Base(file), func(t *testing.T) {
+			configs, err := LoadWorkflowAgents(file, &mockLLMProvider{})
+			require.NoError(t, err, "shipped example failed to load in the registry: %s", file)
```
Ed-Jaramillo, line 750: Could this test, or a companion integration test, also exercise `orchestration.LoadWorkflowFromYAML`? `LoadWorkflowAgents` only proves that agent IDs can be registered. It currently passes `technology-swarm.yaml`, while the user-facing `looms workflow validate` command rejects that file because `spec.question` is missing.

pkg/agent/config_loader.go
```diff
@@ -1574,6 +1503,145 @@ Sub-agents: %s`,
...
+		c.addPipelineStages(spec)
+
+	case "iterative":
+		pipelineSpec, ok := spec["pipeline"].(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("iterative workflow requires 'spec.pipeline' object")
+		}
+		c.addPipelineStages(pipelineSpec)
+
+	case "conditional":
```
Ed-Jaramillo, line 1604: Could we validate that `condition_agent_id` and `condition_prompt` are present strings here (and ideally non-empty)? `convertConditionalPattern` rejects either field when it is absent or not a string, but the collector currently accepts a conditional with valid branches when either field is missing. That allows the registry to register a workflow that later fails during execution.

Ed-Jaramillo (changes requested):
Thank you for the opportunity to review this PR. The recursive agent collection, deterministic branch ordering, de-duplication, and the canonical pattern-type contract test are thoughtful improvements.

I found one blocking gap in the shipped-example coverage. `TestLoadWorkflowAgents_ShippedOrchestrationExamples` verifies that the registry can collect agent IDs, but not that the workflows can be parsed by the canonical converter (`orchestration.LoadWorkflowFromYAML`) — the path `looms workflow validate` uses. On this branch, 12 of 13 examples pass runtime validation, but `technology-swarm.yaml` fails:

    invalid workflow structure: swarm pattern requires 'question' field

Repro:

    looms workflow validate examples/reference/workflows/orchestration-patterns/technology-swarm.yaml

This is the same class of drift this PR fixes for `spec.type`: the vocabulary is now unified, but required-field rules still live separately in `pkg/validation` (which accepts this file) and the converter (which rejects it). Could we add the required `spec.question` to that example and add converter coverage for all shipped examples, so a workflow can't be registry-loadable but not runnable?

Once that gap is addressed, the overall approach looks good.

---
Reviewed using Claude (Fable) and GitHub Copilot as assistants — findings were cross-checked between them and verified locally (tests run on both this branch and main) before writing this up.

pkg/validation/real_workflows_test.go
```diff
@@ -52,9 +64,10 @@ func TestRealWorkflowExamples(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			fullPath := filepath.Join(projectRoot, tt.path)
 
-			// Check if file exists
+			// A moved or deleted example must fail loudly: silent skips let
+			// every subtest here rot away unnoticed when examples/ moved.
 			if _, err := os.Stat(fullPath); os.IsNotExist(err) {
-				t.Skipf("Workflow file not found: %s", fullPath)
+				t.Fatalf("Workflow file not found (update the path): %s", fullPath)
 				return
```
Ed-Jaramillo: Nit: `t.Fatalf` stops the test goroutine, so this `return` is unreachable and can be removed (same for the second occurrence below).

pkg/agent/config_loader_workflow_test.go
ilsiepotamus: Added `TestLoadWorkflowFromYAML_ShippedOrchestrationExamples` in `pkg/orchestration` (0a68e7ed) — it cannot live next to this sweep because `orchestration` imports `agent`, so the agent-side test cannot import the converter without an import cycle. Red-verified against your repro: with the example fix stashed, it fails with exactly `swarm pattern requires 'question' field`. The validator now also requires `question` for swarm, so `looms workflow validate` catches this class directly.

pkg/agent/config_loader.go
ilsiepotamus: Done in 0a68e7ed — the collector now rejects missing or empty `condition_agent_id` / `condition_prompt` with dedicated errors, and rejection rows were added to `TestLoadWorkflowAgents_ConditionalRejectsMalformedSpecs`. One deliberate asymmetry to flag: the converter accepts a present-but-empty string for these fields, while the loader (and the validator presence checks) treat empty as missing, since an empty condition agent ID can never resolve at execution time. Happy to relax to presence-only if you would rather keep exact converter parity.

pkg/validation/real_workflows_test.go
ilsiepotamus: Removed — both occurrences (0a68e7ed).

ilsiepotamus:
All three findings addressed in 0a68e7ed. Thanks — the swarm example gap was a genuine miss, and it's the same drift class this PR targets, one seam over: the *vocabulary* is now unified, but the *required-field* rules also lived in two places (validator vs converter). Round 2 closes both sides:

**1. `technology-swarm.yaml` registry-loadable but not runnable (blocking)**
- Added the required `question: '{{input}}'` to the example.
- New `TestLoadWorkflowFromYAML_ShippedOrchestrationExamples` in `pkg/orchestration` parses every shipped orchestration example through the canonical converter (the `looms workflow validate` / `ExecuteWorkflow` path). Verified red-then-green: without the example fix it fails with exactly your repro error (`swarm pattern requires 'question' field`). It lives in `pkg/orchestration` rather than next to the registry sweep because `orchestration` imports `agent` — the agent-side test can't import the converter without a cycle.
- Root cause fixed, not just the instance: `validatePatternSpecificFields` now enforces the converter's required fields for the three patterns it didn't cover — swarm (`question`, `agent_ids`), conditional (`condition_agent_id`, `condition_prompt`, `branches`), iterative (`pipeline`) — so `looms workflow validate` rejects what the converter rejects. `TestOrchestrationRequiredFieldsMatchConverter` pins each check.

**2. Conditional collector accepts missing condition fields**
- The loader now requires non-empty `condition_agent_id` and `condition_prompt` (`conditional workflow requires non-empty '...'`), with rejection rows added to `TestLoadWorkflowAgents_ConditionalRejectsMalformedSpecs`. One deliberate asymmetry to flag: the converter accepts a *present-but-empty* string for these fields; the loader (and now the validator's presence checks) treat empty as missing, since an empty condition agent ID can never resolve at execution time. Strictly that makes the loader marginally stricter than the converter for garbage-only inputs — happy to relax to presence-only if you'd rather keep them exactly aligned.

**3. Unreachable `return` after `t.Fatalf`** — both removed.

Full `just check` (lint + `-race` suite + build) green locally on the updated branch.

Ed-Jaramillo (approved):
All three findings verified fixed on my side — re-ran `looms workflow validate` across all 13 examples (all pass), confirmed the new converter sweep goes red without the example fix, and the loader/validation tests pass with `-race`. The root-cause fix (validator required-fields matching the converter, pinned by test) goes beyond what I asked for — nice.

On the strictness asymmetry: keep the loader strict. An empty `condition_agent_id` can never resolve at execution time, so rejecting it at load is the correct behavior; I wouldn't relax it to match the converter's leniency.

---
Reviewed using Claude (Fable) and GitHub Copilot as assistants; findings verified locally before writing this up.

Ed-Jaramillo (approved):
Thank you for addressing the review feedback. I verified the runtime-loader coverage, swarm example fix, conditional field validation, and dead-code cleanup. These changes resolve my concerns. Approved.

PR #311, opened 2026-08-11
fix(server): join background worker goroutines on shutdown (data race)

ilsiepotamus:
## Summary

Fixes the flaky data race that failed the merge-ref CI runs of #309 (first occurrence) and this PR's own round 1 (second occurrence) — same disease, two sibling subsystems:

1. [Run 31527032022](https://github.com/teradata-labs/loom/actions/runs/31527032022/job/93897772366): `TestWovenWorkflow_ReplyReachesCoordinator` passed, then the **workflow coordinator notification handler's** deferred exit log raced `testing.tRunner`.
2. [Run 31529136666](https://github.com/teradata-labs/loom/actions/runs/31529136666/job/93904656538): `TestSpawnSubAgent_BackgroundContextCarriesTenantIdentity` passed, then the **spawned-agent monitor's** exit log (`spawn_agent.go:337`) raced the same way.

## Root cause (one pattern)

`MultiAgentServer` runs detached background goroutines that log through `s.logger` on their way out, and every stop path is a fire-and-forget context cancel — nothing joins them before the logger's owner (a test's zaptest sink, or the process logger at shutdown) goes away. Whack-a-mole was inevitable, so this PR joins the whole class:

| Goroutine | Spawn site |
|---|---|
| Message-queue monitor | `StartMessageQueueMonitor` |
| Coordinator notification handler | `spawnWorkflowSubAgents` |
| `runWorkflowSubAgent` | named spawn + auto-spawn |
| `runSubAgentBroadcastHandler` / `runCoordinatorBroadcastHandler` | `spawnWorkflowSubAgents` |
| `monitorSpawnedAgent` / `runSpawnedAgentLoop` | `SpawnSubAgent` |
| MCP tool re-indexers (self-expire ≤10s) | `AddMCPServer` / `DeleteMCPServer` |

The remaining `go` statements in `pkg/server` are request-joined (progress streaming, workflow execution, health-check pings into buffered channels) and don't log after their handler returns.

## Fix (revised in round 2)

- **`goWorker(name, fn)`** is now the only way to start a tracked background worker: it registers the goroutine with `backgroundWorkerWG` (`Done` after each worker's deferred exit log) and **refuses admission once shutdown has begun**, so the join is a stable lifecycle barrier — a request racing shutdown cannot register a worker behind it. Every spawn site stores its cancel func in a tracked map *before* calling `goWorker`, so any admitted worker is visible to the cancel sweep.
- **`ShutdownBackgroundWorkers(ctx) error`** is the one exported lifecycle operation (round 1's `WaitBackgroundWorkers` + unexported cancel are gone): closes admission, cancels every tracked worker context (workflow message + broadcast, spawned-agent monitor + loop), joins with the caller's deadline, and returns `ctx.Err()` on timeout. Production can now cancel everything it waits for.
- **Refusal semantics**: MCP re-index and workflow handler spawns are skipped (debug-logged); `SpawnSubAgent` racing shutdown fails cleanly (`cleanupSpawnedAgent` + error) instead of creating an unmonitored agent; auto-spawn returns an error the monitor logs.
- **Tests**: `setupBroadcastTestServer` (the shared fixture behind the workflow-reply, broadcast, spawn-task, and turn-identity tests) registers one `ShutdownBackgroundWorkers(context.Background())` cleanup; `t.Cleanup` LIFO ordering still guarantees test-registered cleanups (e.g. `stopMonitor`) → cancel+join workers → close stores.
- **`looms serve`**: `cancelMonitor()` then `ShutdownBackgroundWorkers` with a 10s deadline — a worker mid-LLM-call may lag; shutdown warns and continues rather than hanging.

No behavior change outside shutdown ordering, plus the new clean-failure path for spawns racing shutdown.

## Verification

- `TestShutdownBackgroundWorkers_ActiveSpawnedAgent`: an active spawned agent (subscribed → monitor + message loop running) plus the queue monitor, through the production `cancelMonitor()` → `ShutdownBackgroundWorkers` sequence: clean join in ~2s, no timeout (round 1's blocking finding).
- `TestShutdownBackgroundWorkers_ClosesAdmission`: post-shutdown `goWorker` is refused; a racing `SpawnSubAgent` fails cleanly and leaves nothing tracked.
- `TestGoWorker_AdmissionShutdownRace`: 8 goroutines × 50 admissions hammering `goWorker` against `ShutdownBackgroundWorkers` under `-race` — every worker either joined or refused, no third outcome.
- Full `pkg/server` `-race` pass: clean.
- `just check` green (lint + full race suite + build).


cmd/looms/cmd_serve.go
```diff
@@ -3475,6 +3475,22 @@ func runServe(cmd *cobra.Command, args []string) {
 		cancelMonitor()
 		logger.Info("Message queue monitor cancelled")
 
+		// Join background worker goroutines (queue monitor, workflow
+		// notification/broadcast handlers, spawned-agent monitors, MCP
+		// re-indexers) so none log after the logger is torn down. Bounded: a
+		// worker mid-LLM-call may take a moment to observe cancellation.
+		workersDone := make(chan struct{})
+		go func() {
+			loomService.WaitBackgroundWorkers()
```
Ed-Jaramillo: **Blocking:** `WaitBackgroundWorkers` includes the spawned-agent monitor and message-loop goroutines, but `cancelMonitor()` only causes the queue monitor to cancel entries in `workflowSubAgents`. The only new code that cancels `spawnedAgents` is the unexported `cancelBackgroundWorkers`, so this package cannot satisfy `WaitBackgroundWorkers`' documented "cancel first" requirement.

With any active spawned agent, this wait therefore reaches the ten-second timeout and shutdown proceeds with that worker still alive. Please expose a production cancel-and-join operation, or otherwise cancel the spawned-agent contexts, before waiting here.

_AI-assisted review: drafted with GitHub Copilot and independently verified against the PR diff and relevant code paths._

cmd/looms/cmd_serve.go
```diff
@@ -3475,6 +3475,22 @@ func runServe(cmd *cobra.Command, args []string) {
...
+			close(workersDone)
+		}()
+		select {
+		case <-workersDone:
+			logger.Info("Background workers stopped")
+		case <-time.After(10 * time.Second):
```
Ed-Jaramillo: **Question:** Background coordinator/broadcast turns intentionally drop worker cancellation, and that behavior is documented and test-pinned. If one of those turns exceeds ten seconds, shutdown proceeds while it is still active. Is that the intended production policy? I am not requesting a change to the cancellation invariant, only confirmation that this bound is acceptable.

_AI-assisted review: drafted with GitHub Copilot and independently verified against the PR diff and relevant code paths._

cmd/looms/cmd_serve.go
```diff
@@ -3475,6 +3475,22 @@ func runServe(cmd *cobra.Command, args []string) {
...
+			logger.Warn("Timed out waiting for background workers; continuing shutdown")
+		}
+
 		// Stop HTTP server
```
Ed-Jaramillo, line 3493: **Non-blocking:** HTTP remains active here, and gRPC is stopped substantially later, so request paths can still reach `backgroundWorkerWG.Add` while or after the join runs. The queue monitor usually keeps the counter above zero during the initial wait, which lowers the practical risk, but this is not a stable admission barrier. Consider stopping admission, or guarding worker registration with shutdown state, before joining.

_AI-assisted review: drafted with GitHub Copilot and independently verified against the PR diff and relevant code paths._

Ed-Jaramillo (changes requested):
I reviewed the complete five-file diff, traced each tracked worker from spawn through cancellation and teardown, and checked the two linked race-detector failures. The required CI checks are currently green.

## What this PR gets right

The reported test race is real, and the test-side fix is correct. Detached workers log during deferred cleanup, so cancellation without a join allows `zaptest` to write through `testing.T` after the test has completed. Registering the detached-worker family in a shared `sync.WaitGroup`, with `Done` occurring after each worker's deferred exit logging, addresses that root cause. The shared fixture's LIFO cleanup order also correctly stops explicit monitors, cancels remaining workers, joins them, and only then closes stores and the test logger.

The five-file scope is appropriate. `multi_agent.go` owns lifecycle accounting, `spawn_agent.go` and `mcp_management.go` contain related spawn sites, the shared broadcast fixture owns test teardown, and `cmd_serve.go` owns the production logger lifetime. I found no unrelated net changes or hidden refactor.

## Blocking production issue

Production cannot cancel every worker that it now waits for. `WaitBackgroundWorkers` includes spawned-agent monitors and message loops, but `runServe` only calls `cancelMonitor()`. That cancellation causes the queue monitor to sweep `workflowSubAgents`; it does not cancel `spawnedAgents`. The only new sweep covering spawned agents is `cancelBackgroundWorkers`, which is unexported and therefore unavailable to `cmd/looms`.

As a result, any active spawned agent keeps the wait-group counter nonzero. SIGTERM then consumes the full ten-second timeout, logs a warning, and continues dependency teardown with the worker still alive. This is a new user-visible shutdown delay, and the wait does not deliver the lifecycle guarantee described by its documentation.

Please expose a production-safe cancel-and-join operation (or export the required cancellation step), call it after `cancelMonitor()`, and verify the production sequence with an active spawned agent.

## Additional feedback

Worker admission also remains open during the join because HTTP and gRPC stop later. I consider that Medium/non-blocking for this PR: the queue monitor normally keeps the counter positive during the early wait, but the ordering still permits a request to register a worker after the join has begun or completed. Closing admission before joining would make the lifecycle boundary reliable.

The intentionally uncancelable in-flight `Chat` contexts are not a defect here: comments document that policy and tests pin it. I have left a question asking whether proceeding after ten seconds is the intended production tradeoff for a turn that is still running.

## Missing coverage

Please add a focused test that starts an active spawned agent, runs the production-style cancel/join sequence, and proves shutdown completes without the ten-second timeout and without workers surviving into dependency teardown. A concurrent admission-versus-shutdown test would also be useful, but is not independently blocking.

## Verdict

**Request changes.** The test-race fix and scope are sound, but production must be able to cancel the spawned-agent workers it waits for before this is safe to merge.

_AI-assisted review: drafted with GitHub Copilot and independently verified against the PR diff, worker lifecycle paths, linked race logs, and current CI results._

cmd/looms/cmd_serve.go
ilsiepotamus: Fixed in 12daaec3. Round 1's `WaitBackgroundWorkers` + unexported cancel pair is replaced by one exported lifecycle operation, **`ShutdownBackgroundWorkers(ctx) error`**: it closes worker admission, cancels every tracked worker context (workflow message + broadcast, spawned-agent monitor + loop), and joins with the caller's deadline, returning `ctx.Err()` on timeout. `runServe` calls it with a 10s context right after `cancelMonitor()`, so production now cancels exactly the set of workers it waits for.

The production sequence with an active spawned agent is pinned by `TestShutdownBackgroundWorkers_ActiveSpawnedAgent`: queue monitor running, a subscribed spawn (monitor + message loop live), then `cancelMonitor()` → `ShutdownBackgroundWorkers` under a 5s deadline. It joins in ~2s; the round-1 sequence rode out the full 10s on this exact scenario.

cmd/looms/cmd_serve.go
ilsiepotamus: Yes — intended. At SIGTERM the real choice is between abandoning an in-flight uncancelable turn after a bounded wait, or blocking shutdown on an LLM call of unbounded latency — and an orchestrator (or a k8s termination grace period) SIGKILLs eventually anyway, so "block until the turn finishes" isn't actually on the table; its honest form is "die later, mid-write, with no logs". 10s matches the other budgets in `runServe` (HTTP stop). With admission now closed before the join, the only way to reach the timeout is a genuinely long in-flight turn, and the warn log makes that visible rather than silent.

cmd/looms/cmd_serve.go
ilsiepotamus: Took this one too — it made the blocking fix cleaner rather than harder. Workers now start only through `goWorker(name, fn)`, which checks a shutdown flag under the same mutex that orders `WaitGroup.Add` against the join; `ShutdownBackgroundWorkers` sets the flag before cancelling and waiting, so admission is closed before the join begins and a request racing shutdown is refused instead of registering behind it. Every spawn site stores its cancel func in a tracked map *before* calling `goWorker`, so any admitted worker is guaranteed visible to the cancel sweep — no window where a worker is admitted but uncancelable.

Refusal semantics: MCP re-index and workflow handler spawns are skipped (debug-logged); `SpawnSubAgent` fails cleanly (`cleanupSpawnedAgent` + error) rather than creating an unmonitored agent; auto-spawn returns an error the monitor logs. Pinned by `TestShutdownBackgroundWorkers_ClosesAdmission` and `TestGoWorker_AdmissionShutdownRace` (8 goroutines × 50 admissions hammering `goWorker` against shutdown under `-race`).

pkg/server/multi_agent.go
```diff
@@ -1395,7 +1409,7 @@ func (s *MultiAgentServer) spawnWorkflowSubAgents(ctx context.Context, coordinat
 	s.logger.Info("Registered coordinator for event-driven message notifications (monitor-based)",
 		zap.String("coordinator", coordinatorID))
 
-	go func() { // #nosec G118 -- intentional: background worker goroutine that must outlive request context
+	s.goWorker("coordinator-notification-handler", func() {
```
Ed-Jaramillo, line 1412: **Non-blocking follow-up:** Could we check this return (and the similar workflow-handler calls below) and unwind staged registrations when admission is closed? After `ShutdownBackgroundWorkers`, `spawnWorkflowSubAgents` currently inserts coordinator/sub-agent entries and registers notification channels, every `goWorker` call is refused, but the function still returns `nil`. I reproduced that sequence: it returned success with two `workflowSubAgents` entries left behind.

No worker escapes the join, so this does not invalidate the race fix. Returning an error and cancelling/deleting/unregistering the staged entries would make the refusal semantics complete and avoid making a racing `Weave` look successfully initialized.

Ed-Jaramillo (approved):
Thank you for the thoughtful follow-up. The original production blocker is resolved: `ShutdownBackgroundWorkers` now closes admission under the same mutex that orders `WaitGroup.Add`, cancels both workflow and spawned-agent contexts, and `runServe` invokes it after cancelling the queue monitor. The active-spawn test exercises the production sequence, and the admission race test pins the stable join barrier.

I also re-ran the focused workflow/spawn/shutdown tests with `-tags fts5 -race -count=10`; they passed. The bounded behavior for an intentionally uncancelable in-flight turn is now explicit and reasonable.

I left one non-blocking cleanup note for workflow setup when admission is already closed. It does not invalidate the race fix or allow a worker past the join.

**Approved.**

_AI-assisted review: drafted with GitHub Copilot and independently verified against commit `12daaec`, the worker lifecycle paths, targeted race tests, and current CI results._

PR #315, opened 2026-08-16
docs: apprentice plan — turning observed work into reusable automation

Josh-Schoen:
## What

A design plan for **apprentice**: a meta-agent that watches work — human or agent — distills it into candidate skills, workflows, agents, and triggers, and proposes them for *deputizing*.

The framing that makes it fit loom: **the weaver authors from intent; the apprentice authors from evidence.** Same artifact types, same restricted write path, opposite input.

Plan only — no code. Review the shape before anyone builds against it.

## Key decisions captured

**Naming.** The working name "track task" is rejected: it collides with `pkg/task`, cloud's `cloud_tasks`, and up-tera's `tasks` feature, all of which are different from this — the task board is the feature's *input*, not its subject. "draft" is also rejected despite being the right weaving term, because `draft_skill_name` / `AdminDraftSkillCache` already mean "an unsaved skill under test in an admin session." Settled vocabulary is **apprentice / candidate / deputize**, none of which collide in loom or tera-backend.

**The task board is the trace spine, not raw spans.** It already carries ordering, dependencies, hierarchy, and an `approach` field per step. `tool_executions` and spans fill in detail for steps the agent never tracked. Since `pkg/skills/tasks` already implements skill → tasks, this feature is its inverse — and the output type is therefore `SkillTaskTemplate`, not free-form YAML.

**P0 is a round-trip oracle.** Emit a shipped skill's `TaskTemplate` onto a board, distill it back, diff structurally (step count, order, dependency edges, parameter positions). Measurable without an LLM judge, and it stays in CI permanently as a regression suite rather than being a one-off proof.

**Where it lives.** The engine is in loom because everything it produces already is. `ApprenticeService` is defined in loom proto and served from `pkg/server`, so the TUI and tera-backend are peers over one API and the cloud layer becomes a multi-tenant wrapper rather than a new surface. loom-knowledge is a phase-5 consumer for org-scale recurrence, not a host.

**The TUI is the first client.** Watch mode needs one command and one dialog; progress streaming and — critically — parameter confirmation during generalization already exist via `ProgressMultiplexer` and the `clarification` dialog the weaver uses. The full loop runs with no cloud dependency, which is where the interaction design gets settled before up-tera inherits it.

## Testing

Splits on the LLM seam: deterministic units plus the round-trip oracle in CI; recorded LLM fixtures for the generalization code path; `pkg/evals` for quality on a schedule. Stated explicitly in the doc: `just test` will not catch a worse prompt, and conflating the two tiers is how quality regressions ship unnoticed.

Flagged as the most severe risk, with required P1 coverage rather than deferral: **a captured trace is untrusted input whose text becomes a skill prompt.** Tool output, table comments, and query results flow into a candidate's `prompt.instructions` and `tools.required`, which a later session loads with tool access — potentially for a different user.

## Next

P0 (`pkg/apprentice` + the round-trip oracle) is the follow-up and needs no proto, server, or TUI work.

One open product question in the doc worth a decision during review: does deputizing a SKILL candidate go straight to an org PR, or land private-first with promotion as a separate act?

### DIFF
--- a/docs/plan-apprentice.md
+++ b/docs/plan-apprentice.md
@@ -22,9 +22,14 @@ Two things are missing:
 2. **Suggestion from observation** — opt-in, passive: notice that a procedure recurs across
    sessions, agents, and people, and offer to take it over.
 
-The second is the more valuable half. Repetition is the evidence that a procedure is worth
-automating, and it is also the evidence that makes generalization tractable — one run tells
-you what happened, three runs tell you which parts were incidental.
+Repetition is the evidence that makes generalization tractable — one run tells you what
+happened, three runs tell you which parts were incidental. That argues for the second mode.
+
+**Measured yield argues the other way at small scale.** Spike finding 6 puts a real corpus at
+roughly six distinct deputizable procedures per 284 sessions, with 43% of those sessions making
+no tool calls at all. Observe mode has to pan all of that to find six; watch mode has a human
+pointing at one. So **build watch mode first and treat observe mode as the org-scale play** — it
+becomes the more valuable half only once the corpus is too large for a human to be the filter.
 
 ---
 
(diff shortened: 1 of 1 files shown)


Josh-Schoen:
Superseded by #316, which now carries both the plan and the P0 code so reviewers see the evidence alongside the design.

For the record, the failing check here was **not** caused by this PR. This branch changed one markdown file. `FuzzSpanTimestamps` failed with `context deadline exceeded` at 30.07s against `-fuzztime=30s` — the test's internal context deadline races the fuzz budget. #316 shares the same base commit, adds actual Go code, and passes the same job in 7m8s. Filed separately as a pre-existing flake in `pkg/observability`.

PR #316, opened 2026-08-16
feat(apprentice): one-command v1 plan + P0 distiller

Josh-Schoen:
## What

**Apprentice**: turn work that actually happened into a reusable skill. The framing that makes it fit loom — **the weaver authors from intent; the apprentice authors from evidence.**

Two parts:
- `docs/plan-apprentice.md` — a deliberately small v1, ten evidence-backed constraints, and an explicit deferred list.
- `pkg/apprentice` — the P0 distiller. Working code, 99.4% coverage.

Supersedes #315, whose only failing check was a pre-existing flaky fuzz timeout unrelated to the change.

## v1 is one command

```
loom apprentice watch
```

At session end: read that session's `tool_executions` and `messages`, abstract the concrete literals into named parameters, emit **one candidate skill YAML**, show it in the TUI for review. The user edits, names, and deputizes it — or discards it.

Not in v1: observe mode, recurrence detection, workflow/agent/trigger candidates, UI-event capture, multi-tenancy, derived task boards. Each is in the Deferred section with its reason.

An earlier draft of this plan had three capture channels, four candidate kinds, two modes, and work in three repos. That was scope accretion, not design — and the prior art argues directly against it.

## Why narrow

This is a 50-year-old field, **programming by demonstration**, whose canonical text is literally titled *Watch What I Do* (Cypher et al., MIT Press 1993). Pygmalion (1975) → macro recorders → Eager (1991) → CoScripter/Vegemite → RPA → ALLOY (2025). **None of the general systems achieved broad adoption**; Lau's retrospective blames usability, not capability. What shipped was narrow: the spreadsheet macro recorder, Selenium IDE, Playwright codegen.

Failures cluster in two buckets: generalizing from one example, and brittleness of a GUI/DOM/pixel capture substrate. **The apprentice's whole bet is bucket 2** — it captures tool calls with structured inputs, and tool schemas are versioned contracts where GUIs are not.

v1 does **not** escape bucket 1. It generalizes from a single session, which is the documented failure mode, and the mitigation is a human reviewing every candidate rather than an algorithm. Stated as a risk in the doc rather than hidden.

xAI shipped **Grok Bot** on 11 Aug 2026 doing this at the GUI layer. Validation that the problem matters; the differentiator is substrate, not concept.

## Ten constraints from evidence

From building P0 and a read-only spike over a real 284-session corpus. Not features — guardrails. The load-bearing ones:

- **Tasks are an output, not an input.** The board is empty in practice because maintaining it is work an agent must *choose* to do, while tool calls are recorded automatically. `pkg/skills/hygiene` exists solely to compensate for agents that don't. Nothing reads `tool_executions` to derive tasks. Compare Claude Code's background tasks, which get created as a byproduct of noticing out-of-scope work — same data model, opposite dynamics.
- **`tool_executions` + `messages` is the only substrate that reliably exists.** 284 sessions held 4,651 messages, 2,047 tool executions, and **zero** tasks or boards (`TaskBoardConfig` defaults to `Enabled: false`).
- **Do not drop failed steps.** The best procedure in the corpus is a trace where all nine of its steps failed — for a purely environmental reason — while the surrounding flailing mostly succeeded. Intent lives in tool call *inputs*. An "only distill successful sessions" filter throws away the single most valuable candidate available.
- **Yield is a trickle.** 43% of sessions made zero tool calls; only 12% made ≥15. The 22 richest hold ~six deputizable procedures. That's why watch mode beats observe mode at single-user scale.
- **The largest corpus category is meta-work that must not be deputized** — 8 of the 22 richest sessions are weaver-authoring runs. Needs an explicit guard.
- **Recurrence detection is the crux and both naive approaches fail.** Same-procedure/different-order false-negatives on sequence alignment; same-intent/different-procedure false-positives on text similarity.

## P0 (`pkg/apprentice`)

The exact inverse of `pkg/skills/tasks`: recover a `SkillTaskTemplate` from a task board. Deterministic, no LLM. Keyed recovery is exact; topological recovery warns when the DAG leaves order unconstrained rather than presenting a guess as evidence.

Tests run the **real emitter against a real sqlite store** over every embedded skill authoring a `task_template`, and check re-emission is a fixpoint. 99.4% coverage, clean under `-tags fts5 -race -count=20`, `golangci-lint` 0 issues.

**Stated honestly in the doc:** since v1 reads `tool_executions`, P0 is *not* on v1's execution path. It keeps its value as a permanent regression suite for the emitter's template contract, as the source of two constraints, and as the design template for v1's reader — but the plan doesn't pretend it's v1's foundation.

It also surfaced a real gap: `SkillTaskTemplate.RootTitle` is documented as naming a parent task that `emitTemplate` never creates. `TestRoundTrip_UnrecoverableTemplateFields` pins it.

## Review asks

1. Is the v1 cut the right line?
2. Does deputizing go straight to an org PR, or land private-first? (Doc defaults to private-first.)
3. Separable, own PR: should `TaskBoardConfig.Enabled` default to true? Cost is bounded at 500 context tokens and the expensive behaviors are separately gated — but constraint 2 says flipping the flag doesn't make boards get *maintained*.


ilsiepotamus (approved):
Reviewed the plan, the code, and every load-bearing claim against the codebase. **Approving** — the distiller is correct, the tests are the right kind, and the plan is honest about what P0 is and isn't.

## Claims verified (not just read)

- **Edge direction is right.** The emitter writes `FromTaskID`=dependent / `ToTaskID`=dependency (`pkg/skills/tasks/emitter.go:248`), and `GetDependencies` queries `WHERE from_task_id = ?` (`pkg/storage/sqlite/task_store.go:513`) — so "each ToTaskID becomes a depends_on index" is exactly correct.
- **The `RootTitle` gap is real**: documented at `pkg/skills/types.go:277`, zero `RootTitle`/`ParentID` hits in the emitter. Pinning it so the test *starts failing* when the emitter gains the feature is the right move.
- Key format `skill:|sess:|step:`, the `skill_step` metadata channel, `DefaultMaxTasks=8`, `TaskBoardConfig` default-false (`registry.go:785`), the weaver+guide gate (`agent_management.go:188`), and the weaver-templates fan-out (steps 4 and 5 both depend on 3) — all check out.
- Ran locally: `-tags fts5 -race` green, coverage 99.4% as claimed, `gofmt` + `golangci-lint` clean.

Design choices worth calling out as *right*: warnings-not-errors for ambiguous recovery, refusing to present a topological guess as authored order, tie-breaking on creation time instead of content-hash IDs, keeping failed steps, and running the round-trip oracle against the real emitter + real sqlite instead of a fake.

## The failing check is a real bug, not flake — fix is up

The PR description says #315's failure was "a pre-existing flaky fuzz timeout." It isn't a timeout and it isn't flake: `FuzzPromptInterpolation` found a **real sanitizer bug** in `pkg/prompts`, reproducible deterministically. `sanitizePromptInjection` replaces a fence with spaces, the space collapse + trim steps eat those spaces, and remnants from adjacent interpolations recombine: value `` ````` `` escapes to `` `` ``, and `{{.a}}{{.a}}` yields `` ```` ``.

Genuinely unrelated to this diff — but it will keep failing CI stochastically on every PR until fixed, so I've fixed it in **#322** (with the CI-found input as a permanent regression seed). Once that merges, rebase/re-run here and this goes green. Would suggest editing the PR description so the "flaky timeout" framing doesn't become repo folklore.

## Two minor nits (non-blocking)

1. **Keyed recovery doesn't warn on non-contiguous step indices.** A board where one emitted task was deleted recovers as a shorter template with no signal a step went missing. A gap in indices is evidence a reviewer would want surfaced, same as the other warnings.
2. **Keyed mode can emit forward `depends_on` silently** when key order contradicts the DAG. Only reachable on tampered/foreign boards (the emitter derives both from the same template), but topological mode asserts backwards-only while keyed mode doesn't check.

## Answers to the review asks

1. **v1 cut**: right line. The 43%-zero-tool-call yield data makes observe mode indefensible at single-user scale, and constraint 9 shows recurrence fingerprinting isn't ready to build.
2. **Deputizing**: private-first, firmly. Risk 2 (captured trace is untrusted input that becomes a prompt) is the most severe risk in the feature; one unreviewed distillation must not be able to reach an org repo. Promotion already exists as a separate deliberate act.
3. **`TaskBoardConfig.Enabled`**: keep false. Constraint 2 is the argument — the board is empty because maintaining it is unrewarded work, not because of the flag. Revisit only if derived task boards get built, since that's what would make boards populate as a byproduct.

PR #317, opened 2026-08-17
feat(orchestration): capability-leveling — tier-aware validation, retry, and escalation for weak/local models

Josh-Schoen:
## What this is

Capability-leveling lets weak/local models approach stronger-model quality on Loom tasks by spending **structure + verification + selective escalation** instead of model capability — a thin controller composed from primitives Loom already ships (catalog, `OutputValidator`, provider pools, FTS5 memory). **Off by default at every layer**, and provably inert when off.

Design record with full evidence: `docs/plan-capability-leveling.md`. Every quantitative claim below traces to committed per-trial JSONL in `docs/experiments/`.

## What's in the box

- **C1 tier abstraction** (`pkg/llm/catalog/tier.go`): `frontier|mid|small-open|local|unknown` derived purely from existing catalog fields. ~16ns, 0 allocs.
- **Leveling executor** (`pkg/orchestration/leveling_executor.go`): composes `ValidateAndRetry` (previously had no production callers); free signals own the verdict; escalation ladder bounded by `max_escalations` and a `max_cost_usd` ceiling (Loom's first mid-run cost enforcement).
- **Proto surface**: `LevelingPolicy` on `PipelineStage` (field 8) and `AgentTask` (field 5). `buf breaking` clean. No judge field in proto — config alone can never add a paid LLM call.
- **Workflow-YAML surface**: `leveling:` block; absent block ⇒ nil field ⇒ feature nonexistent for that stage. Also closes the pre-existing YAML gap for `session_mode`/`feedback_template`/`cooldown_ms`.
- **Ollama fixes found en route**: tag-tolerant catalog lookup (`llama3.2:latest` previously resolved to no tier), opt-in `RegisterOllamaCatalogSource` (both halves of the discovery→catalog bridge were dead code), opt-in `seed` on `LLMConfig` (proto-first).
- **Global behavior change — `catalog.Lookup` (all providers)**: after an exact miss, `Lookup` retries with the model tag/version suffix stripped. Exact hits are unchanged; lookups that previously returned `nil` may now resolve. Verified paid-provider impact: the Bedrock inference-profile ID `us.anthropic.claude-opus-4-6-v1:0` previously missed the catalog and now resolves to the `-v1` entry, so `ResolveContextLimits`, `EffectiveOutputReservation`, and `ProviderFactory.resolveMaxOutput` read catalog values instead of a caller default. Pinned by `TestBaseModelID` (Bedrock cases) and `TestLookupTagFallbackBedrockInferenceProfile`; documented under "Behavior changes for existing configs" in the plan doc.
- **Five live experiment harnesses** (env-gated on `LOOM_LIVE_OLLAMA=1`, skipped in `-short`, never touch CI).

## The evidence (five live experiments, local models, $0)

| Gap | Verdict | Key numbers |
|---|---|---|
| Format | ✅ closed, free | llama2 schema-valid 0/30 raw → 30/30 leveled, zero added latency on pass |
| Knowledge | ✅ closed fully — weak beats strong | fictional corpus: weak+BM25 30/30 vs strong-alone 0/30; utilization 75/75 |
| Reasoning | ❌ not closable locally — only routed around | retry repaired 0 reasoning errors; self-critique 0/18; scaffolding −13 to −37pp; escalation behind an oracle signal 40%→90% |

**Home-turf validation (SQL generation, llama3.2 → deepseek-r1)**: one same-model rung carrying the execution error fixed all 8 execution-visible failures (57%→77%; 8 same-model escalations, zero strong-model calls; 0.19s of sqlite as the judge). The quantified limit: 7 silently-wrong queries (execute cleanly, wrong results) are structurally invisible to any execution-only signal — that blind spot needs result verification or a reasoning-class critic (measured: llama3.1-as-judge 35% agreement; deepseek-r1-as-judge 20/20).

## Breaking change (exported `pkg/` API)

`OutputValidator.ValidateAndRetry` (`pkg/orchestration/output_validator.go`) gained two parameters — a `CoerceFunc` for free JSON extraction and a `BudgetFunc` consulted before each retry — and its middle return value changed from `bool` to `ValidationOutcome`. There were no production callers in this repo before this PR (only tests); external Go callers must update the call site. Stated in the plan doc at the "ValidateAndRetry signature" note and marked `!` on the commit.

## Safety notes for review

- `OutputPolicy` is persisted today but enforced nowhere. This PR does **not** switch that on: leveling activates only behind its own explicit `enabled`, and an adversarial test (`TestPipelineOutputPolicyInertWithoutLeveling`) proves a stage carrying an OutputPolicy plus a hostile-but-disabled leveling policy makes exactly one LLM call and passes schema-violating output through unvalidated.
- `RETRY_SESSION_MODE_ESCALATE`'s proto comment claims an LLM upgrade that has never existed in the implementation; the leveling executor is where that behavior now actually lives.
- Negative results are documented as prominently as wins, per the repo's documentation-honesty rules.

## Verification

`go test -tags fts5 -race ./...` green; `buf lint` / `buf breaking --against origin/main` clean; `golangci-lint` 0 issues; `gofmt` clean.

Round-2 review remediation (see the reply comment for the finding-by-finding map): same-model and `role: agent` rungs rejected at load and at resolution; `output_policy` + legacy `output_schema`/`retry_policy` on a leveled stage rejected rather than silently dropping the legacy contract; `tier_policies` now honored on the short-circuit path (was dead config for frontier/unknown/mid); `leveling.rung` and `leveling.judge` spans; `ctx.Err()` checked between rungs; per-rung tier resolved with a downward-ladder warning; proto comments on `max_cost_usd`, `LevelingRung`, `RETRY_SESSION_MODE_ESCALATE`, and the warn-and-continue semantics corrected; user-facing reference at `docs/reference/workflow-leveling.md`, linked from `docs/_index.md`.

Live harnesses reproduce deterministically (seeded; the SQL run was killed mid-flight once and reproduced identically on rerun).


ilsiepotamus (changes requested):
Reviewed at `d622756d` (merge base `6ee35770`, 5 commits behind current main — no conflicts). Every finding below was verified against the code at this head; line numbers refer to it. Verification of the PR's own claims first, because most of them hold and that should be said plainly: `go test -tags fts5 -race -short ./...` is green here across exactly the claimed 95 packages; `buf lint` and `buf breaking --against origin/main` are clean; golangci-lint over the touched trees is clean; the live harnesses are genuinely double-gated (`LOOM_LIVE_OLLAMA` **and** `-short`, checked before any socket opens — verified empirically in all three gate modes, all skips at 0.00s); the committed experiment JSONL **regenerates byte-identically** from the PR's own generators, and essentially every quantitative claim in the plan doc reproduces from the committed evidence to the decimal; the headline adversarial inertness test (`TestPipelineOutputPolicyInertWithoutLeveling`) is genuinely probative, not circular; and the no-paid-call-from-config claim holds — `LevelingPolicy` carries no judge field, `LevelingPolicyFromProto` always leaves the judge nil, and the free validator is schema-only. This is an unusually well-evidenced PR. The findings below are why it still isn't mergeable yet.

## Blocker

1. **Negative `max_retries` + enabled leveling is a remotely-triggerable nil-deref panic.** `output_validator.go:120-126` caps `maxRetries` above (10) but has no floor; with `RetryPolicy.MaxRetries = -1` the attempt loop (`for attempt := 0; attempt <= maxRetries`, :138) runs zero times and `ValidateAndRetry` returns `(nil, ValidationOutcome{}, nil)`. Both leveling paths forward that as a non-error (`leveling_executor.go:213-226` short-circuit, `:336-343` exhaustion — "exhaustion is not an error here"), and the executors dereference it: `pipeline_executor.go:342` (`result.Output`) and the parallel merge path. There is no `recover()` in `pkg/orchestration` and no server-wide recovery interceptor. Reachable input: `ExecuteWorkflow` accepts an inline `WorkflowPattern` with no policy validation, so a gRPC client sending `leveling_policy{enabled:true}` + `output_policy{retry_policy{max_retries:-1}}` (and "-1 = unlimited" is a natural guess) panics the process. The YAML loader is immune (it nils non-positive retry policies) — raw proto is not, and the legacy path was guarded (`MaxRetries > 0`) where this new path is not. Fix is two lines of defense: floor `maxRetries` at 0 in `ValidateAndRetry`, and reject negative `RetryPolicy.MaxRetries` in the leveling config validation alongside the other negative-bound checks.

## Major

2. **`max_cost_usd` is not the "hard ceiling summed over every leveling-visible LLM call" it documents (`leveling_executor.go:66-68`).** Three verified gaps, worst first: (a) **primary-rung retries are never budget-gated** — `wrappedExecute`/`wrappedFeedback` accumulate cost, but the retry loop inside `ValidateAndRetry` has no budget hook, so a mid-tier primary with `MaxRetries: 10` runs up to 11 full agent conversations before the ceiling is consulted for the first time at `:289`; (b) the rung and judge checks are check-before-spend (`:289` before `:315`; `:467` before `:477`), so `total=$0.009` under a `$0.01` ceiling still launches a frontier rung — inherent to post-hoc cost, but it means "hard" is the wrong word even for the gated part; (c) the per-attempt cost feeding the ceiling comes from `response.Usage` (`pipeline_executor.go:749-754`), and `agent.Chat` returns only the final LLM call's usage (`pkg/agent/agent.go:2570-2576`), not the tool-loop total — that undercount pre-exists this PR, but this PR is what makes it load-bearing for a spend limit. Suggest: budget hook threaded into the retry loop, and the doc comment rewritten to say what the mechanism actually guarantees (a between-actions check on undercounted attempt costs) — or the mechanism upgraded to match the comment.

3. **Leveling spend is invisible to workflow cost accounting.** Workflow cost sums only each stage's winning result (`calculateCost`, `pipeline_executor.go:894-915`); retry attempts, losing escalation rungs, and judge calls — the exact new spend categories this PR introduces — never enter `allResults`. The correct number is computed in `LevelingReport.TotalCostUSD` and then dropped at both integration points (`pipeline_executor.go:360-366` converts the report to warnings only; the parallel path only logs it), surviving solely as a span attribute. A stage that escalates twice plus a judge call reports the winning rung's cost alone in `WorkflowResult.Cost`, so billing/attribution built on it under-reports precisely when leveling is active. (The comment at `leveling_config.go:133-135` claiming escalation results feed "workflow cost aggregation" is true only for the winning rung.)

4. **A `role:`-based ladder rung silently "escalates" to the same primary model when the role LLM isn't configured.** `resolveLevelingLadder` (`leveling_config.go:166-171`) treats `GetLLMForRole` returning nil as a config error, but `GetLLMForRole` ends with `return a.llm` (`pkg/agent/agent.go:3608-3632`) — it falls back to the main agent LLM, so the error branch is dead unless the agent has no LLM at all. A ladder rung `role: judge` on an agent with no judge LLM re-runs the identical primary model at full price, labeled in metadata as the rung. In a cost-control feature, paying for a capability upgrade that never happens is a correctness failure. Suggest a non-falling-back lookup (or an explicit `HasLLMForRole`) for ladder resolution.

5. **Leveling on stages inside an `iterative` pipeline is parsed, validated, and silently ignored.** `convertIterativePattern` reuses `convertPipelinePattern` (`workflow_config.go:569-573`), which accepts and semantically validates per-stage `leveling:` blocks — but the iterative executor calls `executeStageWithSpan` directly (`iterative_pipeline_executor.go:354`); the `GetLevelingPolicy().GetEnabled()` gate exists only in `PipelineExecutor.executeFrom` (`pipeline_executor.go:324`). Moving a working leveled pipeline under `type: iterative` silently disables leveling and the OutputPolicy enforcement it carries. This is the exact failure mode the PR's own loader comment condemns (`workflow_config_leveling.go:31-35`). Either wire the gate into the iterative stage path or reject the block on iterative stages at load.

6. **Enabling leveling on a strong-tier stage silently removes the JSON-extraction normalization the legacy path always applied.** Legacy validation extracts JSON from prose/fences and normalizes `result.Output` (`pipeline_executor.go:369-380`, `:921-940`); the leveling short-circuit path (frontier/mid/unknown) passes `coerce = nil` (`leveling_executor.go:210-226`) into a strict validator that rejects fenced JSON as "not valid JSON" (`output_validator.go:245-256`). A frontier agent returning ` ```json {valid} ``` ` passes today and fails under leveling — then, per graceful degradation, the pipeline continues with the raw fenced text downstream where the legacy path would have normalized it. The same asymmetry on the active path burns real money: a mid-tier primary (coercion off in `DefaultTierPolicies`) with fenced-but-valid output triggers a paid retry and a paid escalation whose fenced output also fails (rung outputs are evaluated with the primary's tier policy, `:436`), exhausting the ladder on output that was valid all along. Suggest: fence-stripping belongs in the free validation path for every tier — it's exactly the "free signals own the verdict" philosophy this PR argues for.

7. **The `parseOutputRetryPolicy` rewrite makes previously-loading non-leveling configs fail to load** (`workflow_config.go:786-838`, also reached from conditional and swarm parsing). Verified against the merge-base parser: `max_retries: 2.5` (was: truncated to 2) and `max_retries: "3"` (was: silently no retries) are now hard load errors, and `retry_policy: {session_mode: fresh}` without positive `max_retries` (was: silently dropped) now refuses to load. The strictness is defensible and clearly deliberate — but it contradicts the PR's central "absent block ⇒ byte-identical behavior" claim as stated, and a deployed workflow YAML can stop loading on upgrade. Either soften to warnings for the previously-tolerated shapes or amend the claim and call it out as a breaking config change in the PR description/release notes.

8. **The tag-tolerant catalog `Lookup` fallback changes behavior for existing configs that never mention leveling** (`catalog.go:997-1006`). It applies to all providers and to every consumer (`model_context_limits.go`, `factory.go`), and the PR's own test edit documents the delta: `ollama/llama3.1:70b-instruct` moves from the legacy 10% reservation (12800) to the catalog entry (8192) with zero user config change; all size/quant variants of a tag inherit one entry's window. The doc comment's "a lookup that succeeds today cannot change answer" is true but covers only half the change — lookups that failed today now succeed, and that is the behavior change. Deserves the same treatment as 7: explicit opt-in, or explicit release-note acknowledgment plus a comment that states the real contract.

9. **Doc honesty: the summary's "Coercion rescued 30/30 malformed replies, twice" (`docs/plan-capability-leveling.md:1485`, echoed at :1047-1048) overstates the evidence.** The committed JSONL contains exactly one such measurement (reasoning arm 2); the doc's own Phase 2b verdict (:720-723) records "0 coercions across 40 trials … untested by this experiment, not validated by it", and Phase 2b's format gap was closed by schema-feedback retries, not coercion. In a PR whose credibility rests on evidence discipline — and which otherwise reproduces to the decimal — the headline claim being the one overstatement matters. One-sentence fix.

## Minor

10. An empty tier entry (`tier_policies: {local: }`) zeroes the tier's knobs while the comment at `workflow_config_leveling.go:261-264` promises defaults — a zero-vs-absent trap in the safe direction, but the comment is wrong.
11. NaN survives every scalar validation: `max_cost_usd: .nan` loads (`GetMaxCostUsd() < 0` is false for NaN, `leveling_config.go:80`) and `budgetExceeded`'s `NaN > 0` is false — a config that appears to have a ceiling has none; same hole misclassifies frontier reasoning models via NaN thresholds (`tier.go:141-147`, :194-197).
12. Negative pricing thresholds silently become built-in defaults (`tier.go:141-147`) while sibling fields reject negatives as "a config mistake" (`leveling_config.go:74-95`) — pick one rule.
13. `LLMConfig.seed` (proto field 11) is honored only via `pkg/agent/registry.go:1182`; `cmd/looms/cmd_serve.go`'s proto-config constructor drops it, so the proto comment "Honored by: ollama" is overbroad for server-built providers. Also worth noting: the lax agent-YAML loader means a pre-existing stray `seed:` key that was silently ignored becomes live via the registry path.
14. The leveling + `validation_prompt` conflict is rejected only at execution time (`pipeline_executor.go:780-784`) after earlier stages have already run and spent; the loader eagerly validates every other semantic error and could catch this one too.
15. Exhaustion returns the last rung's result while comments say "best result" (`leveling_executor.go:334-343`) — a failing escalation's garbage can replace a nearly-valid primary output downstream.
16. The short-circuit path reports `TotalCostUSD` as the final attempt only (`:217`) while the active path sums all attempts — inconsistent accounting for the same field.
17. `ModelsUsed` always reports the primary (`pipeline_executor.go:819`) even when an escalation rung produced the winning output, contradicting the result metadata.
18. A judge skipped by the cost ceiling fails open as documented but emits no report warning (`leveling_executor.go:467-475`) — an unexamined output surfaces as fully validated with only a span attribute as witness (Go-API-only today, since proto carries no judge).
19. Live-only: the "happy path adds no call" assertion (`leveling_live_ollama_test.go:454-458`, :537-539) fires falsely on a legitimate retry-then-pass trial — `len(report.Warnings)==0` is the missing conjunct; the plan doc (:656-658) describes the intended predicate, not the implemented one.
20. Doc summary cleanups: the "~16 ns" enabled-but-unneeded overhead line (:1464-1466) contradicts the doc's own Phase 2c measurement (~1.3–4 µs — both negligible, but the summary understates its own data ~100×); the header still says "**Status**: Draft — ready to start"; the 30.4s figure at :1438 is the one number that doesn't reproduce from committed JSONL (16.0× conclusion unaffected); and `docs/experiments/probe_r1_judge.py` / `probe_scaffold.py` hardcode author-machine paths, making their two JSONL outputs the only evidence files not regenerable from the repo.

## Verified clean, for the record

Concurrency: no races found — per-stage executor/policy/ladder construction, mutex-guarded agent getters, replace-not-mutate provider pools, RWMutex'd catalog; `-race` green on every affected package across repeated runs. Escalation bounds honored on every path (no infinite-loop shape; counter increments on rung errors). No goroutine/span/session leaks. Proto surface: field numbers, reserved-field hygiene, import graph, and generated code all check out; `enabled: false` with a hostile block is runtime-identical to no block; partial-block defaults land on documented values; `ValidateAndRetry`'s signature change had zero non-test callers at the merge base, exactly as claimed. Evidence: datasets regenerate byte-identically; every accuracy/call/escalation/judge statistic in the doc reproduces from committed JSONL (upper-median convention noted); negative results are reported with unusual candor.

## Verdict

Requesting changes for the panic (1) and the cost/inertness gaps (2-8). The bones here are genuinely good — the evidence discipline is the best I've seen in this repo — and nothing above touches the architecture: 1, 4, 5, 9 are small fixes; 2, 3, 6 are contract-vs-implementation reconciliations where either side moving would satisfy; 7 and 8 mostly need honest labeling as the behavior changes they are. Also recommend rebasing onto current main before merge (the branch predates the MCP server-side merge; CI gained an interop job since). Happy to re-review quickly.

Josh-Schoen:
Thank you for this review — reproducing the evidence chain to the decimal and then finding a remotely-triggerable panic on top of it is the most useful review this branch could have gotten. All 20 findings are addressed. The branch is also rebased onto current main per your recommendation (one add-add conflict in `config_loader.go` — your new `RateLimit` field and this PR's `Seed` landed on the same struct tails — resolved keeping both, build- and test-verified).

Two commits: the F1 floor is isolated and cherry-pickable (`fix(orchestration): floor ValidateAndRetry's retry bound`); everything else is in the follow-up remediation commit.

## Blocker

**1 — Fixed, both defenses.** Floor at `output_validator.go` (clamp to [0,10]; doc now states the invariant "a nil error guarantees a non-nil result") plus conversion-time rejection. One placement deviation from your suggestion: the negative-bound check lives in a sibling `validateLevelingOutputPolicy` wired at both conversion call sites rather than inside `LevelingPolicyFromProto`, because that function never sees the stage `OutputPolicy` where `RetryPolicy` lives. Regression tests construct your exact attack input as raw proto (YAML bypassed) on both the pipeline and parallel paths.

## Major

**2 — Both sides moved.** `ValidateAndRetry` gained an optional `BudgetFunc` threaded into the retry loop (check-before-spend, retries only — the first attempt is never gated), `ValidationOutcome` gained `BudgetExhausted`, and the `MaxCostUSD` doc now states the real guarantee: a between-actions check on post-hoc, potentially undercounted attempt costs — not a billing limit. The tool-loop undercount (your 2c) remains pre-existing and is now named in the comment.

**3 — Fixed.** `adoptLevelingCost` stamps `LevelingReport.TotalCostUSD` onto the winning result at both integration points before it enters `allResults`, so `WorkflowResult.Cost` and `AgentCostsUsd` carry the whole ladder's spend. The false comment you caught in `resolveLevelingLadder` is corrected. Test: 1 retry + 1 escalation ⇒ summed cost in the workflow result.

**4 — Fixed.** `GetLLMForRoleStrict(role) (LLMProvider, bool)` added to `pkg/agent` (additive; documented as existing for ladder resolution). A `role:` rung on an agent without that role's LLM is now a resolution error instead of a silent full-price re-run of the primary.

**5 — Wired, not rejected.** `iterative_pipeline_executor.go` now has the same enabled gate as the plain pipeline; the legacy restart-retry loop was extracted intact, and validation warnings ride the previously-unset `WorkflowCheckpoint.validation_warnings` through HITL suspend/resume. Test covers escalation-wins (primary called exactly once, proving the legacy loop doesn't double-run) and exhaustion-surfaces-warnings.

**6 — Your suggestion taken to its conclusion.** JSON extraction is now unconditional on every schema-bearing leveling path (short-circuit, active, rung evaluation). That made `AggressiveCoercion` gate nothing, so it got the `scaffolding_depth` treatment: Go field removed, proto field 2 reserved by number and name (message is new in this PR; `buf breaking` clean), explicit YAML key is a load-time error. Test pins fenced-valid JSON passing free on all four tier paths with zero retries/escalations.

**7 — Tolerance restored.** The three previously-tolerated shapes load again with structured zap warnings (fractional `max_retries` truncates, string is ignored, retry-only keys without a positive bound drop the policy — each warning names the YAML path and resolution). Implemented via new `...WithLogger` loader variants; all pre-existing exported signatures unchanged. Leveling keys stay strict. One test-suite consequence: three pre-existing cases pinned `max_retries: "many"` as an error and were retargeted to a never-accepted shape (`max_retries: true`).

**8 — Labeled, not gated.** The `catalog.go` comment now states the two-direction contract (hits unchanged; previously-failing lookups may now succeed) and names the affected callers including the context-limit/output-reservation paths. The plan doc gained a "Behavior changes for existing configs" section covering this and finding 7, written for release notes.

**9 — Corrected in three places** (you found two; a third copy of the claim was in the gap-matrix table). The doc now records one coercion measurement, not two, and names the actual mechanism per experiment.

## Minor

**10** — Null and `{}` tier entries now resolve to the built-in defaults the comment promised; explicit `retry_budget: 0` stays 0. **11** — NaN/±Inf rejected at convert time for `max_cost_usd` and both thresholds; `TierThresholds` treats non-finite like non-positive for direct-Go callers. **12** — Negative thresholds are convert-time errors like their siblings. **13** — Proto comment scoped to the agent-registry path; cmd_serve wiring stays the documented follow-up; the stray-`seed:`-key note is in the behavior-changes section. **14** — Conflict rejected at load time; execution-time check kept as defense. **15** — Exhaustion returns the primary's last result when the final rung errored, else the last rung's; comment matches. Note: per the letter of this, "good rung then errored final rung" returns the primary. **16** — Short-circuit path totals every attempt. **17** — `ModelsUsed` reports the winning rung's model. **18** — Budget-skipped judge leaves a report warning. **19** — The live predicate gained its missing `warnings == 0` conjunct. **20** — µs figure restored (~1.3–4µs delta), header truthful, and the 30.4s figure turned out to be *misattributed* rather than irreproducible — it is arm 2's sum; arm 3 is 29.7s, ratio corrected to ~16×. Both probe scripts now take paths from flags/env so their JSONLs regenerate from the repo.

Gates on the rebased head: `go build -tags fts5 ./...`, full `go test -tags fts5 -race -short ./...` (95 packages), `buf lint`, `buf breaking --against main`, `gofmt` — all clean. Ready for re-review.

ilsiepotamus (changes requested):
Review complete. All commands run in the isolated worktree; nothing was posted to GitHub.

---

# PR #317 — capability-leveling (Josh-Schoen, `feat/capability-leveling` → `main`)

## 1. DESCRIPTION

**What it does.** Capability-leveling is an opt-in per-stage / per-task controller that lets a weak or local model meet an output contract by spending *structure + verification + selective escalation* instead of model capability. It is composed from primitives Loom already ships (model catalog, `OutputValidator`, agent provider pool / role LLMs) rather than a new subsystem.

**Tiers.** `pkg/llm/catalog/tier.go:41-55` defines five ordinal tiers — `unknown < local < small-open < mid < frontier`. A model's tier is *derived*, not configured: `TierFromInfoWith` (`tier.go:208-224`) reads the catalog's existing `ModelInfo` fields — 0/0 pricing ⇒ `local`; reasoning + output price ≥ `$10/1M` ⇒ `frontier`; reasoning **or** output ≥ `$1.5/1M` ⇒ `mid`; else `small-open`; not in the catalog ⇒ `unknown`. Both thresholds are overridable per policy (`frontier_min_output_cost_usd`, `mid_min_output_cost_usd`). Nothing is stored or cached; the tier is recomputed per call.

**Judging failure.** Only the *primary* rung's tier is computed (`leveling_executor.go:217`). `shouldShortCircuit` (`:401-413`) then decides: `frontier` → always short-circuit; `mid` → short-circuit unless `short_circuit_mid: false`; `unknown` → short-circuit (refuse to guess); `local`/`small-open` → run the active path. Quality is judged by free signals first: the stage/task `OutputPolicy.output_schema` owns the verdict when present, with free JSON extraction from fenced/prose-wrapped text applied before declaring failure. A `LevelingJudge` (paid LLM) is consulted **only** when there is no schema — and it is deliberately unreachable from proto/YAML (`leveling_config.go:52-57`), so config alone can never add a paid call.

**On failure.** Two bounded mechanisms, in order: (1) *same-model retry* on the primary, budget from `DefaultTierPolicies()` (`leveling_executor.go:44-52`: local/small-open = 2, mid = 1, frontier/unknown = 0), overridable per tier, and always losing to an explicit caller `retry_policy`; (2) *escalation* up an ordered `ladder` of rungs, each resolved from the executing agent's role LLMs (`GetLLMForRoleStrict`) or provider pool — bounded by `max_escalations` (default 1) and `max_cost_usd`. Rungs are one-shot `LLMProvider.Chat` calls: no tools, no system prompt, no session. Exhaustion is **not** an error: the best output obtained is returned with `Passed=false` and warnings.

**Shape/risk.** 64 files, ~19.5k additions, but the hand-written production core is small: ~2.3k lines across `tier.go` (224), `leveling_executor.go` (572), `leveling_config.go` (380), `workflow_config_leveling.go` (504), `catalog_source.go` (151), plus edits to `pipeline_executor.go`, `parallel_executor.go`, `iterative_pipeline_executor.go`, `workflow_config.go`, `output_validator.go`. The rest is ~7.6k lines of CI-running tests, ~3.4k lines of env-gated live experiment rigs (`LOOM_LIVE_OLLAMA=1`, skipped in `-short`), a 1,611-line design record, and committed per-trial JSONL. Off by default at every layer, with adversarial inertness tests. Genuine risk is concentrated in three places: a **global** `catalog.Lookup` semantic change affecting all providers; the pipeline's validation-failure semantics changing when leveling is enabled; and dead config surface.

**Verification in this worktree:** `go build -tags fts5 ./...` ✅ · `go vet` ✅ · `go test -tags fts5 -race ./pkg/orchestration/... ./pkg/llm/{catalog,factory,ollama}/... ./pkg/agent/...` ✅ (exit 0, no races) · `gofmt -l` clean · `golangci-lint 2.9.0` **0 issues** · `buf lint` clean. `buf breaking --against origin/main` reports one violation in `llm_scheduler.proto` (`ceiling_pinned` field 10) — that file is **not** in this diff; it is branch staleness (11 commits behind main, head 2026-08-25). Rebase needed; no `pkg/orchestration` overlap on main, but `pkg/agent/agent.go` has moved 736 lines.

---

## 2. SHAPE REVIEW (fit against `CLAUDE.md`)

**Done well:**

- **No role prompting.** `git diff origin/main...HEAD | grep -iE "You are (a|an|the)|As a .* (expert|engineer)"` → **zero hits** across all added lines, including the retry/escalation prompt builder (`output_validator.go:329-343`) and every live-experiment prompt.
- **No banned marketing vocabulary.** Zero hits for `production-ready|comprehensive|seamless|enterprise-grade|robust|powerful|…` in added lines. The PR actively *removes* two instances of "comprehensive" from doc comments it touches (`pipeline_executor.go:657`, `parallel_executor.go:218`).
- **Reuse over parallel system.** `LevelingExecutor` composes `OutputValidator.ValidateAndRetry` rather than duplicating retry (`leveling_executor.go:233,267`); rungs resolve through the *existing* `Agent.GetLLMForRole*` / `GetProviderPool` (`leveling_config.go:220-245`) — no provider construction, no new router; tiering is a pure function over existing `ModelInfo` fields. No new store, no pricing table.
- **Proto first.** Proto changes precede implementation; `reserved 2/3` + `reserved "aggressive_coercion"/"scaffolding_depth"` (`orchestration.proto:474-489`) for knobs cut pre-merge; `buf lint` clean; `optional bool/int32` used correctly so absent ≠ proto3 zero.
- **Table-driven tests, extensively.** `TestPipelineLevelingConfigErrors` (`pipeline_leveling_test.go:314`), `TestBaseModelID` (`modelid_test.go:26`), `tier_thresholds_test.go` (424 lines), `workflow_config_leveling_scalars_test.go` (398). 50+ leveling test functions.
- **Documentation honesty is genuinely exemplary.** Negative results are published as prominently as wins (C2 scaffolding rejected: −13 to −37pp; C4 rejected; self-critique 0/18). Status markers (✅/⚠️/📋/❌) used correctly and *updated in place across phases*. Limitations stated plainly, e.g. `docs/plan-capability-leveling.md:1409-1413` — *"a YAML-configured leveling policy can close the **format** gap and nothing else."* I recomputed the headline SQL numbers from the committed JSONL and they reproduce exactly: 17/30 (56.7%) → 23/30 (76.7%), 8 escalations, r1-alone 30/30.

**Flagged:**

- **Escalation LLM calls are not traced.** `levelingRungExecute` (`leveling_config.go:279-307`) calls `llm.Chat` with the bare ctx and **no span**, and `judgeVerdict` (`leveling_executor.go:534`) likewise — while every *primary* attempt gets a `pipeline.agent.*` span (`pipeline_executor.go:693-695`). Only one aggregate `leveling.execute` span exists (`leveling_executor.go:199-212`) with counts, not per-rung detail. Against principle #4 / "instrument all LLM calls". The doc claims it matches `validateStageOutput`'s precedent, but that function *does* thread the tracer through a `mergeContext` (`pipeline_executor.go:871-875`).
- **Docs not wired into `docs/`.** The only documentation is `docs/plan-capability-leveling.md`, unlinked from `docs/_index.md` and `docs/README.md`. There is **no reference doc** for the new `leveling:` YAML block, and `docs/reference/workflow-output-retry.md` is untouched even though this PR makes `session_mode` / `feedback_template` / `cooldown_ms` functional from YAML for the first time and turns some previously-tolerated `retry_policy` values into load errors.
- **Exported `pkg/` API broken without a shim.** `OutputValidator.ValidateAndRetry` gained two parameters and changed a return type (`output_validator.go:120-129`). Correctly marked `!` in the commit and stated plainly at `plan-capability-leveling.md:932`, but omitted from the PR description; only in-repo callers were tests.
- **Python in `docs/experiments/`.** `probe_r1_judge.py` (107), `probe_scaffold.py` (107) — one-off research probes, against "Use Golang over Python". Minor.

---

## 3. COMPREHENSIVE REVIEW

### (a) Nil-pointer / missing-case panic risk — **no reachable panic found**

I traced every map/slice index and pointer deref in the new code:

| Site | Verdict |
|---|---|
| `ladder[0]` (`leveling_executor.go:195,217,227,258`) | Guarded by `len(ladder)==0` at `:186` |
| `ladder[i]` (`:294-333`) | Loop-bounded |
| `e.policy.TierPolicies[tier]` (`:418`) | Nil-map read is legal Go |
| `DefaultTierPolicies()[tier]` (`:421`) | Missing key ⇒ zero `TierPolicy`; all five tiers present anyway |
| `outcome.Err.Error()` (`:453`) | **Safe, but only by cross-file invariant** — see below |
| `pool[pr.GetProvider()]` (`leveling_config.go:238`) | Nil-pool guarded at `:232`, `!ok || llm == nil` at `:239` |
| `warnings[len(warnings)-1]` (`output_validator.go:206,210,215,220`) | Only reachable at `attempt≥1`, which implies attempt 0 appended |
| `levelingReport.Passed` (`iterative_pipeline_executor.go:349`) | Non-nil only because the branch is gated on `GetEnabled()` in a different file |
| `result.GetMetadata()` / `r.GetCost()` | Proto getters, nil-safe |

The `outcome.Err.Error()` deref at `leveling_executor.go:453` is safe **only** because commit `66f2b5cc` added the floor at `output_validator.go:146-153` (a negative `max_retries` previously made the attempt loop run zero times and return `(nil, {Err:nil}, nil)`). That fix is real and correct, and is defended by four tests (`leveling_retry_bound_test.go`) plus a second, explicit rejection at the config layer (`validateLevelingOutputPolicy`, `leveling_config.go:152-157`) wired into both executors. Good.

Two residual, low-severity vectors, both requiring a Go caller (not reachable from proto/YAML):
- `LevelingRung.Execute` is an exported field and `Execute` never nil-checks it (`leveling_executor.go:195,324`) → nil func call.
- `output_validator.go:232` derefs `result.Output` if an `ExecuteFunc` returns `(nil, nil)`. Pre-existing; none of the shipped rung builders can.

### (b) Cost ceiling — **enforced as a gate, but the proto over-promises** *(MAJOR)*

The gate is real and is checked in all three spend sites: before each primary retry (`BudgetFunc` at `output_validator.go:182-190`), before each escalation rung (`leveling_executor.go:298`), before each judge call (`:519`). `TestPipelineLevelingEscalatesBounded/bounded by max_cost_usd` proves a paid rung is skipped. That is a genuine hard stop on *further* escalation.

But it is not a spend ceiling, for two reasons the Go doc states honestly (`leveling_executor.go:65-79`) and the **proto does not**:
1. It never interrupts an in-flight call, so the ceiling is normally overshot by one call.
2. The running total is a *floor* on real spend: `resultCostUSD` reads `AgentResult.Cost.CostUsd`, which for the agent path comes from `Response.Usage = llmResp.Usage` — the **final** LLM response only (`agent.go:2603`, `:2947`). A stage that ran a 12-turn tool loop reports one turn's cost, so the gate lets calls through that accurate accounting would block.

`proto/loom/v1/orchestration.proto:502` says **"Hard cost ceiling in USD across all leveling-visible LLM calls (attempts on every rung plus judge spend)"**. That is the surface users read, proto is law, and it makes a guarantee the implementation explicitly disclaims. Fix: bring the proto comment in line with `leveling_executor.go:65-79`. Also note the judge fail-open (`:519-532`) means hitting the ceiling *accepts an unexamined output as passing* — correctly warned in the report, but a second reason "hard ceiling" is the wrong word.

### (c) Do all retry/escalation paths go through the gate? — **yes; the iterative bypass is closed** *(clean)*

This is handled well. `IterativePipelineExecutor` now branches on `stage.GetLevelingPolicy().GetEnabled()` (`iterative_pipeline_executor.go:341`) and routes to the *same* `executeStageWithLeveling`; the old restart-policy retry loop was extracted verbatim into `executeStageWithValidationRetries` (`:1120-1230`) for the non-leveled branch, so the two never both run and never double-charge. `TestIterativePipelineHonorsLeveling` (`leveling_workflow_cost_test.go:105`) covers it. `ParallelExecutor.executeTaskWithSpan` gates identically at `parallel_executor.go:223`. `AgentTask` is consumed only by the parallel executor and `PipelineStage` only by the two pipeline executors (verified by grep) — no third executor silently ignores the field.

Within `Execute`, primary and rungs use equivalent verdicts: the validator's schema check (`output_validator.go:274-289`) vs. `evaluate`'s (`leveling_executor.go:485-499`) both check `output_schema` and both apply free JSON extraction. No path skips leveling once enabled.

One gap: **no `ctx.Err()` check in the escalation loop** (`leveling_executor.go:294-344`). `ValidateAndRetry` checks per attempt (`output_validator.go:168`); the ladder does not, so a cancelled context makes every remaining rung fire and fail, and the caller receives a non-error "exhausted" result instead of the cancellation. *(minor)*

### (d) Fallback when a model/role has no rung — **safe for a missing rung, unsafe for a useless one** *(MAJOR ×2)*

Handled correctly:
- Role rung with no role LLM → hard error naming rung index and agent (`leveling_config.go:226-230`). `GetLLMForRoleStrict` (`agent.go:3807-3836`) was added specifically so the lenient `GetLLMForRole` fallback can't silently build a same-model ladder. Right call.
- Provider absent from the pool, nil pool, nil rung, rung with neither role nor provider → all hard errors (`:217-245`), each with an actionable message, checked twice (load time + execute time). `TestPipelineLevelingConfigErrors` covers them.
- Unclassifiable primary (`TierUnknown`) → conservative short-circuit, never adds cost.

Two real gaps, both empirically confirmed:

**(d1) `role: agent` builds a rung identical to the primary.** `parseLLMRoleName("agent")` → `LLM_ROLE_AGENT` (non-zero, `agent_config.proto:213`), which `GetLLMForRoleStrict` resolves to `a.llm` — the primary's own model. `validateLevelingLadderShape` accepts it. Verified:
```
parseLLMRoleName("agent") = LLM_ROLE_AGENT, ok=true
validateLevelingLadderShape accepts a role:agent rung
```
That is exactly the "escalation that spends a call and cannot improve anything" the strict lookup exists to prevent. Nothing rejects a rung repeating the primary's `provider`/`model` either, and **rung tiers are never computed** — only `ladder[0]`'s (`leveling_executor.go:217`) — so a "ladder" may escalate sideways or downward with no diagnostic, despite the proto describing rungs as *"a stronger model tried when a weaker primary's output fails"* (`orchestration.proto:444-445`).

**(d2) `tier_policies` for short-circuiting tiers is dead config.** The short-circuit branch passes the caller's `outputPolicy` straight to the validator (`leveling_executor.go:233-235`) and never reaches `tierPolicyFor` (`:251`). So `retry_budget` under `frontier`, `unknown` (and `mid` under the default `short_circuit_mid: true`) can never take effect, yet the YAML loader, the proto, and `LevelingPolicyFromProto` all accept and validate it. Verified:
```
frontier retry_budget=3 -> 1 call   (tier policy ignored)
unknown  retry_budget=3 -> 1 call   (tier policy ignored)
local    retry_budget=3 -> 4 calls  (tier policy honored)
```
This is the *precise* failure mode the PR goes out of its way to eliminate elsewhere — `workflow_config_leveling.go:31-43` rejects the removed `scaffolding_depth`/`aggressive_coercion` keys because *"a config the loader silently drops is the exact failure mode removing the knob was meant to prevent."* The same rule is not applied here. Fix: reject (or warn on) a `tier_policies` entry for a tier that will short-circuit under the same policy, and document it on `orchestration.proto:519`.

**(d3) Legacy `output_schema` is silently dropped when `output_policy` is also present.** `effectiveLevelingOutputPolicy` (`leveling_config.go:369-380`) returns `stage.output_policy` if non-nil and never looks at `stage.output_schema` — and the leveled path *skips the entire legacy validation block* (`pipeline_executor.go:360-365`). So a stage with a working legacy `output_schema` that also gains an `output_policy` goes from **enforced** to **completely unenforced** the moment leveling is enabled. Verified:
```
effective schema = ""   (legacy schema was {"type":"object","required":["legacy"]})
```
Compare the loud, well-argued rejection of `validation_prompt` + leveling at `pipeline_executor.go:786` — *"the criteria would be silently dropped"*. Identical reasoning applies; the check is missing. *(MAJOR)*

**(d4) Enabling leveling converts a hard-fail stage into warn-and-continue.** Pre-PR, a stage whose `output_schema` failed with no retry policy returned a workflow error (`pipeline_executor.go:411-413`). On the leveled path, exhaustion only appends a warning (`:360-364`, `:831-845`) and the bad output flows to the next stage. `plan-capability-leveling.md:411` calls this *"matching the existing retries-exhausted behavior"* — true of the retry branch, **not** of the no-retry branch it also replaces. Nothing on the proto or YAML surface warns of it. *(MAJOR)*

**(d5) Wrong warning text for `tier=unknown`.** `levelingStageWarnings` hardcodes `"leveling short-circuited on a strong primary and did not escalate"` for *any* `ShortCircuited` (`pipeline_executor.go:839`), so an unclassified model is reported to the operator as a strong one. The `tier=unknown` field is present alongside, so it's discoverable, but the sentence is false. *(minor)*

### (e) Test coverage — **strong; the end-to-end ladder is covered**

The escalation ladder is exercised end-to-end through the real executors, not just in isolation:
- `TestPipelineLevelingEscalationOutputWins` (`pipeline_leveling_test.go:222`) — weak primary emits invalid JSON → tier retries → escalation rung succeeds → *its* output becomes `MergedOutput`, carries `leveling_rung_provider/model`, is attributed to the stage's agent, and total cost = 0.9.
- `TestPipelineLevelingEscalatesBounded` (`:151`) — two independent bounds proven separately: `max_escalations` (3 primary calls + rung1 only, rung2 untouched) and `max_cost_usd` (0 paid rungs, `budget_exhausted=true`).
- `TestPipelineOutputPolicyInertWithoutLeveling` / `TestParallelOutputPolicyInertWithoutLeveling` — adversarial: a *hostile* disabled policy (bad tier key, negative bounds, unresolvable rung) plus a violated schema still makes exactly one LLM call and passes bad output through. This is the right test to have written.
- `TestIterativePipelineHonorsLeveling`, `TestPipelineLevelingCostCoversTheWholeLadder`, `TestLevelingBudgetGatesPrimaryRetries`, `TestLevelingCoercionBeatsEscalation`, `TestLevelingJudgeErrorFailsOpen`, `TestLevelingRungExecutionErrorContinues`, plus a YAML→executor integration through `testdata/leveling-pipeline.yaml`.

Gaps, all corresponding to findings above and all cheap to add: no test that a `frontier`/`unknown` `tier_policies` entry is honored *or* rejected; no test for a `role: agent` rung; no test for legacy `output_schema` + `output_policy` coexisting; no Bedrock case in `TestBaseModelID`; no test that a cancelled context short-circuits the ladder.

### (f) Documentation accuracy

`docs/plan-capability-leveling.md` is the most honest design record I have reviewed in this repo — it publishes disconfirming evidence, marks rejected components ❌ with the measurement that killed them, revises earlier claims in place, states the exported-API break plainly (`:932`), and records that `just lint` could not be run locally (I ran it: golangci-lint 2.9.0, 0 issues). The committed JSONL reproduces the published figures exactly.

Inaccuracies found:
1. **`orchestration.proto:502` "Hard cost ceiling"** — contradicted by the implementation's own doc. *(MAJOR, see (b))*
2. **`collaboration.proto:51-52`** still claims `RETRY_SESSION_MODE_ESCALATE` retries *"with an upgraded LLM (switches to orchestrator_llm if available)"*. The author identifies this as never-implemented in both the PR body and the plan doc, and this PR touches `effectiveMode` (`output_validator.go:318-326`) which proves it — but leaves the false proto comment in place. One-line honesty fix, squarely in scope.
3. **Escalation rungs run with no tools and no system prompt.** Honest at `plan-capability-leveling.md:390,478`, absent from `LevelingRung`'s proto comment (`orchestration.proto:444-459`) and the YAML doc (`workflow_config_leveling.go:95-110`) — where a user configuring a ladder actually looks. A tool-dependent stage cannot be rescued by escalation, and nothing at the config surface says so.
4. **PR body: "zero escalations needed"** for the SQL result. The JSONL records `escalations: 8` for arms 2 and 3; what was zero is `strong_model_calls`. The 8 same-model rungs *are* what produced the +20pp. Substance is right, wording inverts the mechanism.
5. **Undisclosed global change:** `catalog.Lookup` now retries with the tag stripped after an exact miss (`catalog.go:1008-1018`). This is documented thoroughly in the Go doc — but it affects **every provider**, not leveling, and the PR description doesn't mention it. I confirmed the paid-provider impact:
   ```
   us.anthropic.claude-opus-4-6-v1:0 -> id=us.anthropic.claude-opus-4-6-v1 ctx=1000000 maxout=128000
   ```
   That ID previously returned `nil`; it now feeds `ResolveContextLimits`, `EffectiveOutputReservation`, and `ProviderFactory.resolveMaxOutput`. `TestBaseModelID` (`modelid_test.go:26-65`) covers ollama, openai fine-tunes, and registry ports — **no Bedrock `:0` case**, the single highest-blast-radius instance. `model_context_limits_test.go:266-296` shows the same class of change for ollama (reservation 12800 → 8192). Likely an improvement, but untested and unannounced. *(moderate)*
6. **`parseOutputRetryPolicy` now errors where it used to ignore.** `include_valid_values` is read via strict `yamlBoolField` (`workflow_config.go:862`); a non-bool value that previously fell through the `.(bool)` assertion now fails the load. The three documented tolerances (`:797-812`) cover `max_retries` only. *(minor back-compat)*

---

**Summary of what needs to change before merge:** proto comment on `max_cost_usd`; proto comment on `RETRY_SESSION_MODE_ESCALATE`; reject-or-document `tier_policies` for short-circuiting tiers; reject-or-document `output_policy` + legacy `output_schema` coexistence; reject-or-warn `role: agent` / same-model rungs; state the hard-fail→warn semantic change on the config surface; add a Bedrock case to `TestBaseModelID`; add spans to rung and judge calls; link/author user-facing docs for the `leveling:` block; rebase (11 commits behind, `pkg/agent/agent.go` moved 736 lines).

**VERDICT: CHANGES NEEDED — 5 majors, 0 blockers** (plus ~9 minor/moderate items detailed above)

Josh-Schoen:
Thank you for the second pass — the (d2) and (d3) findings in particular were the kind that only show up by running the config through the executor, and both were real. Everything below is on the rebased head (11 commits behind → 0; the `llm_scheduler.proto` `buf breaking` violation you saw was branch staleness and is gone). Finding-by-finding, in your numbering:

## Majors

**(b) `max_cost_usd` proto comment — rewritten to match the implementation.** `orchestration.proto` now says what `leveling_executor.go:65-79` says: a between-actions gate, not a hard cap; never interrupts an in-flight call so it is normally overshot by one; the running total is a floor because per-attempt cost comes from the final LLM response's usage; and a judge skipped by the ceiling accepts the output unexamined, with a warning. Closing line: "Use it to bound runaway escalation, not to guarantee a billing limit."

**(d1) `role: agent` and same-model rungs — rejected, at load and at resolution.** `validateLevelingLadderShape` refuses `LLM_ROLE_AGENT` by name (it is the agent's own LLM by definition), and `resolveLevelingLadder` closes the other two routes to the same model: the resolved LLM being the agent's own instance (a pool entry that maps to the main LLM), and the resolved LLM reporting the primary's provider+model or the proto fields naming the primary rung. One shared wording for the load-time and execution-time rejections. Your other point in (d1) — rung tiers never computed — is fixed: each attempted rung's tier is resolved and put on the span (`leveling.rung.N.tier`), and a rung below the primary's tier produces a "ladder escalates downward" warning. Warned rather than rejected, because tiers are pricing-derived heuristics and the wrong authority to fail a run on; equal tiers stay silent since a same-tier rung carrying a different signal is a working pattern (it is what the SQL experiment's +20pp came from). Follow-on fix: the unknown-role error no longer lists `agent` among the valid roles, since following that advice would land in the next error. Tests: five `TestResolveLevelingLadder` subtests (all three routes plus a same-provider/different-model control), a YAML case, a raw-proto executor case, and `TestLevelingRungTierDiagnostics`.

**(d2) `tier_policies` dead for short-circuiting tiers — made live, rather than rejected or warned.** The short-circuit branch now resolves `tierPolicyFor` and the effective output policy exactly as the active path does, and passes the same budget hook. What a short circuit skips is the ladder and the judge, not the tier's own retry budget. The built-in defaults (frontier 0, unknown 0) preserve the historical single call, so nothing changes for anyone who leaves them alone; the one expectation change is `TestLevelingMidShortCircuitConfigurable`, where mid's built-in budget of 1 now yields one retry under `short_circuit_mid: true`. Reasoning for choosing this over your two options: the `scaffolding_depth` precedent removed a knob that gated nothing and could not be given a meaning; `retry_budget` under `frontier` has an obvious meaning a reader will assume it has, so the honest fix is to make it true. The proto comment on `tier_policies` states the defaults and that entries apply on every tier. Test: `TestLevelingShortCircuitHonorsTierPolicies`, 13 cases — defaults, spent budgets, an explicit caller `retry_policy` still winning, the ceiling bounding the retries, and the active-path control.

**(d3) legacy `output_schema` + `output_policy` — rejected, both halves.** `effectiveLevelingOutputPolicy` now returns an error when the stage carries `output_policy` and either legacy field (execution time, so raw proto cannot bypass it), and `validateLevelingLegacyContractConflict` fails the YAML load. Same shape and wording pattern as the `validation_prompt` conflict you compared it to. Tests: `TestEffectiveLevelingOutputPolicy` (conflict cases for schema and for retry_policy, plus the synthesized-legacy and unified-alone cases), `TestPipelineLevelingConfigErrors`, `TestWorkflowYAMLLevelingErrors` (stage and task).

**(d4) hard-fail → warn-and-continue — stated on the config surface.** On `PipelineStage.leveling_policy`, `AgentTask.leveling_policy`, and the `LevelingPolicy` message in proto; in the YAML doc block in `workflow_config_leveling.go`; and the plan doc sentence you quoted now says it matches the retry branch only, and that the no-retry legacy branch becomes warn-and-continue. The new reference doc has a "Failure Semantics" section with the hard-fail contrast side by side.

## Minor / moderate

- **`RETRY_SESSION_MODE_ESCALATE` comment** — corrected in `collaboration.proto`: no model change across retries, points to `LevelingPolicy.ladder` for the thing that actually escalates.
- **Rungs run with no tools and no system prompt** — now on the `LevelingRung` proto comment and in the YAML doc block, where a user configuring a ladder reads.
- **Bedrock `:0`** — `TestBaseModelID` gains two Bedrock inference-profile cases, and `TestLookupTagFallbackBedrockInferenceProfile` pins the exact impact you measured: `us.anthropic.claude-opus-4-6-v1:0` resolves to the `-v1` entry with its 1M context / 128K max-output, an ID stored with its suffix is an exact hit (the 4.7 entries), an unknown model still misses, and stripping is suffix-only, not prefix matching. The PR body now announces `catalog.Lookup` as a global behavior change with that example.
- **Spans on rung and judge calls** — `leveling.rung` carries provider, model, `agent.id`, `agent.session_id`, cost, input/output tokens, duration, and the error on failure; the call runs under a `mergeContext` so the provider's own spans nest, following the `validateStageOutput` precedent you pointed at. `leveling.judge` carries pass, cost, the reason truncated to 200 chars, and any error. Tests: `TestLevelingRungCallIsTraced`, `TestLevelingRungNilTracerIsNoOp`, `TestLevelingJudgeCallIsTraced`.
- **Docs wired into `docs/`** — new `docs/reference/workflow-leveling.md` (YAML key reference, tier derivation, failure semantics, rejected configurations with their error text, span table, measured results, limitations), linked from `docs/_index.md` and `docs/README.md`; `docs/reference/workflow-output-retry.md` updated for `session_mode` / `feedback_template` / `cooldown_ms` becoming functional from YAML, and for the loader tolerances. One correction from writing it: `type: fork-join` has no per-task map and does not carry the block — documented as not supported.
- **`ctx.Err()` in the escalation loop** — checked at the top of every rung iteration; a cancelled context returns the cancellation as an error with the primary's result and spend attached, instead of a non-error "exhausted" report. Test: `TestLevelingEscalationStopsOnCancelledContext`.
- **(d5) wrong warning for `tier=unknown`** — an unclassified primary now gets its own sentence ("model not in the catalog") instead of being called strong. Test: `TestLevelingStageWarningsNameTheCause`.
- **nil `LevelingRung.Execute`** — the whole ladder is checked before either path runs, so a Go caller gets the rung named in an error instead of a nil func call. Test: `TestLevelingNilRungExecuteRejected`.
- **`include_valid_values` back-compat** — a present non-bool is now ignored with a structured warning, resolving to the default `true` where the pre-helper `.(bool)` assertion left it; the value is deliberately not parsed from a string. The round-1 test that pinned it as a load error is replaced by `TestWorkflowYAMLRetryPolicyIncludeValidValuesTolerance`.
- **Exported API break in the PR body** — added as its own section.
- **"zero escalations needed"** — PR body corrected to what the JSONL says: 8 same-model escalations, zero strong-model calls.
- **Python probes in `docs/experiments/`** — left as they are, deliberately. They are one-off generators whose only job is to regenerate the committed JSONL; they are not shipped, imported, or on any build path, and rewriting them in Go would be churn for a research artifact. Happy to move them out of `docs/` if you would rather not have Python under there.
- **Residual `result.Output` deref on `(nil, nil)` from an `ExecuteFunc`** (`output_validator.go`) — pre-existing, not reachable from any shipped rung builder; left alone to keep the diff to the review.

## Gates on the pushed head

- `go build -tags fts5 ./...` — clean
- `go vet -tags fts5 ./...` — clean
- `gofmt -l` (all non-generated Go) — clean
- `golangci-lint run --build-tags fts5` (2.9.0) on `pkg/orchestration`, `pkg/llm/catalog` — 0 issues
- `buf lint` — clean; `buf format --diff --exit-code` — clean
- `buf breaking --against '.git#branch=origin/main'` — clean (the `llm_scheduler.proto` hit was staleness; gone after rebase)
- `buf generate` — `gen/` unchanged (generation is up to date)
- `go test -tags fts5 -race ./pkg/orchestration/... ./pkg/llm/catalog/...` — ok
- `go test -tags fts5 -race ./...` — 98 packages ok, **0 data races**. In the full parallel run, `pkg/mcp/conformance` went red (4 tests, each `initialize failed: context deadline exceeded` at 50s): it launches `npx -y @modelcontextprotocol/server-filesystem` at test time and the startup lost to the 50s deadline under full-suite load. Re-run alone it passes (`ok pkg/mcp/conformance 89.1s`), and `pkg/mcp/` is untouched by this PR.

ilsiepotamus (approved):
Re-reviewed at `7ba0e6c0` against the round-2 head `96d1f77d`. `git range-diff` confirms
the delta is one commit plus a clean rebase onto `fb727480` — which also cleared the
`llm_scheduler.proto` `buf breaking` violation that was branch staleness.

**All five majors are fixed, and I re-ran the same probes I used to raise them.**

- **(b) `max_cost_usd`** — `orchestration.proto:537` now says "Cost gate in USD applied
  between actions, not a hard cap", and states the one-call overshoot, the floor-not-exact
  accounting, and the judge fail-open. That is the surface matching the implementation.
- **(d1) same-model rungs** — rejected at load (`leveling_config.go:182`) and at resolution
  (`:259`), and `rungRepeatsPrimaryModel` (`:330`) closes the two routes I hadn't raised
  (a pool entry holding the agent's own LLM, and explicit provider/model equal to the
  primary's). `agent` is gone from the advertised role list. Re-verified: the rung is
  rejected with a message that names the fix.
- **(d2) dead `tier_policies`** — re-ran the call-count check that produced the original
  `frontier retry_budget=3 -> 1 call`:

  ```
  frontier retry_budget=3 -> 4 calls   (was 1)
  unknown  retry_budget=3 -> 4 calls   (was 1)
  mid      retry_budget=3 -> 4 calls   (was 1)
  local    retry_budget=3 -> 4 calls   (unchanged)
  ```

  Splitting "short-circuit skips the ladder and the judge, not the retry budget" is the
  right resolution — better than the reject-or-warn I suggested.
- **(d3) legacy contract** — `effectiveLevelingOutputPolicy` (`leveling_config.go:512`) now
  errors, with the load-time half at `workflow_config_leveling.go:308` sharing one wording.
  The probe that printed `effective schema = ""` now prints a rejection. The raw-proto test
  cases matter here — that's the path a loader check alone wouldn't have caught.
- **(d4) warn-and-continue** — stated on `PipelineStage` (`:134`), `AgentTask` (`:456`),
  `LevelingPolicy` (`:523`), the YAML doc-comment and the reference doc. That's what I asked
  for; the semantics themselves are a defensible design choice.

Minors: all fixed except the Python probes in `docs/experiments/` (style, your call). The
rung and judge spans, the `ctx.Err()` check, the unknown-tier warning text, the nil-`Execute`
guard, the Bedrock `TestBaseModelID` cases, the `RETRY_SESSION_MODE_ESCALATE` comment, the
`include_valid_values` tolerance, and the 388-line `docs/reference/workflow-leveling.md`
(linked from `_index.md`) all check out. The exported-API break is now disclosed in the PR
body rather than shimmed — fine by me.

Verification here: build / vet / `-race` on `pkg/orchestration`, `pkg/llm/{catalog,factory}`
and `pkg/agent` all green, `golangci-lint` 0 issues, `gofmt` clean, `buf lint` +
`buf breaking` clean, `buf generate` no drift.

Five small things, none blocking:

1. **A short-circuiting `mid` primary now costs 2 calls by default, not 1.** Verified: with
   no `tier_policies`, `frontier -> 1`, `unknown -> 1`, `mid -> 2`, and you updated
   `TestLevelingMidShortCircuitConfigurable` accordingly. It's documented on the proto and in
   the reference doc, but the commit message's "the historical single call is unchanged
   unless configured" is true only for `frontier`/`unknown` — and `mid` is the paid tier most
   likely to hit it. Worth an explicit line in the plan doc's "Behavior changes for existing
   configs".
2. **The new docs claim v1.3.0** (`workflow-leveling.md:3,5`, `workflow-output-retry.md:3,52`
   — "v1.3.0 is the first release in which the workflow YAML loader parses them") while
   `VERSION` is `1.4.0`. The badge matches the rest of `docs/reference/`, but those two
   sentences assert a release that shipped without this.
3. **`leveling_executor.go:221` still says the disabled path is "byte-for-byte the behavior
   of not having this executor at all"**, which the new whole-ladder nil-`Execute` guard
   above it contradicts (its own comment notes it fires when disabled). Confirmed: a disabled
   policy with a nil-`Execute` rung now errors. No in-repo impact — the executors gate on
   `GetEnabled()` — but the two comments should agree.
4. **`truncateForError` slices bytes** and is now applied to free-form judge text on a span
   attribute (`:617`). A non-ASCII reason cut at 200 bytes can split a rune, and OTLP string
   fields require valid UTF-8. `utf8.ValidString` trim or a rune-aware cut would close it.
5. **`output_validator.go:232`** still derefs `result.Output` when an `ExecuteFunc` returns
   `(nil, nil)`. Pre-existing and unreachable from the shipped rung builders, but it's now
   the one uncovered case of the exact class the nil-`Execute` guard was added for.

Nice work on this round — the fixes go further than the findings did in a few places (the
three-route same-model check, the shared conflict wording across load and execution, and the
rung-span nesting through `mergeContext`).

PR #320, opened 2026-08-17
feat(longmemeval): anchor replayed conversations via occurred_at + judge-model backends for the official scorer

ilsiepotamus:
## Summary
- Wire `WeaveRequest.occurred_at` (added in #308) into the LongMemEval harness: haystack session dates on ingest Weaves, question date on the question Weave, all three modes. Without this, post-#306 arrival-time anchoring stamps every replayed memory with the run date.
- `--occurred-at` flag (default `true`); server must set `server.allow_time_override: true` (documented in the guide).
- Extend the vendored official scorer with `bedrock` (boto3 converse) and `azure` (AzureOpenAI) judge backends — the paper's judge prompts are unchanged; the judge model is a disclosed parameter. Add `requirements.txt`.

## Validation
- `go test -tags fts5 -race ./cmd/loom-longmemeval/` passes; gofmt clean.
- Fresh temporal-reasoning x30 (oracle, ingest, Bedrock Opus 4.6, post-#306 main): 93.3% on the official metric — no regression vs the April 80% baseline, and anchored runs answer ~6x faster (51s vs ~5min/entry) with far less answer-time tool use.
- A/B with `--occurred-at=false` on the same 30: 100% official metric, confirming the delta is answer-time efficiency, not accuracy, on the oracle set.

### DIFF
--- a/cmd/loom-longmemeval/main.go
+++ b/cmd/loom-longmemeval/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -375,18 +376,21 @@ func runBenchmark(cmd *cobra.Command, args []string) error {
 
 	// Run benchmark and collect results
 	resultCh := make(chan EntryResult, len(entries))
+	runErrCh := make(chan error, 1)
 	startTime := time.Now()
 
 	go func() {
-		if err := runner.Run(ctx, entries, resultCh); err != nil {
-			logger.Error("runner error", zap.Error(err))
-		}
+		runErrCh <- runner.Run(ctx, entries, resultCh)
 		close(resultCh)
 	}()
 
 	var results []EntryResult
 	for r := range resultCh {
 		results = append(results, r)
+	}
+	runErr := <-runErrCh
+	if runErr != nil {
+		logger.Error("runner error", zap.Error(runErr))
 	}
 
 	elapsed := time.Since(startTime)
@@ -411,6 +415,14 @@ func runBenchmark(cmd *cobra.Command, args []string) error {
 	// Print summary
 	summary := Summarize(results, mode, serverAddr)
 	PrintSummary(summary)
+
+	// An aborted run (e.g. the server rejects occurred_at) must exit
+	// non-zero so callers don't mistake a results file full of failed rows
+	// for a completed run. A user interrupt (SIGINT/SIGTERM) still exits
+	// cleanly with whatever finished.
+	if runErr != nil && !errors.Is(runErr, context.Canceled) {
+		return fmt.Errorf("benchmark run aborted: %w", runErr)
+	}
 
 	return nil
 }
--- a/cmd/loom-longmemeval/main.go
+++ b/cmd/loom-longmemeval/main.go
@@ -63,6 +63,7 @@ var (
 	questionTypes string
 	mode          string
 	isolate       bool
+	useOccurredAt bool
 )
 
 func main() {
@@ -171,6 +172,7 @@ Three benchmark modes:
 	cmd.Flags().StringVar(&questionTypes, "types", "", "Comma-separated question types to include (empty = all)")
 	cmd.Flags().StringVar(&mode, "mode", "ingest", "Run mode: ingest (default), multi-session, or context-stuffing")
 	cmd.Flags().BoolVar(&isolate, "isolate", true, "Create a fresh agent per entry for graph memory isolation (default: true)")
+	cmd.Flags().BoolVar(&useOccurredAt, "occurred-at", true, "Send haystack/question dates as WeaveRequest.occurred_at so memories anchor at historical dates (requires server.allow_time_override: true)")
 
 	return cmd
 }
@@ -341,16 +343,18 @@ func runBenchmark(cmd *cobra.Command, args []string) error {
 		zap.Int("entries", len(entries)),
 		zap.Int("concurrency", concurrency),
 		zap.Bool("isolate", isolate),
+		zap.Bool("occurred_at", useOccurredAt),
 	)
 
 	// Create runner (connects to running Loom server via gRPC)
 	runner, err := NewRunner(RunConfig{
-		Mode:        RunMode(mode),
-		ServerAddr:  serverAddr,
-		AgentID:     agentID,
-		Concurrency: concurrency,
-		Verbose:     verbose,
-		Isolate:     isolate,
+		Mode:          RunMode(mode),
+		ServerAddr:    serverAddr,
+		AgentID:       agentID,
+		Concurrency:   concurrency,
+		Verbose:       verbose,
+		Isolate:       isolate,
+		UseOccurredAt: useOccurredAt,
 	}, logger)
 	if err != nil {
 		return err
--- a/cmd/loom-longmemeval/runner.go
+++ b/cmd/loom-longmemeval/runner.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 // RunMode determines how conversation history is processed.
@@ -60,6 +61,13 @@ type RunConfig struct {
 	Concurrency int
 	Verbose     bool
 	Isolate     bool // create a fresh agent per entry for graph memory isolation
+
+	// UseOccurredAt sends each haystack session's date (and the question date)
+	// as WeaveRequest.occurred_at. Without it, the server anchors temporal
+	// grounding and graph-memory extraction to the arrival time (the run day),
+	// which contradicts the historical dates inside the replayed conversations.
+	// Requires server.allow_time_override: true on the target server.
+	UseOccurredAt bool
 }
 
 // EntryResult holds the result of evaluating a single entry.
@@ -230,13 +238,26 @@ func (r *Runner) runEntry(ctx context.Context, entry Entry) EntryResult {
 		agentID = tempAgentID
 	}
 
+	// The question turn anchors at the entry's question date, matching the
+	// "Current date:" line in the prompt.
+	var questionAt time.Time
+	if r.config.UseOccurredAt {
+		t, err := ParseDate(entry.QuestionDate)
+		if err != nil {
+			result.Error = fmt.Sprintf("parse question date %q: %v", entry.QuestionDate, err)
+			result.Duration = time.Since(start)
+			return result
+		}
+		questionAt = t
+	}
+
 	switch r.config.Mode {
 	case ModeIngest:
-		result = r.runIngestWith(ctx, entry, sessions, result, agentID)
+		result = r.runIngestWith(ctx, entry, sessions, result, agentID, questionAt)
 	case ModeMultiSession:
-		result = r.runMultiSessionWith(ctx, entry, sessions, result, agentID)
+		result = r.runMultiSessionWith(ctx, entry, sessions, result, agentID, questionAt)
 	case ModeContextStuffing:
-		result = r.runContextStuffingWith(ctx, entry, sessions, result, agentID)
+		result = r.runContextStuffingWith(ctx, entry, sessions, result, agentID, questionAt)
 	default:
 		result.Error = fmt.Sprintf("unknown mode: %s", r.config.Mode)
 	}
@@ -275,14 +296,33 @@ func (r *Runner) deleteTempAgent(ctx context.Context, agentID string) {
 	}
 }
 
-// weave sends a Weave RPC and accumulates token usage into result.
-func (r *Runner) weave(ctx context.Context, sessionID, query, agentID string, result *EntryResult) (*loomv1.WeaveResponse, error) {
+// buildWeaveRequest constructs a Weave request, attaching occurred_at when
+// a non-zero time is given so replayed turns anchor at their historical date.
+func buildWeaveRequest(sessionID, query, agentID string, occurredAt time.Time) *loomv1.WeaveRequest {
 	req := &loomv1.WeaveRequest{
 		Query:          query,
 		SessionId:      sessionID,
 		TimeoutSeconds: 120,
 		AgentId:        agentID,
 	}
+	if !occurredAt.IsZero() {
+		req.OccurredAt = timestamppb.New(occurredAt)
+	}
+	return req
+}
+
+// sessionOccurredAt returns the haystack session's date as the occurred_at
+// override, or the zero time when the override is disabled.
+func (r *Runner) sessionOccurredAt(s SessionWithDate) time.Time {
+	if r.config.UseOccurredAt {
+		return s.ParsedAt
+	}
+	return time.Time{}
+}
+
+// weave sends a Weave RPC and accumulates token usage into result.
+func (r *Runner) weave(ctx context.Context, sessionID, query, agentID string, occurredAt time.Time, result *EntryResult) (*loomv1.WeaveResponse, error) {
+	req := buildWeaveRequest(sessionID, query, agentID, occurredAt)
 
 	resp, err := r.client.Weave(ctx, req)
 	if err != nil {
@@ -321,7 +361,7 @@ func (r *Runner) deleteSession(ctx context.Context, sessionID string) {
 // runIngestWith processes an entry by feeding sessions through Weave, then querying.
 // All turns happen in a single session so the conversation history accumulates,
 // compression triggers, and the agent can search back through everything.
-func (r *Runner) runIngestWith(ctx context.Context, entry Entry, sessions []SessionWithDate, result EntryResult, agentID string) EntryResult {
+func (r *Runner) runIngestWith(ctx context.Context, entry Entry, sessions []SessionWithDate, result EntryResult, agentID string, questionAt time.Time) EntryResult {
 	sessionID, err := r.createSession(ctx, fmt.Sprintf("lme-%s", entry.QuestionID), agentID)
 	if err != nil {
 		result.Error = fmt.Sprintf("create session: %v", err)
@@ -337,7 +377,7 @@ func (r *Runner) runIngestWith(ctx context.Context, entry Entry, sessions []Sess
 			i+1, sess.Date, FormatSession(sess),
 		)
 
-		_, err = r.weave(ctx, sessionID, ingestMsg, agentID, &result)
+		_, err = r.weave(ctx, sessionID, ingestMsg, agentID, r.sessionOccurredAt(sess), &result)
 		if err != nil {
 			result.Error = fmt.Sprintf("ingest session %d: %v", i, err)
 			return result
@@ -354,7 +394,7 @@ func (r *Runner) runIngestWith(ctx context.Context, entry Entry, sessions []Sess
 		entry.QuestionDate, entry.Question,
 	)
 
-	resp, err := r.weave(ctx, sessionID, questionMsg, agentID, &result)
+	resp, err := r.weave(ctx, sessionID, questionMsg, agentID, questionAt, &result)
 	if err != nil {
 		result.Error = fmt.Sprintf("ask question: %v", err)
 		return result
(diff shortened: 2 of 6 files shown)


ilsiepotamus:
Fresh full review (independent reviewer + my verification of every finding): the core anchoring design verified coherent end to end — the `allow_time_override` gate is default-off and rejects at all four Weave entry points, hostile-but-valid timestamps can only mis-stamp what they're designed to stamp, live session loads stay seq-ordered, and the judge prompts were programmatically diffed byte-identical to upstream LongMemEval. Six LOW findings, all fixed in `4d04f1d7`: the guide's claim that `--occurred-at=false` degrades temporal reasoning contradicted this PR's own A/B (100% vs 93.3% — the flag is now framed as mechanism correctness, not a measured accuracy win); the official scorer's backends are now documented with usage + requirements; the 64-token judge reply budget (upstream: 10) is disclosed in the guide and script header — it interacts with upstream's substring verdict check in a score-inflating direction; the bedrock client now respects the full AWS region chain instead of forcing us-west-2 over profile/SSO regions; and a server missing the override now aborts the run on the first rejection instead of burning one futile temp-agent cycle per entry. One dormant item noted, not fixed here (it also exists on main and belongs to a rebase): `LoadMessagesForAgent`/parent-session loads order by second-granularity timestamp, which under replayed identical timestamps could split tool-call pairs — zero production callers today.

docs/guides/longmemeval-benchmark.md
```diff
@@ -37,9 +37,30 @@ The six question categories are:
...
+The official metric comes from `scripts/longmemeval-eval/evaluate_qa.py` (adapted from
+[xiaowu0162/LongMemEval](https://github.com/xiaowu0162/LongMemEval); judge prompts are byte-identical to upstream):
+
+```bash
+pip install -r scripts/longmemeval-eval/requirements.txt
+# judge model backends: gpt-5.1 (OpenAI, OPENAI_API_KEY), azure-gpt
+# (AZURE_OPENAI_ENDPOINT/_API_KEY/_DEPLOYMENT_ID), or
+# claude-opus-4-6-bedrock (AWS default credential chain)
+python3 scripts/longmemeval-eval/evaluate_qa.py claude-opus-4-6-bedrock \
+    results/hypotheses.jsonl data/longmemeval_oracle.json
```
Ed-Jaramillo: This command has the arguments in upstream's order, but our vendored copy reversed them. `evaluate_qa.py` parses `hyp_file = sys.argv[1]`, `ref_file = sys.argv[2]`, `metric_model_short = sys.argv[3]` and prints its own usage as `Usage: python evaluate_qa.py hyp_file ref_file [metric_model]`.

So as written, `metric_model_short` becomes `data/longmemeval_oracle.json`, which isn't in `model_zoo` -- the script prints "Requested metric model is not supported: data/longmemeval_oracle.json", exits, and because it's a bare `exit()` the status is still 0. Nothing gets scored, and the error points a reader at the wrong thing (it names the reference dataset as an unsupported "model").

I checked upstream to be sure the difference is real -- upstream `main` uses `metric_model_short = sys.argv[1]; hyp_file = sys.argv[2]; ref_file = sys.argv[3]`, so the docs match upstream and not this fork.

```bash
python3 scripts/longmemeval-eval/evaluate_qa.py \
    results/hypotheses.jsonl data/longmemeval_oracle.json claude-opus-4-6-bedrock
```

Worth noting the likely root cause: the reversed argument order is the one deviation from upstream that *isn't* listed in the script's own "Changes from upstream" docstring. Since this PR is what formalises the scorer's usage docs, would you mind adding it? Something like:

```
  - Argument order reversed vs upstream: hyp_file ref_file [metric_model]
    (upstream: metric_model hyp_file ref_file); metric_model is optional.
```

cmd/loom-longmemeval/runner.go
```diff
@@ -230,13 +257,26 @@ func (r *Runner) runEntry(ctx context.Context, entry Entry) EntryResult {
...
+	var questionAt time.Time
+	if r.config.UseOccurredAt {
+		t, err := ParseDate(entry.QuestionDate)
+		if err != nil {
+			result.Error = fmt.Sprintf("parse question date %q: %v", entry.QuestionDate, err)
+			result.Duration = time.Since(start)
+			return result
+		}
+		questionAt = t
+	}
```
Ed-Jaramillo, line 297: `questionAt` is computed for every mode, so `--mode context-stuffing` now sends `occurred_at` too. The mode table documents that mode as the baseline where no memory system is involved -- all history goes into the prompt -- so the anchoring buys nothing there, but it does mean the baseline hard-aborts on a stock server (`server.allow_time_override` defaults to `false` at `cmd/looms/config.go:1098`) where the same command worked before this PR.

Since context-stuffing exists to be the easy control, could we leave `questionAt` zero for it?

```go
if r.config.UseOccurredAt && r.config.Mode != ModeContextStuffing {
```

Alternatively, if you'd rather keep it uniform, a line in the mode table saying the baseline also needs the server flag would cover it.

cmd/loom-longmemeval/runner.go
```diff
@@ -170,7 +187,14 @@ loop:
 			defer wg.Done()
 			defer func() { <-sem }()
 
-			result := r.runEntry(ctx, e)
+			result := r.runEntry(runCtx, e)
+			if strings.Contains(result.Error, "allow_time_override") {
```
Ed-Jaramillo: Minor/nit: this matches a substring of an already-formatted string -- `result.Error` came from `fmt.Sprintf("ingest session %d: %v", i, err)`, so the gRPC status code is gone by the time the check runs, and the abort is coupled to the exact wording at `pkg/server/occurred_at.go:39`. If that message is ever reworded the abort silently stops firing (degrading to the pre-PR behaviour, so low stakes).

If you wanted to harden it, carrying the error or its `codes.Code` on `EntryResult` would let you match `codes.FailedPrecondition` *and* the substring. Entirely optional.

cmd/loom-longmemeval/runner.go
```diff
@@ -170,7 +187,14 @@ loop:
...
+				msg := "server rejects occurred_at (server.allow_time_override is not enabled); aborting the run — enable it in looms.yaml or pass --occurred-at=false"
+				if abortErr.CompareAndSwap(nil, &msg) {
+					r.logger.Error(msg, zap.String("first_failed_entry", e.QuestionID))
+					abort()
+				}
```
Ed-Jaramillo, line 203: Good call aborting rather than burning 500 futile cycles -- one wrinkle in the cleanup path though.

`runEntry` does `defer r.deleteTempAgent(ctx, tempAgentID)` on the same context `abort()` cancels. So with the default `--concurrency 3` against a stock server: all three goroutines create their `lme-tmp-<qid>` agents, all three Weaves return `FAILED_PRECONDITION`, the first to finish calls `abort()`, and any sibling that hasn't yet reached its deferred delete now runs `DeleteAgent` on a cancelled context. That fails, gets logged as a warning, and leaves up to `--concurrency - 1` orphaned agents (plus their graph-memory stores) on the server.

Since the abort exists to *avoid* wasted create/delete cycles, having it strand server-side state cuts against the goal -- and because it triggers on the default server config, it's the common first-run path rather than an edge case.

There's already precedent for the fix in `pkg/shuttle/admission_ask.go:315`:

```go
cctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 10*time.Second)
defer cancel()
r.deleteTempAgent(cctx, tempAgentID)
```

Same applies to `deleteSession` in the context-stuffing path.

Separately and much smaller: the loop's `select` has both `<-runCtx.Done()` and `sem <- struct{}{}`, so once `abort()` fires Go picks between them at random and roughly one or two more entries start on the cancelled context. They fail fast at `createTempAgent` (no leak), but they do land in `results.jsonl` as `context canceled` rows. A `runCtx.Err() != nil` check before the `select` would make the abort crisp.

cmd/loom-longmemeval/main.go
```diff
@@ -341,16 +343,18 @@ func runBenchmark(cmd *cobra.Command, args []string) error {
...
-		Concurrency: concurrency,
-		Verbose:     verbose,
-		Isolate:     isolate,
+		Mode:          RunMode(mode),
+		ServerAddr:    serverAddr,
+		AgentID:       agentID,
+		Concurrency:   concurrency,
+		Verbose:       verbose,
+		Isolate:       isolate,
+		UseOccurredAt: useOccurredAt,
```
Ed-Jaramillo, line 358: Anchoring here because the relevant code is just below the diff, at `main.go:377`:

```go
go func() {
    if err := runner.Run(ctx, entries, resultCh); err != nil {
        logger.Error("runner error", zap.Error(err))
    }
    close(resultCh)
}()
```

`Run`'s new abort error is logged and then discarded -- `runBenchmark` still returns `nil`. So an aborted run writes `results.jsonl` (the `--output` default) containing N rows that are all error/empty-hypothesis, prints `Entries: 0 succeeded / N failed`, and exits 0. Interactively you'd catch that from the summary line, but a `just` recipe or CI wrapper wouldn't -- and the write clobbers any previous good `results.jsonl` at the default path.

To be clear, the swallowing predates this PR; it's just that `Run` now has something meaningful to say, which is what makes it worth threading out:

```go
runErrCh := make(chan error, 1)
go func() { runErrCh <- runner.Run(ctx, entries, resultCh); close(resultCh) }()
// ... drain resultCh, write output, print summary ...
return <-runErrCh
```

scripts/longmemeval-eval/evaluate_qa.py
```diff
@@ -23,9 +32,33 @@
...
+    import random
+    import time
+    from botocore.exceptions import ClientError
+    retryable = ('ThrottlingException', 'ServiceUnavailableException', 'ModelNotReadyException')
+    for attempt in range(max_tries):
+        try:
+            resp = client.converse(
+                modelId=model_id,
+                messages=[{'role': 'user', 'content': [{'text': prompt}]}],
+                inferenceConfig={'maxTokens': 64, 'temperature': 0},
```
Ed-Jaramillo: Really appreciate the explicit disclosure of this deviation, including which direction the bias runs -- that's the part most benchmark PRs leave out.

One small consequence: the 64-token budget now lives in two places -- here, and `'max_completion_tokens': 64` in the OpenAI/Azure `kwargs` further down. So the guide's "flip a single constant in the script to reproduce upstream exactly" is no longer accurate, and the two paths can drift independently. A module-level constant would make the sentence true again and keep them in lockstep:

```python
JUDGE_MAX_TOKENS = 64  # upstream LongMemEval uses 10
```

For what it's worth, I diffed `get_anscheck_prompt` against upstream `main` to check the "prompts unchanged" claim -- all the template literals are byte-identical, and the only difference in that function is a removed trailing space on a `prompt = template.format(...)` line. Claim holds.

cmd/loom-longmemeval/runner_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestBuildWeaveRequestFromDatasetDate(t *testing.T) {
+	// The exact format the LongMemEval dataset uses for haystack/question dates.
+	parsed, err := ParseDate("2023/04/10 (Mon) 17:50")
+	require.NoError(t, err)
+
+	req := buildWeaveRequest("sess-1", "query", "agent-1", parsed)
+	require.NotNil(t, req.OccurredAt)
+	assert.Equal(t, parsed.UTC(), req.OccurredAt.AsTime())
```
Ed-Jaramillo: Nit: `assert.Equal` on `time.Time` compares the struct internals via `reflect.DeepEqual` rather than comparing instants, so it can in principle fail for two equal times with different internal representations. It passes today (I ran the package), and the two sibling tests already use the robust form -- so this is really just consistency:

```go
assert.True(t, parsed.Equal(req.OccurredAt.AsTime()))
```

Ed-Jaramillo (commented):
Reviewed at head SHA `4d04f1d7`. I tried to verify the mechanism end to end rather than read the diff in isolation, so most of this is "here's what I checked" rather than "here's what I think."

**One thing I'd like fixed before merge** (the documented scorer command can't run — details inline on the docs block). Everything else is non-blocking, and a couple of items are questions rather than findings. Leaving this as a comment rather than a formal changes-request since it's a one-line docs fix.

---

## What I verified

| Claim | Status | How |
|---|---|---|
| `occurred_at` reaches every persisted row | **VERIFIED** | `applyOccurredAt` → `agent.WithOccurredAt` → `appendMessage` sets `msg.Timestamp = at` unconditionally (`pkg/agent/agent.go:2003-2008`) |
| Anchoring actually changes extraction | **VERIFIED** | `graph_memory_extractor.go:119-120` renders `msg.Timestamp` into the extraction prompt; `newestMessageDate` (`:504-511`) anchors on the newest non-zero `Timestamp` |
| Gate is default-off and rejects at every entry point | **VERIFIED** | `config.go:1098` (`false`); `occurred_at.go:37-40` returns `FailedPrecondition`; called from `server.go:126,190` and `multi_agent.go:832,1041` |
| The abort trigger fires | **VERIFIED** | Server message at `occurred_at.go:39` does contain the `allow_time_override` substring that `runner.go:191` matches |
| Judge prompts unchanged from upstream | **VERIFIED** | Fetched upstream `main` and diffed `get_anscheck_prompt`. All prompt template literals byte-identical; only delta is one removed trailing space on a `prompt = template.format(...)` statement line |
| Upstream really is `max_tokens: 10` | **VERIFIED** | Confirmed in upstream source |
| Verdict check matches upstream | **VERIFIED** | Both `label = 'yes' in eval_response.lower()`, byte-identical |
| Your disclosed dormant item ("zero production callers") | **VERIFIED** | `LoadMessagesForAgent` and `LoadMessagesFromParentSession` resolve only to the interface declaration, the two store impls, and tests. And the live path `LoadMessages` orders by `id ASC` in both SQLite (`session_store.go:808,917`) and Postgres (`:191,453,547`), so identical replayed timestamps can't reorder a real session. Accurate call. |
| Tests + gofmt | **VERIFIED** | Reproduced locally at head SHA: `go test -tags fts5 -race -count=1 ./cmd/loom-longmemeval/` → `ok, 10.6s`; `gofmt -l` and `go vet` clean |
| 93.3% / 100% / ~6x latency | **AUTHOR-REPORTED** | No hypotheses, eval-results, or timing artifacts attached, so I couldn't reproduce these. Not a defect — just flagging that they're reported rather than demonstrated. The harness does emit `duration_ms` per entry, so the latency claim is reproducible in principle. |

I also went looking for a counterexample I didn't find, so it's on the record: I expected out-of-order `occurred_at` (question date earlier than a haystack date) to be able to scramble a session, since every message in one Weave gets an *identical* timestamp. It can't — the live load path is `id ASC` in both stores. Worth knowing that the `id ASC` ordering is what's carrying that safety, not any property of the harness.

---

## Blocking-ish: the documented scorer command can't run

Full detail inline on the `bash` block. Short version: the command passes the judge model first, but our vendored copy parses `hyp_file ref_file [metric_model]`. I confirmed against upstream that *upstream* takes the model first — so the docs match upstream's convention, not this fork's. The reversed order is the one deviation from upstream that isn't in the script's own "Changes from upstream" docstring, which is very likely why it slipped.

Running it as written makes `metric_model_short` = `data/longmemeval_oracle.json`, which isn't in `model_zoo`, so it prints "Requested metric model is not supported", exits, and — because it's a bare `exit()` — returns status 0.

---

## Non-blocking

Inline on the relevant lines:

- **`--mode context-stuffing` now needs `allow_time_override` for zero benefit** — `runEntry` computes `questionAt` for every mode, so the documented no-memory baseline hard-aborts on a stock server where it used to work.
- **Abort cancels the context its own cleanup RPCs use** — leaks up to `--concurrency − 1` orphaned `lme-tmp-*` agents, on the default server config.
- **The abort error never reaches the exit code** — aborted run writes an all-failed `results.jsonl` (default path) and exits 0. This one predates the PR; it's just that `Run` now has something meaningful to say.
- **"flip a single constant" is now two constants** — `maxTokens: 64` (bedrock) and `max_completion_tokens: 64` (openai/azure).
- **Substring match on a formatted error string** couples the harness to the server's exact wording.
- **`assert.Equal` on `time.Time`** in one of the three new tests, where the siblings correctly use `.Equal()`.

Two with no line in the diff to hang off:

- **`RunSummary` doesn't record the `--occurred-at` setting.** This PR's own docs establish that the two settings score differently on the same data, so a `results.jsonl` that can't tell you which arm produced it is awkward to defend later. A single `OccurredAt bool` on `RunSummary` (`report.go:52-63`) would do it. Related: the guide's "Results We Observed" table predates anchoring and doesn't say so — a one-line provenance note would clear that up.
- **`global.anthropic.claude-opus-4-6-v1` has no row in `docs/reference/llm-bedrock.md`.** That table lists `us.anthropic.claude-opus-4-6-v1` and a `global.` variant for 4.7 only. The string already appears in the benchmark guide, so this is consistent with the repo — but promoting it into an executable `model_zoo` entry makes it worth a table row, or a note that it depends on the account's enabled inference profiles.

---

## Two questions

**1. Defaulting `--occurred-at=true` when the measured A/B favours `false`.** I think your reasoning holds, and I can add support for it: the 93.3% vs 100% delta is 2 entries out of 30 — two-tailed Fisher exact **p ≈ 0.49**, i.e. indistinguishable from noise at that n. Might be worth putting that in the guide, because "scored slightly HIGHER with the override off" currently reads as a real effect the default is overriding on principle, when the honest reading is that there's no measurable difference yet. That's a stronger position than the one the text currently takes.

**2. Does the default `gpt-5.1` judge emit a verdict within 64 tokens?** On OpenAI reasoning models `max_completion_tokens` bounds reasoning *plus* visible output — if it's spent on reasoning, `content` could come back empty, making `label = 'yes' in ''` → `False` for every entry and silently reporting 0%. I couldn't verify OpenAI's token accounting from my environment, so this is a genuine unknown, not an assertion. Both the `gpt-5.1` entry and the `64` are pre-existing on `main` and this PR only re-indents that block — raising it only because this PR is what documents how to run the scorer, and the omitted-model default is the path a reader is most likely to hit. Probably a follow-up issue rather than anything for this PR.

---

## Things I thought were done well

- **The abort is the right instinct and cleanly built** — `atomic.Pointer[string]` + `CompareAndSwap` to elect exactly one reporter, derived cancel context, error surfaced after `wg.Wait()`. And the message names both remedies (`looms.yaml` or `--occurred-at=false`), which is exactly what someone in that hole needs.
- **The zero-`time.Time` sentinel is a good call.** `buildWeaveRequest` never constructs a `timestamppb` on the disabled path, so a mis-set flag can't accidentally send a bogus timestamp. `sessionOccurredAt` makes the intent legible at each call site.
- **The tests test behaviour, not shape.** `TestBuildWeaveRequestFromDatasetDate` pushes the *actual dataset date format* through `ParseDate` into the request — that's the thing that would actually break.
- **The disclosure discipline is well above average.** Documenting that your own A/B scored *lower* with the flag on, and that the judge budget deviates from upstream in a score-inflating direction, is the opposite of what benchmark PRs usually do. It also made this review much faster and let me verify instead of guess.
- **`requirements.txt` is exactly right** — all five pins are genuinely imported, including `numpy`, which is easy to miss since it's only used at the accuracy-aggregation lines.
- **The Bedrock retry helper is appropriately narrow** — bounded tries, jittered backoff capped at 30s, explicit retryable-code allowlist rather than a blanket `except ClientError`.

ilsiepotamus:
All review items addressed in `83815523` — the must-fix and every non-blocking item.

1. **Scorer command (must-fix)**: the guide now matches the vendored script's argv order (`evaluate_qa.py <hyp_file> <ref_file> [metric_model]`), verified against the script's own usage string.
2. **Baseline sends occurred_at**: `context-stuffing` no longer sends it — new `occurredAtEnabled()` gates both the question and session timestamps; the baseline runs against a stock server again. End-to-end pinned by `TestRunEntryContextStuffingSendsNoOccurredAt` / `TestRunEntryIngestSendsOccurredAt` through `runEntry` with a fake client.
3. **Abort stranding temp agents**: cleanup now runs on `context.WithoutCancel` with its own 30s timeout, so siblings' deferred deletes survive the abort. `TestRunAbortsOnTimeOverrideRejection` runs at concurrency 3 and asserts every `DeleteAgent` saw a live context.
4. **Swallowed Run error**: the runner goroutine's error is threaded out via a buffered channel and returned from `runBenchmark` — an aborted run exits non-zero (SIGINT still exits cleanly). Outputs/summary still written first so partial results stay inspectable.
5. **Fragile abort trigger**: `EntryResult` now carries the Weave call's gRPC code (unexported, never serialized); the abort matches `FailedPrecondition` + an `occurred_at` mention, keeping the bare substring only as fallback. `TestIsTimeOverrideRejection` covers the cases including an unrelated `FailedPrecondition` not aborting.
6. **Judge budget in two places**: single `JUDGE_MAX_TOKENS = 64` module constant feeds both the bedrock and OpenAI/Azure paths; the guide's "flip one constant" sentence is true again and names it.
7. **time.Time assert**: switched to `parsed.Equal(...)` like the sibling tests.

Validation: `gofmt` clean, `go build -tags fts5 ./...`, `go vet`, and `go test -tags fts5 -race ./cmd/loom-longmemeval/...` all green; the Python script parses.

One heads-up: #331 (stacked on this branch) introduces a separate `server.allow_assistant_override` gate for its replay field rather than renaming `allow_time_override` — the time flag already shipped in v1.4.0 via #308, so the rename you floated there would break released config. Details in the #331 thread.

PR #321, opened 2026-08-17
fix(server): judge/registry/eval-store wiring no longer requires a provider pool

ilsiepotamus:
## Summary
`cmd_serve.go` gated judge wiring — plus registry task-manager, graph-memory, suppressed-tools, and eval-store injection — inside `if providerPool != nil` (introduced with #100's pool refactor; later PRs appended more wiring inside the gate). Servers configured with only the standard `llm:` block get `FAILED_PRECONDITION: no LLM provider available for evaluation` on every judge call, and registry-built agents silently miss the task/graph-memory/eval subsystems.

This hoists the non-pool-specific wiring out of the gate. The judge falls back to the server's default provider when no `providers:` pool exists; the active pool provider still wins when configured.

## How found
LongMemEval scoring (`loom-longmemeval score`) failed with the FAILED_PRECONDITION above on a single-provider (Bedrock-only) server. April's benchmark runs predate the gating, which is why this wasn't seen earlier.

## Validation
- `go test -tags fts5 -race ./cmd/looms/` passes; gofmt clean.
- On a single-provider server: startup now logs "Judge server wired into loom service"; `score` over a 30-entry result set completes with zero judge errors (previously 30/30 errors).

### DIFF
--- a/cmd/looms/cmd_serve.go
+++ b/cmd/looms/cmd_serve.go
@@ -373,6 +373,53 @@ type judgeServerSetter interface {
 	SetJudgeServer(js *server.JudgeServer)
 }
 
+// registrySubsystemSink is the subset of *agent.Registry that receives the
+// server-level subsystems. Extracted so wireRegistrySubsystems is unit-testable
+// with a spy.
+type registrySubsystemSink interface {
+	SetTaskManager(manager *task.Manager, decomposer *task.Decomposer)
+	SetGraphMemoryStore(store memory.GraphMemoryStore, embedder memory.Embedder)
+	SetSuppressedBuiltinTools(names []string)
+}
+
+// wireRegistrySubsystems injects the server-level subsystems (task manager,
+// graph-memory store, tool-surface policy) into the agent registry so that
+// registry-built agents — gRPC-created, config-hot-reloaded, or benchmark
+// --isolate temp agents — receive them.
+//
+// This MUST run independently of whether a provider pool is configured. These
+// injections previously lived inside the `providerPool != nil` branch of the
+// serve wiring, so a single-provider server silently gave every registry-built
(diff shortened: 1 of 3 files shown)


ilsiepotamus:
Fresh full review found the wiring complete and correctly ordered (everything runs before any listener; nil-provider judge fails clean with FailedPrecondition), but flagged a HIGH merge-coordination issue: this PR and #342 rewrote the same pool-branch region of `cmd_serve.go`, and a naive both-merge would double-run `evals.NewStore` (leaking a sqlite handle). Rebuilt as a true delta in `2f884b85`: the branch now stacks on #342 (base retargeted — GitHub flips it back to main when #342 merges), keeps #342's `wireRegistrySubsystems` helper, hoists the provider pool out of its if-scope (its build error is now logged with the actual error instead of swallowed), and moves eval-store + judge wiring to unconditional blocks with `resolveJudgeFallback` — which also fixes the pre-existing unchecked pool lookup that handed the judge a nil provider on a misspelled `active_provider`. Added `TestResolveJudgeFallback` (no-pool / pool-hit / misspelled-active / nil-entry) — the review's no-tests finding. One inherited observation for a future look: `evals.NewStore` eagerly opens/migrates at startup on every deployment now, defaulting into the main DB file — pre-existing path choice, flagged, not changed.

ilsiepotamus:
Reopened: this PR was auto-closed on 2026-08-21 as a side effect of closing #342 — its base had been retargeted onto #342's branch (per the stacking note above), so deleting that branch closed this PR too, orphaning the fix. Base is now main; the diff is the complete fix (#342's registry-subsystem wiring commit plus the judge/eval-store delta), which is what was always intended to land once the stack resolved.

PR #322, opened 2026-08-17
fix(prompts): stop sanitized backtick remnants reassembling a code fence

ilsiepotamus:
## What

`FuzzPromptInterpolation` has been intermittently failing CI (most recently on #316, previously on #315, where it was taken for a flaky timeout). It is not flake and not a timeout — the fuzzer keeps rediscovering a real sanitizer bug. Deterministic repro:

```
template: {{.a}}{{.a}}
value:    ````` (5 backticks)
result:   ````  (4 backticks — contains ```)
```

**Mechanism**: `sanitizePromptInjection` replaces each ``` with spaces, then `escapeString` steps 6–7 (space collapse + trim) eat exactly those spaces. A value of 5 backticks escapes to two backticks, and two adjacent interpolations concatenate the remnants into a new fence.

## Fix

`escapeString` gains step 8: strip backticks (and interleaved spaces) from the value's edges. Pattern removal already caps interior backtick runs below fence length, so with clean edges an escaped value can never contribute to a fence — alone, adjacent to another value, or adjacent to template backticks. Interior backticks (`a``b`) are preserved.

## Why the fuzz property also changes

The old Property 3 (`pattern ∈ value && pattern ∈ result → fail`) is unsatisfiable as written, independent of any sanitizer:

1. **Trusted templates keep their own fences.** `template="```{{.a}}"`, `value="```"` — the template's fence must survive (templates legitimately contain code fences), but the property demands no fence anywhere in the result. The fuzzer would have found this eventually; it's a two-mutation input.
2. **Text-pattern fragments straddle boundaries.** A value whose escaped form ends in `Sys` and starts with `tem:` reassembles `System:` when interpolated twice adjacently, even though the escaped value contains no pattern. No per-value transformation can prevent two innocent fragments from meeting.

Replaced with two provable properties:
- **(a) Per value**: `escapeString(value)` never contains a dangerous pattern (mirrors `FuzzEscapeString` Property 5, now asserted at the interpolation level too).
- **(b) Whole result, fences only**: `count(result, "```") ≤ count(baseline, "```")` where baseline interpolates every variable as empty — values can never *add* fences beyond what the trusted template text itself joins into. Provable from (edge-stripped values + interior runs ≤ 2) by run-merge monotonicity.

`FuzzEscapeString` gains Property 9 (no edge backticks) and matching seeds; the CI-found input is now a permanent seed.

## Scope note

`###` and `---` have the same remnant arithmetic in theory, but stripping `-` from value edges would corrupt legitimate values (negative numbers, CLI flags), and header/separator injection requires line starts that step 2 (newline flattening) already denies to values. Left alone deliberately.

## Verification

- New regression tests **fail on the old implementation** (verified by stash/run/pop) and pass on the fix
- `go test -tags fts5 -race` clean on `pkg/prompts` and every importer (`pkg/agent`, `pkg/builder`, `pkg/patterns`, `pkg/shuttle`, `pkg/shuttle/builtin`, `cmd/looms`, `cmd/loom-standalone`)
- Both fuzz targets clean for 3.3M / 4.9M execs (old code failed within ~2s)
- `golangci-lint` 0 issues, `gofmt` clean

Unblocks #316, whose only red check is this fuzzer.

### DIFF
--- a/pkg/prompts/interpolation.go
+++ b/pkg/prompts/interpolation.go
@@ -88,10 +88,15 @@ func escapeValue(value interface{}) string {
 // Implements multiple escaping strategies for production use:
 //   - Control character removal
 //   - XML/HTML entity escaping
-//   - Unicode normalization
 //   - Prompt injection pattern detection
-//   - Edge backtick stripping, so remnants cannot reassemble a ``` fence
-//     across value/template boundaries
+//   - Edge backtick/tilde stripping, so remnants cannot reassemble a
+//     ``` or ~~~ fence across value/template boundaries
+//
+// Ordering invariant: every character-DELETING step runs before the fence
+// capping in step 5, and steps 6-8 only join fields with single spaces or
+// trim edges — nothing after step 5 may transform characters (e.g. a future
+// Unicode normalization would map U+1FEF to a backtick and silently
+// resurrect the fence-injection bug).
 func escapeString(s string) string {
 	// 1. Remove null bytes and invalid UTF-8
 	s = strings.ReplaceAll(s, "\x00", "")
@@ -130,15 +135,18 @@ func escapeString(s string) string {
 	// 7. Trim leading/trailing whitespace
 	s = strings.TrimSpace(s)
 
-	// 8. Strip backticks (and any space interleaved with them) from the edges.
-	// sanitizePromptInjection caps interior backtick runs below fence length,
-	// but a remnant run at the value's edge can merge with backticks in the
-	// surrounding template or an adjacent interpolated value and reassemble a
-	// ``` fence: interpolating "`````" twice, adjacently, used to yield "````".
-	// With clean edges and short interior runs, no concatenation of escaped
-	// values with template text can form a fence the template did not already
-	// contain. Interior backticks (e.g. "a``b") are preserved.
-	s = strings.TrimFunc(s, func(r rune) bool { return r == '`' || r == ' ' })
+	// 8. Strip fence runes — backticks and tildes — (and any space interleaved
+	// with them) from the edges. sanitizePromptInjection caps interior runs
+	// below fence length, but a remnant run at the value's edge can merge
+	// with fence runes in the surrounding template or an adjacent
+	// interpolated value and reassemble a ``` or ~~~ fence: interpolating
+	// "`````" twice, adjacently, used to yield "````". With clean edges and
+	// short interior runs, no concatenation of escaped values with template
+	// text can form a fence the template did not already contain. Interior
+	// occurrences (e.g. "a``b", "x~~y") are preserved. Documented cost: a
+	// value legitimately ending in inline code ("use `+"`ls`"+`") loses its edge
+	// backtick and renders unbalanced.
+	s = strings.TrimFunc(s, func(r rune) bool { return r == '`' || r == '~' || r == ' ' })
 
 	return s
 }
@@ -147,7 +155,8 @@ func escapeString(s string) string {
 func sanitizePromptInjection(s string) string {
 	// Remove common prompt injection delimiters and commands
 	injectionPatterns := []string{
-		"```",              // Code blocks
+		"```",              // Code blocks (backtick fence)
+		"~~~",              // Code blocks (tilde fence, CommonMark)
 		"###",              // Headers
 		"---",              // Separators
 		"System:",          // System prompts
--- a/pkg/prompts/interpolation.go
+++ b/pkg/prompts/interpolation.go
@@ -86,10 +86,12 @@ func escapeValue(value interface{}) string {
 // escapeString escapes special characters to prevent prompt injection.
 //
 // Implements multiple escaping strategies for production use:
-// - Control character removal
-// - XML/HTML entity escaping
-// - Unicode normalization
-// - Prompt injection pattern detection
+//   - Control character removal
+//   - XML/HTML entity escaping
+//   - Unicode normalization
+//   - Prompt injection pattern detection
+//   - Edge backtick stripping, so remnants cannot reassemble a ``` fence
+//     across value/template boundaries
 func escapeString(s string) string {
 	// 1. Remove null bytes and invalid UTF-8
 	s = strings.ReplaceAll(s, "\x00", "")
@@ -127,6 +129,16 @@ func escapeString(s string) string {
 
 	// 7. Trim leading/trailing whitespace
 	s = strings.TrimSpace(s)
+
+	// 8. Strip backticks (and any space interleaved with them) from the edges.
+	// sanitizePromptInjection caps interior backtick runs below fence length,
+	// but a remnant run at the value's edge can merge with backticks in the
+	// surrounding template or an adjacent interpolated value and reassemble a
+	// ``` fence: interpolating "`````" twice, adjacently, used to yield "````".
+	// With clean edges and short interior runs, no concatenation of escaped
+	// values with template text can form a fence the template did not already
+	// contain. Interior backticks (e.g. "a``b") are preserved.
+	s = strings.TrimFunc(s, func(r rune) bool { return r == '`' || r == ' ' })
 
 	return s
 }
--- a/pkg/prompts/interpolation.go
+++ b/pkg/prompts/interpolation.go
@@ -22,10 +22,78 @@ import (
 	"unicode/utf8"
 )
 
+// placeholderRe matches {{.variable}} placeholders ({{.name}} syntax, like Go
+// templates but simpler).
+var placeholderRe = regexp.MustCompile(`\{\{\.(\w+)\}\}`)
+
+// injectionPatterns is the single, canonical list of prompt-injection
+// delimiters this package suppresses. Two mechanisms enforce it and MUST stay
+// on this one list so they cannot drift apart:
+//
+//  1. sanitizePromptInjection removes every occurrence inside an escaped
+//     value, so no single value can carry a pattern.
+//  2. Interpolate's junction guard (appendGuarded/junctionFormsInjection)
+//     prevents remnants from reassembling any of these patterns across the
+//     boundaries where a value meets template text or another value.
+//
+// The junction guard derives its correctness from list invariants that are
+// pinned by TestInjectionPatternInvariants: every pattern is non-empty ASCII,
+// never starts or ends with a space, and never contains two consecutive
+// spaces.
+var injectionPatterns = []string{
+	"```",              // Code blocks (backtick fence)
+	"~~~",              // Code blocks (tilde fence, CommonMark)
+	"###",              // Headers
+	"---",              // Separators
+	"System:",          // System prompts
+	"Assistant:",       // Assistant prompts
+	"Human:",           // Human prompts
+	"[INST]",           // Instruction markers
+	"[/INST]",          // Instruction markers
+	"<|im_start|>",     // Instruction markers
+	"<|im_end|>",       // Instruction markers
+	"### Instruction:", // Alpaca-style
+	"### Response:",    // Alpaca-style
+}
+
+// maxJunctionSeparators bounds appendGuarded's insertion loop, derived from
+// injectionPatterns rather than hardcoded: once the junction holds a run of
+// spaces longer than the longest run of spaces inside any suppressed pattern,
+// no pattern can straddle it. A straddling occurrence would have to either
+// contain that whole space run in its interior (impossible: no pattern holds a
+// space run that long) or start/end inside it (impossible: no pattern starts
+// or ends with a space — see TestInjectionPatternInvariants).
+var maxJunctionSeparators = func() int {
+	maxRun := 0
+	for _, p := range injectionPatterns {
+		run := 0
+		for _, r := range p {
+			if r == ' ' {
+				run++
+				if run > maxRun {
+					maxRun = run
+				}
+			} else {
+				run = 0
+			}
+		}
+	}
+	return maxRun + 1
+}()
+
 // Interpolate performs safe variable substitution in a prompt template.
 //
 // Uses {{.variable_name}} syntax (like Go templates but simpler).
 // All values are escaped to prevent prompt injection attacks.
+//
+// Escaping alone cannot stop remnants of two sanitized values (or a value and
+// the template's own text) from concatenating back into a suppressed pattern
+// — e.g. two adjacent "##" remnants forming "####" ⊃ "###", or "…Sys"+"tem:…"
+// forming "System:". Interpolate therefore guards every junction where a
+// substituted value meets template text or another substituted value: if the
+// concatenation would complete any pattern in injectionPatterns across that
+// boundary, a single benign space is inserted at the junction. Value bytes are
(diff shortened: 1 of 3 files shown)


ilsiepotamus:
Fresh full review: the backtick fix is **provably complete** — the reviewer traced the invariant (interior runs capped at ≤2, character-deleting steps all precede the capping, joins insert exactly one space, edges stripped) and a 45s bounded fuzz (18.7M execs) found nothing. Three adjacent findings, fixed in `30eabc0f`: tilde fences (`~~~`, CommonMark) passed through the sanitizer fully intact and the fuzz property only counted backticks so it could never notice — tildes now get the same capping/edge-stripping and the fuzz covers them; the documented cost of edge stripping (a value ending in inline code renders unbalanced; a lone fence rune vanishes) now has explicit tests so the tradeoff is visible; and the doc comment advertising nonexistent Unicode normalization is replaced with the ordering invariant that makes it dangerous — NFC after the capping step would map U+1FEF to a backtick and resurrect the injection.

pkg/prompts/interpolation.go
```diff
@@ -128,14 +135,28 @@ func escapeString(s string) string {
...
+	// below fence length, but a remnant run at the value's edge can merge
+	// with fence runes in the surrounding template or an adjacent
+	// interpolated value and reassemble a ``` or ~~~ fence: interpolating
+	// "`````" twice, adjacently, used to yield "````". With clean edges and
+	// short interior runs, no concatenation of escaped values with template
+	// text can form a fence the template did not already contain. Interior
+	// occurrences (e.g. "a``b", "x~~y") are preserved. Documented cost: a
+	// value legitimately ending in inline code ("use `+"`ls`"+`") loses its edge
+	// backtick and renders unbalanced.
+	s = strings.TrimFunc(s, func(r rune) bool { return r == '`' || r == '~' || r == ' ' })
```
Ed-Jaramillo: **MAJOR:** This unconditional `TrimFunc` prevents boundary fences by deleting valid data from every interpolated string. At this head, `~/bin` becomes `/bin`, `~10` becomes `10`, and balanced inline code becomes unbalanced; the public prompt registry does not document this destructive conversion. Please preserve benign edge characters and break only a boundary that would actually form a run of three or more matching fence runes. Add regressions for home-relative paths, approximation notation, and inline code alongside the existing malicious-boundary cases.

pkg/prompts/interpolation_fuzz_test.go
```diff
@@ -94,30 +102,58 @@ func FuzzPromptInterpolation(f *testing.F) {
...
+		//     Text patterns get no result-level assertion: fragments like
+		//     "Sys" and "tem:" can straddle a boundary without any single
+		//     value containing the pattern, so no per-value sanitizer can
+		//     prevent it.
+		emptyVars := make(map[string]any, len(vars))
+		for k := range vars {
+			emptyVars[k] = ""
+		}
+		baseline := Interpolate(template, emptyVars)
+		if got, allowed := strings.Count(result, "```"), strings.Count(baseline, "```"); got > allowed {
```
Ed-Jaramillo: **MAJOR:** The result-level property is backtick-only, but the same remnant reconstruction still occurs for other patterns that `sanitizePromptInjection` classifies as dangerous. `Interpolate("{{.a}}{{.a}}", {"a":"#####"})` returns `####`, `{"a":"-----"}` returns `----`, and `{"a":"tem:System:Sys"}` reconstructs `System:`. Because the trusted template can place the placeholder at line start, newline flattening does not prevent the `###`/`---` cases. Please make boundary handling and the trusted-template baseline property cover every delimiter the sanitizer promises to suppress.

Ed-Jaramillo (changes requested):
Requesting changes at `30eabc0f`.

The original five-backtick reconstruction is fixed, and the updated backtick/tilde cases pass targeted race tests and fuzzing. Two deterministic issues still need pre-merge resolution:

1. Edge cleanup deletes legitimate leading/trailing fence characters from every string value, changing values such as `~/bin` to `/bin`, `~10` to `10`, and balanced inline code to unbalanced Markdown.
2. Other patterns that `sanitizePromptInjection` classifies as dangerous still reconstruct across adjacent substitutions: `#####` yields `####`, `-----` yields `----`, and `tem:System:Sys` reconstructs `System:` with template `{{.a}}{{.a}}`.

Both counterexamples were reproduced under `go test -tags fts5 -race`. All GitHub checks are green, but they do not cover these cases. The smallest safe direction is boundary-aware, non-destructive escaping generalized to every delimiter the sanitizer promises to suppress; the inline comments include focused regression requests.

ilsiepotamus:
Both MAJORs addressed in `d3bca10d`, in the direction you suggested — boundary-aware, non-destructive, and generalized to every suppressed delimiter.

**Design**: the destructive edge `TrimFunc` is gone; per-value escaping now preserves benign edges byte-for-byte. In its place, `Interpolate` builds the result piece by piece and guards every junction where a substituted value meets template text or another value: `junctionFormsInjection` inspects, per pattern of length L, a window of the last L−1 bytes of accumulated output plus the first L−1 bytes of the incoming piece — both halves are shorter than L, so a window hit is exactly a straddling occurrence — and `appendGuarded` inserts single spaces (never deleting user data) until the junction is clean. The pattern list is hoisted to one canonical package-level `injectionPatterns` shared by the sanitizer, the junction guard, and the fuzz properties, so the two mechanisms cannot drift; the guard's termination/non-interference bound is derived from the list, with `TestInjectionPatternInvariants` pinning the invariants it rests on (pure ASCII, no edge spaces, no double spaces).

**Your counterexamples, reproduced-then-fixed** (each was a failing test against `30eabc0f` first):
- `~/bin` → stays `~/bin` (was `/bin`); `~10` → stays `~10` (was `10`); balanced inline code stays balanced
- `{{.a}}{{.a}}` with `#####` → `## ##` (was `####` ⊃ `###`); `-----` → `-- --` (was `----` ⊃ `---`); `tem:System:Sys` no longer reconstructs `System:` across the junction

**Fuzz property generalized**: the result-level property now asserts, for **every** `injectionPatterns` entry, that its count in the interpolated result never exceeds the trusted-template baseline — not backtick-only — plus new junction-coverage tests across all patterns (including `<|im_start|>` bridged by a value and Alpaca `### Instruction:`) and regressions for benign-edge preservation.

Validation: `gofmt` clean, `go build -tags fts5 ./...`, `go test -tags fts5 -race ./pkg/prompts/...` green, and all five fuzz targets ran 30s each with zero crashers (~14.5M execs on FuzzPromptInterpolation).

pkg/prompts/interpolation.go
```diff
@@ -22,11 +22,79 @@ import (
...
+//     value, so no single value can carry a pattern.
+//  2. Interpolate's junction guard (appendGuarded/junctionFormsInjection)
+//     prevents remnants from reassembling any of these patterns across the
+//     boundaries where a value meets template text or another value.
+//
+// The junction guard derives its correctness from list invariants that are
+// pinned by TestInjectionPatternInvariants: every pattern is non-empty ASCII,
+// never starts or ends with a space, and never contains two consecutive
+// spaces.
+var injectionPatterns = []string{
```
Ed-Jaramillo, line 43: Collapsing the three duplicate lists into one canonical declaration is what actually fixes the round-1 blind spot — the drift between them (the sanitizer claiming eleven patterns while the fuzz property counted only backticks) is precisely what let MAJOR-2 hide.

Worth noting this list now has three declared consumers plus a fourth implicit one: `escapeValue`'s `", "` slice separator at line 212 depends on no pattern containing a comma. Good call pinning that in `TestInjectionPatternInvariants` rather than leaving it as a comment — it's the kind of dependency that is easy to break from a distance.

pkg/prompts/interpolation.go
```diff
@@ -22,11 +22,79 @@ import (
...
+}
+
+// maxJunctionSeparators bounds appendGuarded's insertion loop, derived from
+// injectionPatterns rather than hardcoded: once the junction holds a run of
+// spaces longer than the longest run of spaces inside any suppressed pattern,
+// no pattern can straddle it. A straddling occurrence would have to either
+// contain that whole space run in its interior (impossible: no pattern holds a
+// space run that long) or start/end inside it (impossible: no pattern starts
+// or ends with a space — see TestInjectionPatternInvariants).
+var maxJunctionSeparators = func() int {
```
Ed-Jaramillo, line 66: Verified this derivation independently, and it holds. With two spaces at the junction, a straddling occurrence would have to either contain a two-space run in its interior (no pattern does) or begin/end inside the run (no pattern begins or ends with a space). So 2 is sufficient, and the loop exiting on the bound rather than on the predicate is safe rather than merely defensive.

Deriving it from the list instead of hardcoding `2`, then pinning the premises the derivation rests on in a test, converts an invisible proof obligation into something that fails loudly if a future pattern breaks it. That's the part of this change I'd most want kept if the implementation is ever revisited.

pkg/prompts/interpolation.go
```diff
@@ -40,25 +108,95 @@ func Interpolate(template string, vars map[string]interface{}) string {
...
+// junctionFormsInjection reports whether appending right after left would
+// create an occurrence of any suppressed injection pattern that straddles the
+// left/right boundary. For each pattern of length L it inspects a window of
+// the last L-1 bytes of left plus the first L-1 bytes of right: both halves
+// are shorter than L, so any occurrence found in the window necessarily
+// straddles the junction, and any straddling occurrence has at most L-1 bytes
+// on each side, so it always fits in the window. Patterns are pure ASCII, so
+// byte slicing that may split a multi-byte rune at a window edge can neither
+// fabricate nor hide a match.
+func junctionFormsInjection(left, right string) bool {
```
Ed-Jaramillo, line 184: Re-derived the window argument: both halves are shorter than L, so any match found in the window necessarily straddles the boundary; and any straddling occurrence has at most L-1 bytes on each side, so it always fits in the window. No false positives, no misses.

The ASCII note is load-bearing and correct — a multi-byte rune split at a window edge contributes only 0x80-0xBF continuation bytes, which can neither fabricate nor hide a match against a pure-ASCII pattern.

Verified empirically at this SHA with a throwaway brute force: ~1.43M interpolations over adversarial alphabets (fence runes, `#`, `-`, role-marker letters, `:`, `|`, space) across 26 template shapes, with values exhaustive to length 2 on the full alphabet and to length 4 on the narrow one, asserting for every one of the 13 patterns that the result count never exceeds the empty-vars baseline. Zero violations.

pkg/prompts/interpolation.go
```diff
@@ -131,28 +282,15 @@ func escapeString(s string) string {
 	return s
 }
 
-// sanitizePromptInjection detects and removes common prompt injection patterns.
+// sanitizePromptInjection removes every occurrence of the suppressed injection
+// patterns (see injectionPatterns) by replacing it with spaces. Replacement
+// with spaces cannot create a new occurrence: no pattern starts or ends with a
+// space, and the space-containing patterns require a "###" prefix that an
```
Ed-Jaramillo, line 288: NIT, non-blocking — comment only.

This correctly explains why the Alpaca entries can't match here: `###` is removed by an earlier entry and nothing reintroduces `#`. The same is true of `<|im_start|>` and `<|im_end|>`, for a different reason worth stating alongside it — `html.EscapeString` at step 3 has already converted `<` and `>` to entities before step 5 runs, so those two can never match inside a value either.

So four of the thirteen entries are unreachable in this loop. That's harmless: all four are live and load-bearing in `junctionFormsInjection` (your `im_start bridged by value` and `alpaca instruction against template` cases exercise exactly them), and they're a useful safety net if step 3 ever changes. Just worth a clause so a reader doesn't assume this loop is what provides that coverage.

Ed-Jaramillo (approved):
**Approving at `d3bca10d`.** Both round-1 MAJORs are resolved, and in the direction that was asked for rather than by patching the two counterexamples.

The redesign is the right shape: protection moved out of the value (delete bytes) and into the boundary (pad the junction). `escapeString` is now non-destructive, and `Interpolate` guards every junction where a value meets template text or another value.

**MAJOR-1 — resolved.** Every counterexample round-trips byte-for-byte at this head: `~/bin`, `~10`, `` `code` ``, ``use `ls` ``, `literal ~`. A probe also confirmed the escaped value always survives as a *contiguous* substring across padded templates, so the guard pads junctions only and never edits a value's interior.

**MAJOR-2 — resolved and generalized.** `#####` -> `## ##`, `-----` -> `-- --`, and `tem:System:Sys` no longer rebuilds `System:`. More to the point, the mechanism is now generic over `injectionPatterns` instead of fence-specific, and the fuzz property tracks all 13 entries instead of backticks only.

**Independent verification at this SHA** (Windows, `-tags fts5`, CGO on):

- **Caller audit.** `escapeString` / `escapeValue` are unexported and reachable only through `Interpolate`; the only `Interpolate` callers are `cache.go:99`, `cache.go:123`, and `file_registry.go:169`. Nothing bypasses the guard. This was the main structural risk of moving protection out of the escaper, and it is closed.
- **Brute force** (throwaway, since removed): ~1.43M interpolations over adversarial alphabets across 26 template shapes, values exhaustive to length 2 on the full alphabet, length 3 paired, and length 4 single-variable, asserting for every one of the 13 patterns that its count in the result never exceeds the empty-vars baseline. **Zero violations.**
- **Fuzz:** `FuzzPromptInterpolation` 3,400,985 execs and `FuzzEscapeString` 3,716,523 execs (45s each, race-enabled), zero crashers.
- `go test -tags fts5 -race ./pkg/prompts/...` green; `gofmt -l` and `go vet` clean.
- Re-derived the three proofs written into the new code — window sufficiency, the termination bound, and sanitizer non-creation. All three hold; notes inline.
- A no-gratuitous-spaces check confirmed ordinary prompts render byte-identically to before.

**Non-blocking — none of these gate merge:**

1. `docs/guides/integration/prompt-integration.md:161-162` and `prompt-management.md:139` list the escaping steps but not the junction guard's space insertion. Nothing there became false (the old destructive trim was never documented), but the inserted space is observable in rendered output, so it's worth a sentence where prompt authors will look for it.
2. One NIT inline on the sanitizer's doc comment.
3. The branch is 11 commits behind base (`BEHIND`), so the merge button isn't available yet. `git diff 6ee35770..main -- pkg/prompts/` is empty, so updating the branch can't affect any conclusion here — just worth confirming checks stay green afterward.

For the record, the one remaining path by which a suppressed pattern can appear in output, which I checked and consider correct: a value that escapes to the empty string lets two template segments join unguarded. With template `##{{.a}}#`, the values `""`, `###`, ``` ``` ``` and `"  "` all yield `###`, while `x` yields `##x#`. That is exactly the empty-vars baseline for that template, so it grants an attacker nothing the template author had not already written, and the code comment reasons about the case deliberately. Noting it only so the boundary of the guarantee is on the record.

Nice touches: deriving `maxJunctionSeparators` from the list instead of hardcoding `2` and then pinning the derivation's premises in a test; and `TestInjectionPatternInvariants` also pinning the comma invariant that `escapeValue`'s `", "` slice separator quietly depends on.

Full review notes are in my local artifact; happy to paste any section here on request.

🤖 Reviewed with the help of Claude Code (Opus 5)

PR #323, opened 2026-08-17
feat(longmemeval): AKS rig for the full 500-question LongMemEval_S run

ilsiepotamus:
Stacked on #320 (the runner image needs `--occurred-at`/`--dataset` from that branch); retarget to `main` after it merges.

## Summary
`deploy/longmemeval/` — an AKS rig modeled on `deploy/benchmark/`, adapted for a real-LLM, multi-day benchmark run:

- **CGO+fts5 image** (real `looms` + `loom-longmemeval`; the bench Dockerfile's CGO-off/distroless combo can't run the real server)
- **Small, cheap cluster** (D8s workload pool, ~$0.55/hr) — the run is Bedrock-quota-bound, not CPU-bound
- **Resumable chunked runner Job**: one output per `(type, offset)` chunk on an RWX PVC; pod restarts resume by skipping completed chunks (motivated by the pilot losing 4 entries to laptop sleep)
- **Bedrock bearer-key auth** via Secret → `LOOM_LLM_BEDROCK_BEARER_TOKEN`
- **Result safety**: mid-run `pull-results.sh`; teardown refuses until results are confirmed pulled
- README covers time/cost projections from the 2026-08 pilot (~60 min/entry serial in multi-session mode), concurrency guidance, off-cluster official scoring, and what a public claim must disclose

## Validation
- All shell scripts `bash -n` clean; all manifests YAML-parse after `envsubst` rendering; the embedded runner slice-loop parses standalone.
- Not yet exercised against a live cluster — first `az acr build` + `run-500.sh` invocation is the integration test, intentionally left for the operator (spends real money).

### DIFF
--- a/deploy/longmemeval/pull-results.sh
+++ b/deploy/longmemeval/pull-results.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+# pull-results.sh — Copy LongMemEval results from the AKS PVC to local disk.
+# The results PVC is RWX (azurefile), so this works mid-run without
+# disturbing the runner.
+#
+# Usage: bash deploy/longmemeval/pull-results.sh [LOCAL_DIR]
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=lme.env
+source "${SCRIPT_DIR}/lme.env"
+[[ -f "${SCRIPT_DIR}/lme.env.local" ]] && source "${SCRIPT_DIR}/lme.env.local"
+
+TIMESTAMP="$(date -u +%Y%m%d-%H%M%S)"
+LOCAL_DIR="${1:-./results/lme-pulled-${TIMESTAMP}}"
+POD_NAME="lme-results-puller-${TIMESTAMP}"
+
+if ! kubectl get pvc lme-results -n "${LME_NAMESPACE}" &>/dev/null; then
+    echo "ERROR: PVC lme-results not found in namespace ${LME_NAMESPACE}"
+    exit 1
+fi
+
+echo "Creating temporary puller pod..."
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ${POD_NAME}
+  namespace: ${LME_NAMESPACE}
+spec:
+  restartPolicy: Never
+  containers:
+    - name: puller
+      image: alpine:3.21
+      command: ["sleep", "3600"]
+      volumeMounts:
+        - name: results
+          mountPath: /results
+          readOnly: true
+      resources:
+        requests:
+          cpu: "100m"
+          memory: "128Mi"
+        limits:
+          cpu: "500m"
+          memory: "256Mi"
+  volumes:
+    - name: results
+      persistentVolumeClaim:
+        claimName: lme-results
+EOF
+
+kubectl wait --for=condition=Ready "pod/${POD_NAME}" -n "${LME_NAMESPACE}" --timeout=120s
+
+echo ""
+echo "=== Files on PVC ==="
+kubectl exec "${POD_NAME}" -n "${LME_NAMESPACE}" -- sh -c 'ls -lh /results/ | tail -20; echo; echo "chunks: $(ls /results/*.jsonl 2>/dev/null | wc -l)"'
+
+mkdir -p "${LOCAL_DIR}"
+kubectl cp "${LME_NAMESPACE}/${POD_NAME}:/results" "${LOCAL_DIR}"
+kubectl delete pod "${POD_NAME}" -n "${LME_NAMESPACE}" --grace-period=5
+
+echo ""
+echo "Saved to: ${LOCAL_DIR} ($(du -sh "${LOCAL_DIR}" | cut -f1))"
+echo "Score with the official evaluator (see README.md § Scoring)."
--- a/deploy/longmemeval/pull-results.sh
+++ b/deploy/longmemeval/pull-results.sh
@@ -21,6 +21,13 @@ if ! kubectl get pvc lme-results -n "${LME_NAMESPACE}" &>/dev/null; then
     exit 1
 fi
 
+# Under set -e, any failure below (wait, exec, cp) would otherwise skip the
+# final delete and leak the pod — always delete exactly the pod we generated.
+cleanup() {
+    kubectl delete pod "${POD_NAME}" -n "${LME_NAMESPACE}" --ignore-not-found --grace-period=5
+}
+trap cleanup EXIT
+
 echo "Creating temporary puller pod..."
 cat <<EOF | kubectl apply -f -
 apiVersion: v1
@@ -55,11 +62,10 @@ kubectl wait --for=condition=Ready "pod/${POD_NAME}" -n "${LME_NAMESPACE}" --tim
 
 echo ""
 echo "=== Files on PVC ==="
-kubectl exec "${POD_NAME}" -n "${LME_NAMESPACE}" -- sh -c 'ls -lh /results/ | tail -20; echo; echo "chunks: $(ls /results/*.jsonl 2>/dev/null | wc -l)"'
+kubectl exec "${POD_NAME}" -n "${LME_NAMESPACE}" -- sh -c 'ls -lhR /results/ | tail -40; echo; echo "completed chunks (all runs): $(find /results -name "*.done" 2>/dev/null | wc -l)"'
 
 mkdir -p "${LOCAL_DIR}"
 kubectl cp "${LME_NAMESPACE}/${POD_NAME}:/results" "${LOCAL_DIR}"
-kubectl delete pod "${POD_NAME}" -n "${LME_NAMESPACE}" --grace-period=5
 
 echo ""
 echo "Saved to: ${LOCAL_DIR} ($(du -sh "${LOCAL_DIR}" | cut -f1))"
--- a/deploy/longmemeval/pvcs.yaml
+++ b/deploy/longmemeval/pvcs.yaml
@@ -1,0 +1,29 @@
+# lme-results holds benchmark outputs — NEVER auto-delete; these represent
+# days of paid LLM compute. teardown-cluster.sh refuses to run while it exists.
+# azurefile (RWX) so pull-results.sh can mount read-only mid-run.
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: lme-results
+  namespace: loom-lme
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: azurefile-csi
+  resources:
+    requests:
+      storage: 10Gi
+---
+# lme-data caches the HuggingFace dataset download (~285 MB) across job runs.
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: lme-data
+  namespace: loom-lme
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: azurefile-csi
+  resources:
+    requests:
+      storage: 5Gi
--- a/deploy/longmemeval/pvcs.yaml
+++ b/deploy/longmemeval/pvcs.yaml
@@ -1,3 +1,5 @@
+# Template — rendered by run-500.sh (render-common.sh, allowlisted envsubst).
+#
 # lme-results holds benchmark outputs — NEVER auto-delete; these represent
 # days of paid LLM compute. teardown-cluster.sh refuses to run while it exists.
 # azurefile (RWX) so pull-results.sh can mount read-only mid-run.
@@ -5,7 +7,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: lme-results
-  namespace: loom-lme
+  namespace: ${LME_NAMESPACE}
 spec:
   accessModes:
     - ReadWriteMany
@@ -19,7 +21,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: lme-data
-  namespace: loom-lme
+  namespace: ${LME_NAMESPACE}
 spec:
   accessModes:
     - ReadWriteMany
--- a/deploy/longmemeval/render-common.sh
+++ b/deploy/longmemeval/render-common.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# render-common.sh — shared manifest rendering for the LongMemEval rig.
+# Sourced by run-500.sh and render-test.sh (not executed directly).
+#
+# Manifest templates use ${LME_*} placeholders. envsubst runs with an explicit
+# allowlist so shell syntax inside embedded scripts (the runner ConfigMap's
+# slice loop) is never touched, and rendering fails loudly if a required
+# variable is unset or an unknown ${LME_*} placeholder survives.
+
+# Every LME_* placeholder that may appear in a manifest template.
+LME_RENDER_VARS='${LME_NAMESPACE} ${LME_IMAGE} ${LME_IMAGE_TAG} ${LME_GRPC_PORT} ${LME_MODEL} ${LME_DATASET} ${LME_DATASET_FILE} ${LME_MODE} ${LME_CONCURRENCY} ${LME_CHUNK} ${LME_OCCURRED_AT} ${LME_RUN_ID} ${LME_RUN_MANIFEST} ${LME_ALLOW_MANIFEST_DRIFT}'
+
+# lme_render <template> — render a manifest template to stdout.
+# Fails if any allowlisted variable is unset/empty, or if any ${LME_*}
+# placeholder survives rendering (a variable name missing from the allowlist).
+lme_render() {
+    local tpl="$1" rendered var
+    for var in ${LME_RENDER_VARS}; do
+        var="${var#\$\{}"
+        var="${var%\}}"
+        if [[ -z "${!var:-}" ]]; then
+            echo "lme_render: required variable ${var} is unset or empty (rendering ${tpl})" >&2
+            return 1
+        fi
+    done
+    rendered="$(envsubst "${LME_RENDER_VARS}" < "${tpl}")" || return 1
+    if grep -q '\${LME_' <<< "${rendered}"; then
+        echo "lme_render: unrendered LME_* placeholder in ${tpl} (add it to LME_RENDER_VARS):" >&2
+        grep -n '\${LME_' <<< "${rendered}" >&2
+        return 1
+    fi
+    printf '%s\n' "${rendered}"
+}
+
+# lme_sha256_12 <string> — first 12 hex chars of sha256 (portable macOS/Linux).
+lme_sha256_12() {
+    if command -v sha256sum >/dev/null 2>&1; then
+        printf '%s' "$1" | sha256sum | cut -c1-12
+    else
+        printf '%s' "$1" | shasum -a 256 | cut -c1-12
+    fi
+}
--- a/deploy/longmemeval/render-common.sh
+++ b/deploy/longmemeval/render-common.sh
@@ -8,7 +8,7 @@
 # variable is unset or an unknown ${LME_*} placeholder survives.
 
 # Every LME_* placeholder that may appear in a manifest template.
-LME_RENDER_VARS='${LME_NAMESPACE} ${LME_IMAGE} ${LME_IMAGE_TAG} ${LME_GRPC_PORT} ${LME_MODEL} ${LME_DATASET} ${LME_DATASET_FILE} ${LME_MODE} ${LME_CONCURRENCY} ${LME_CHUNK} ${LME_OCCURRED_AT} ${LME_RUN_ID} ${LME_RUN_MANIFEST} ${LME_ALLOW_MANIFEST_DRIFT}'
+LME_RENDER_VARS='${LME_NAMESPACE} ${LME_IMAGE} ${LME_IMAGE_TAG} ${LME_GRPC_PORT} ${LME_MODEL} ${LME_DATASET} ${LME_DATASET_FILE} ${LME_MODE} ${LME_CONCURRENCY} ${LME_CHUNK} ${LME_OCCURRED_AT} ${LME_RUN_ID} ${LME_RUN_MANIFEST} ${LME_ALLOW_MANIFEST_DRIFT} ${LME_MAX_CHUNK_ATTEMPTS}'
 
 # lme_render <template> — render a manifest template to stdout.
 # Fails if any allowlisted variable is unset/empty, or if any ${LME_*}
@@ -32,11 +32,42 @@ lme_render() {
     printf '%s\n' "${rendered}"
 }
 
+# lme_sha256_12_stdin — first 12 hex chars of sha256 over stdin (portable).
+lme_sha256_12_stdin() {
+    if command -v sha256sum >/dev/null 2>&1; then
+        sha256sum | cut -c1-12
+    else
+        shasum -a 256 | cut -c1-12
+    fi
+}
+
 # lme_sha256_12 <string> — first 12 hex chars of sha256 (portable macOS/Linux).
 lme_sha256_12() {
-    if command -v sha256sum >/dev/null 2>&1; then
-        printf '%s' "$1" | sha256sum | cut -c1-12
-    else
-        printf '%s' "$1" | shasum -a 256 | cut -c1-12
+    printf '%s' "$1" | lme_sha256_12_stdin
+}
+
+# lme_build_id <repo-root> — the image tag identifying what will actually be
+# built. `az acr build` uploads the working tree, not the commit, so a commit
+# hash alone does not identify a build: a dirty tree would be tagged with a
+# name a clean build of that commit may already own, and ACR tags are mutable.
+# A clean tree yields the short commit; a dirty one yields
+# <commit>-dirty-<fingerprint>, where the fingerprint covers the uncommitted
+# tracked changes and the contents of every untracked, non-ignored file — the
+# parts of the upload that differ from the commit.
+lme_build_id() {
+    local root="$1" commit fingerprint f
+    commit="$(git -C "${root}" rev-parse --short HEAD)"
+    if [[ -z "$(git -C "${root}" status --porcelain)" ]]; then
+        printf '%s' "${commit}"
+        return 0
     fi
+    fingerprint="$( {
+        git -C "${root}" diff HEAD --binary
+        git -C "${root}" ls-files --others --exclude-standard -z |
+            while IFS= read -r -d '' f; do
+                printf '%s\n' "${f}"
+                cat "${root}/${f}" 2>/dev/null
+            done
+    } | lme_sha256_12_stdin )"
+    printf '%s-dirty-%s' "${commit}" "${fingerprint}"
 }
--- a/deploy/longmemeval/render-common.sh
+++ b/deploy/longmemeval/render-common.sh
@@ -8,7 +8,7 @@
 # variable is unset or an unknown ${LME_*} placeholder survives.
 
 # Every LME_* placeholder that may appear in a manifest template.
-LME_RENDER_VARS='${LME_NAMESPACE} ${LME_IMAGE} ${LME_IMAGE_TAG} ${LME_GRPC_PORT} ${LME_MODEL} ${LME_DATASET} ${LME_DATASET_FILE} ${LME_MODE} ${LME_CONCURRENCY} ${LME_CHUNK} ${LME_OCCURRED_AT} ${LME_RUN_ID} ${LME_RUN_MANIFEST} ${LME_ALLOW_MANIFEST_DRIFT} ${LME_MAX_CHUNK_ATTEMPTS}'
+LME_RENDER_VARS='${LME_NAMESPACE} ${LME_IMAGE} ${LME_IMAGE_TAG} ${LME_GRPC_PORT} ${LME_MODEL} ${LME_DATASET} ${LME_DATASET_FILE} ${LME_MODE} ${LME_CONCURRENCY} ${LME_CHUNK} ${LME_OCCURRED_AT} ${LME_RUN_ID} ${LME_RUN_MANIFEST} ${LME_ALLOW_MANIFEST_DRIFT} ${LME_MAX_CHUNK_ATTEMPTS} ${LME_CONFIG_HASH}'
 
 # lme_render <template> — render a manifest template to stdout.
 # Fails if any allowlisted variable is unset/empty, or if any ${LME_*}
--- a/deploy/longmemeval/render-common.sh
+++ b/deploy/longmemeval/render-common.sh
@@ -18,6 +18,13 @@ lme_render() {
     for var in ${LME_RENDER_VARS}; do
         var="${var#\$\{}"
         var="${var%\}}"
+        # Require only what THIS template references. The allowlist is global,
+        # so demanding every entry made it impossible to render one template
+        # in order to compute a value another template needs — which is
+        # exactly what run-500.sh does for LME_CONFIG_HASH, and it died on the
+        # documented invocation. A variable that never appears here cannot
+        # affect this render.
+        grep -qF "\${${var}}" "${tpl}" || continue
         if [[ -z "${!var:-}" ]]; then
             echo "lme_render: required variable ${var} is unset or empty (rendering ${tpl})" >&2
             return 1
--- a/deploy/longmemeval/render-test.sh
+++ b/deploy/longmemeval/render-test.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+# render-test.sh — Renders every manifest template with NONDEFAULT
+# namespace/image/port/model values and asserts the manifests agree, so a
+# hardcoded value can never desync the workloads when lme.env is overridden.
+# No cluster access needed. Usage:
+#   bash deploy/longmemeval/render-test.sh
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=render-common.sh
+source "${SCRIPT_DIR}/render-common.sh"
+
+# Deliberately nondefault values for every rendered field.
+export LME_NAMESPACE="lme-render-test"
+export LME_IMAGE="testreg.azurecr.io/lme-render-test"
+export LME_IMAGE_TAG="cafef00d"
+export LME_GRPC_PORT="51234"
+export LME_MODEL="test.model-v9"
+export LME_DATASET="small"
+export LME_DATASET_FILE="test_dataset.json"
+export LME_MODE="ingest"
+export LME_CONCURRENCY="3"
+export LME_CHUNK="7"
+export LME_OCCURRED_AT="false"
+export LME_RUN_ID="runid0123abcd"
+export LME_RUN_MANIFEST="dataset=test_dataset.json mode=ingest occurred_at=false model=test.model-v9 image=testreg.azurecr.io/lme-render-test:cafef00d chunk=7"
+export LME_ALLOW_MANIFEST_DRIFT="0"
+
+MANIFESTS=(namespace pvcs server-config runner-script server-deployment runner-job)
+
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "${TMP_DIR}"' EXIT
+
+failures=0
+fail() { echo "FAIL: $*"; failures=$((failures + 1)); }
+
+for m in "${MANIFESTS[@]}"; do
+    lme_render "${SCRIPT_DIR}/${m}.yaml" > "${TMP_DIR}/${m}.yaml" \
+        || fail "${m}.yaml did not render"
+done
+
+# 1. Every rendered manifest parses as YAML.
+if command -v python3 >/dev/null 2>&1 && python3 -c 'import yaml' 2>/dev/null; then
+    for m in "${MANIFESTS[@]}"; do
+        python3 -c 'import sys, yaml; list(yaml.safe_load_all(open(sys.argv[1])))' "${TMP_DIR}/${m}.yaml" \
+            || fail "${m}.yaml does not parse as YAML after rendering"
+    done
+else
+    echo "NOTE: python3+pyyaml not available; skipping YAML parse assertions"
+fi
+
+# 2. The embedded runner script parses as bash.
+if command -v python3 >/dev/null 2>&1 && python3 -c 'import yaml' 2>/dev/null; then
+    python3 -c 'import sys, yaml; sys.stdout.write(yaml.safe_load(open(sys.argv[1]))["data"]["run-slices.sh"])' \
+        "${TMP_DIR}/runner-script.yaml" > "${TMP_DIR}/run-slices.sh" \
+        || fail "could not extract run-slices.sh from rendered runner-script.yaml"
+    bash -n "${TMP_DIR}/run-slices.sh" || fail "rendered run-slices.sh has bash syntax errors"
+fi
+
+# 3. Namespace: the Namespace object and every metadata.namespace agree.
+grep -q "name: ${LME_NAMESPACE}\$" "${TMP_DIR}/namespace.yaml" \
+    || fail "namespace.yaml does not create ${LME_NAMESPACE}"
+for m in pvcs server-config runner-script server-deployment runner-job; do
+    grep -q "namespace: ${LME_NAMESPACE}\$" "${TMP_DIR}/${m}.yaml" \
+        || fail "${m}.yaml missing namespace ${LME_NAMESPACE}"
+    if grep -n '^[[:space:]]*namespace: ' "${TMP_DIR}/${m}.yaml" | grep -v "namespace: ${LME_NAMESPACE}\$"; then
+        fail "${m}.yaml has a namespace other than ${LME_NAMESPACE}"
+    fi
+done
+
+# 4. Images: every image reference is the pinned nondefault tag; never :latest.
+for m in server-deployment runner-job; do
+    grep -q "image: ${LME_IMAGE}:${LME_IMAGE_TAG}\$" "${TMP_DIR}/${m}.yaml" \
+        || fail "${m}.yaml missing pinned image ${LME_IMAGE}:${LME_IMAGE_TAG}"
+    if grep -n '^[[:space:]]*image: ' "${TMP_DIR}/${m}.yaml" | grep -v "image: ${LME_IMAGE}:${LME_IMAGE_TAG}\$"; then
+        fail "${m}.yaml has an image other than ${LME_IMAGE}:${LME_IMAGE_TAG}"
+    fi
+done
+if grep -rn ':latest' "${TMP_DIR}"/*.yaml; then
+    fail "a rendered manifest still references :latest"
+fi
+
+# 5. Port: server config, containerPort, Service port, and the runner's SERVER
+# address must all agree; the old default must not survive anywhere.
+grep -q "port: ${LME_GRPC_PORT}\$" "${TMP_DIR}/server-config.yaml" \
+    || fail "server-config.yaml port != ${LME_GRPC_PORT}"
+grep -q "containerPort: ${LME_GRPC_PORT}\$" "${TMP_DIR}/server-deployment.yaml" \
+    || fail "server-deployment.yaml containerPort != ${LME_GRPC_PORT}"
+grep -q "port: ${LME_GRPC_PORT}\$" "${TMP_DIR}/server-deployment.yaml" \
+    || fail "server-deployment.yaml Service port != ${LME_GRPC_PORT}"
+grep -q "value: \"lme-server:${LME_GRPC_PORT}\"" "${TMP_DIR}/runner-job.yaml" \
+    || fail "runner-job.yaml SERVER != lme-server:${LME_GRPC_PORT}"
+if grep -rn '60051' "${TMP_DIR}"/*.yaml; then
+    fail "a rendered manifest still hardcodes port 60051"
+fi
+
+# 6. Model, dataset, and run identity land where they must.
+grep -q "bedrock_model_id: ${LME_MODEL}\$" "${TMP_DIR}/server-config.yaml" \
+    || fail "server-config.yaml model != ${LME_MODEL}"
+grep -q "value: \"/data/longmemeval/${LME_DATASET_FILE}\"" "${TMP_DIR}/runner-job.yaml" \
+    || fail "runner-job.yaml DATASET_FILE not rendered"
+grep -q "value: \"${LME_RUN_ID}\"" "${TMP_DIR}/runner-job.yaml" \
+    || fail "runner-job.yaml RUN_ID not rendered"
+grep -q "value: \"${LME_RUN_MANIFEST}\"" "${TMP_DIR}/runner-job.yaml" \
+    || fail "runner-job.yaml RUN_MANIFEST not rendered"
+
+# 7. No unrendered placeholders (lme_render also guards; belt and suspenders).
+if grep -rn '\${LME_' "${TMP_DIR}"/*.yaml; then
+    fail "unrendered LME_* placeholder in rendered output"
+fi
+
+if [[ "${failures}" -gt 0 ]]; then
+    echo ""
+    echo "render-test: ${failures} failure(s)"
+    exit 1
+fi
+echo "render-test: OK — all manifests agree on nondefault namespace/image/port/model/run-id"
--- a/deploy/longmemeval/render-test.sh
+++ b/deploy/longmemeval/render-test.sh
@@ -27,6 +27,7 @@ export LME_RUN_ID="runid0123abcd"
 export LME_RUN_MANIFEST="dataset=test_dataset.json mode=ingest occurred_at=false model=test.model-v9 image=testreg.azurecr.io/lme-render-test:cafef00d chunk=7"
 export LME_ALLOW_MANIFEST_DRIFT="0"
 export LME_MAX_CHUNK_ATTEMPTS="4"
+export LME_CONFIG_HASH="cfg123456789"
 
 MANIFESTS=(namespace pvcs server-config runner-script server-deployment runner-job)
 
@@ -47,6 +48,8 @@ if command -v python3 >/dev/null 2>&1 && python3 -c 'import yaml' 2>/dev/null; t
         python3 -c 'import sys, yaml; list(yaml.safe_load_all(open(sys.argv[1])))' "${TMP_DIR}/${m}.yaml" \
             || fail "${m}.yaml does not parse as YAML after rendering"
     done
+elif [[ "${LME_RIG_STRICT:-0}" == "1" ]]; then
+    fail "python3+pyyaml is required to validate rendered manifests (LME_RIG_STRICT=1)"
 else
     echo "NOTE: python3+pyyaml not available; skipping YAML parse assertions"
 fi
@@ -107,6 +110,10 @@ grep -q "value: \"${LME_RUN_MANIFEST}\"" "${TMP_DIR}/runner-job.yaml" \
     || fail "runner-job.yaml RUN_MANIFEST not rendered"
 grep -q "value: \"${LME_MAX_CHUNK_ATTEMPTS}\"" "${TMP_DIR}/runner-job.yaml" \
     || fail "runner-job.yaml MAX_CHUNK_ATTEMPTS not rendered"
+# A config-only change must roll the server pod: looms reads its config once
+# at startup, so without this a new model would never actually be served.
+grep -q "loom.dev/config-checksum: \"${LME_CONFIG_HASH}\"" "${TMP_DIR}/server-deployment.yaml" \
+    || fail "server-deployment.yaml pod template missing the config checksum annotation"
 
 # 7. No unrendered placeholders (lme_render also guards; belt and suspenders).
 if grep -rn '\${LME_' "${TMP_DIR}"/*.yaml; then
--- a/deploy/longmemeval/render-test.sh
+++ b/deploy/longmemeval/render-test.sh
@@ -115,6 +115,22 @@ grep -q "value: \"${LME_MAX_CHUNK_ATTEMPTS}\"" "${TMP_DIR}/runner-job.yaml" \
 grep -q "loom.dev/config-checksum: \"${LME_CONFIG_HASH}\"" "${TMP_DIR}/server-deployment.yaml" \
     || fail "server-deployment.yaml pod template missing the config checksum annotation"
 
+# 6b. run-500.sh computes LME_CONFIG_HASH by rendering server-config.yaml
+# BEFORE that variable exists. lme_render must therefore require only the
+# variables a given template references — requiring the whole global allowlist
+# broke the documented invocation, and every test here pre-seeding the value
+# is precisely why CI stayed green through it.
+(
+    unset LME_CONFIG_HASH
+    lme_render "${SCRIPT_DIR}/server-config.yaml" >/dev/null 2>&1
+) || fail "server-config.yaml must render with LME_CONFIG_HASH unset (run-500.sh's hash step)"
+
+# ...but a variable the template DOES reference must still be required.
+(
+    unset LME_MODEL
+    lme_render "${SCRIPT_DIR}/server-config.yaml" >/dev/null 2>&1
+) && fail "lme_render accepted server-config.yaml with LME_MODEL unset" || true
+
 # 7. No unrendered placeholders (lme_render also guards; belt and suspenders).
 if grep -rn '\${LME_' "${TMP_DIR}"/*.yaml; then
     fail "unrendered LME_* placeholder in rendered output"
--- a/deploy/longmemeval/render-test.sh
+++ b/deploy/longmemeval/render-test.sh
@@ -26,6 +26,7 @@ export LME_OCCURRED_AT="false"
 export LME_RUN_ID="runid0123abcd"
 export LME_RUN_MANIFEST="dataset=test_dataset.json mode=ingest occurred_at=false model=test.model-v9 image=testreg.azurecr.io/lme-render-test:cafef00d chunk=7"
 export LME_ALLOW_MANIFEST_DRIFT="0"
+export LME_MAX_CHUNK_ATTEMPTS="4"
 
 MANIFESTS=(namespace pvcs server-config runner-script server-deployment runner-job)
 
@@ -104,6 +105,8 @@ grep -q "value: \"${LME_RUN_ID}\"" "${TMP_DIR}/runner-job.yaml" \
     || fail "runner-job.yaml RUN_ID not rendered"
 grep -q "value: \"${LME_RUN_MANIFEST}\"" "${TMP_DIR}/runner-job.yaml" \
     || fail "runner-job.yaml RUN_MANIFEST not rendered"
+grep -q "value: \"${LME_MAX_CHUNK_ATTEMPTS}\"" "${TMP_DIR}/runner-job.yaml" \
+    || fail "runner-job.yaml MAX_CHUNK_ATTEMPTS not rendered"
 
 # 7. No unrendered placeholders (lme_render also guards; belt and suspenders).
 if grep -rn '\${LME_' "${TMP_DIR}"/*.yaml; then
--- a/deploy/longmemeval/run-500.sh
+++ b/deploy/longmemeval/run-500.sh
@@ -8,7 +8,7 @@
 #
 # Usage:
 #   LME_BEDROCK_BEARER_TOKEN=ABSK... bash deploy/longmemeval/run-500.sh
-#   bash deploy/longmemeval/run-500.sh --skip-build   # reuse the pushed image
+#   bash deploy/longmemeval/run-500.sh --skip-build   # reuse a pushed image tag
 
 set -euo pipefail
 
@@ -17,6 +17,8 @@ REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
 # shellcheck source=lme.env
 source "${SCRIPT_DIR}/lme.env"
 [[ -f "${SCRIPT_DIR}/lme.env.local" ]] && source "${SCRIPT_DIR}/lme.env.local"
+# shellcheck source=render-common.sh
+source "${SCRIPT_DIR}/render-common.sh"
 
 SKIP_BUILD=0
 [[ "${1:-}" == "--skip-build" ]] && SKIP_BUILD=1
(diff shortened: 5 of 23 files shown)


ilsiepotamus:
Fresh full review as a delta on #320: **no findings**. Deploy scripts and manifests only; verified: no secrets committed (the Bedrock token is env-injected into a k8s Secret at deploy time, docs show only the `ABSK...` placeholder), and — the thing I most wanted to see — `teardown-cluster.sh` refuses to run without `LME_CONFIRM_RESULTS_PULLED=1`, naming the pull-and-verify flow first, so days of paid benchmark compute can't be vaporized by a reflexive teardown; `scale-down.sh` preserves the PVCs. Merge after #320.

deploy/longmemeval/runner-script.yaml
```diff
@@ -0,0 +1,57 @@
...
+    failed=0
+    for i in "${!types[@]}"; do
+      t="${types[$i]}"; n="${counts[$i]}"
+      offset=0
+      while [ "$offset" -lt "$n" ]; do
+        lim="$CHUNK"
+        [ $((offset + lim)) -gt "$n" ] && lim=$((n - offset))
+        out="${RESULTS_DIR}/s500-${t}-o$(printf '%03d' "$offset").jsonl"
+        det="${out%.jsonl}-detailed.json"
+        if [ -s "$out" ]; then
```
Ed-Jaramillo: `-s` only proves that the file contains at least one byte, not that this chunk completed. `WriteJSONL` writes directly to the final path and skips errored entries, so a kill, cancellation, or one failed entry can leave a nonempty partial file that every Job retry will skip. I reproduced this with one-byte invalid JSONL files: the actual slice script skipped all six chunks and exited 0.

Could we write to temporary files, validate the expected IDs/count with no entry errors, atomically rename both outputs, and only then create/check a completion marker?


_Review completed with GitHub Copilot assistance. Findings and recommended changes were verified against the current PR head._

deploy/longmemeval/runner-script.yaml
```diff
@@ -0,0 +1,57 @@
...
+    counts=(133 133 78 70 56 30)
+
+    failed=0
+    for i in "${!types[@]}"; do
+      t="${types[$i]}"; n="${counts[$i]}"
+      offset=0
+      while [ "$offset" -lt "$n" ]; do
+        lim="$CHUNK"
+        [ $((offset + lim)) -gt "$n" ] && lim=$((n - offset))
+        out="${RESULTS_DIR}/s500-${t}-o$(printf '%03d' "$offset").jsonl"
```
Ed-Jaramillo: These names identify only `(type, offset)`, while the PVC survives reruns and `lme.env` can change dataset, mode, `occurred_at`, model/image, and other semantic inputs. A later run therefore silently reuses chunks from the earlier configuration; both workloads also use `:latest` even though the build creates a commit tag.

Could we persist a run manifest/hash, namespace outputs by it, reject mismatched resume attempts, and pin the server and runner to the immutable commit tag or digest?


_Review completed with GitHub Copilot assistance. Findings and recommended changes were verified against the current PR head._

deploy/longmemeval/run-500.sh
```diff
@@ -0,0 +1,75 @@
...
+        --registry "${LME_ACR_NAME}" \
+        --image "loom-longmemeval:latest" \
+        --image "loom-longmemeval:${GIT_COMMIT}" \
+        --file "${SCRIPT_DIR}/Dockerfile" \
+        --build-arg "GIT_COMMIT=${GIT_COMMIT}" \
+        "${REPO_ROOT}"
+fi
+
+echo "=== Applying namespace, PVCs, config ==="
+kubectl apply -f "${SCRIPT_DIR}/namespace.yaml"
```
Ed-Jaramillo: The `LME_*` overrides are only partially applied. Only the runner Job is rendered, and its namespace is still hardcoded; the server/config/PVC manifests hardcode `loom-lme`, the server image, and port `60051`. For example, rendering port `61234` makes the runner connect there while the Service still exposes `60051`, and a custom namespace makes secret/rollout commands target a namespace where the workloads were not deployed.

Please render all configurable fields consistently, including the ACR build target, or remove/reject unsupported overrides and add a nondefault render test.


_Review completed with GitHub Copilot assistance. Findings and recommended changes were verified against the current PR head._

deploy/longmemeval/pull-results.sh
```diff
@@ -0,0 +1,66 @@
...
+TIMESTAMP="$(date -u +%Y%m%d-%H%M%S)"
+LOCAL_DIR="${1:-./results/lme-pulled-${TIMESTAMP}}"
+POD_NAME="lme-results-puller-${TIMESTAMP}"
+
+if ! kubectl get pvc lme-results -n "${LME_NAMESPACE}" &>/dev/null; then
+    echo "ERROR: PVC lme-results not found in namespace ${LME_NAMESPACE}"
+    exit 1
+fi
+
+echo "Creating temporary puller pod..."
```
Ed-Jaramillo, line 31: Because the script uses `set -e`, any failure after pod creation exits before the final delete. Please add an `EXIT` trap that deletes this exact generated puller pod so failed waits/copies and user interrupts do not leave stale pods. This one is non-blocking.


_Review completed with GitHub Copilot assistance. Findings and recommended changes were verified against the current PR head._

Ed-Jaramillo (changes requested):
Thanks for putting together this AKS rig. The deployment structure is clear, the Bedrock credential is kept in a Kubernetes Secret, and the guarded teardown/scale-down paths are good foundations.

I re-reviewed the current head (`efb772d`) and am requesting changes because three issues affect the correctness and reproducibility of the multi-day benchmark:

1. **Chunk completion is not durable or validated.** Any nonempty JSONL is treated as complete, even if it is truncated or omits failed entries. I reproduced the slice loop skipping one-byte invalid files and exiting successfully.
2. **Resume state has no run identity.** Existing chunks are reused across changes to dataset, mode, `occurred_at`, model, or code revision, and both workloads run `:latest` rather than the commit tag already produced by the build.
3. **Configuration overrides are only partially rendered.** Nondefault namespace, image, or port values can make the scripts and manifests target different resources.

I left inline suggestions for the smallest practical fixes, plus one non-blocking cleanup note for the result-puller pod. The default manifests and shell syntax validate locally, the required FTS5 build/test path passes, and the `_S`/oracle category counts and Bedrock model/auth wiring check out. I did not run the funded live AKS/Bedrock integration, so additional cloud-specific issues may still surface after these deterministic problems are addressed.

_Review completed with GitHub Copilot assistance. Findings and recommended changes were verified against the current PR head._

ilsiepotamus:
All three blockers plus the cleanup note addressed in `3d40809a`.

**1. Durable, validated chunk completion**: the runner now writes to `*.tmp` files; a chunk is promoted only after validation — exactly the expected entry count of parseable JSON objects with unique non-empty `question_id` and a string `hypothesis`, plus a detailed-results check that every `error` field is empty (the harness's `EntryResult.Error` convention, since `WriteJSONL` silently skips errored entries). Promotion is an atomic `mv`, then a `.done` marker (also written tmp+mv); resume checks the marker, never `-s`. Your reproduction now behaves: pre-seeded one-byte invalid JSONL → zero chunks skipped, all re-run and promoted with markers; and if the harness itself emits garbage and exits 0, validation fails the chunk, nothing is promoted, and the Job retries. Failed attempts delete only their own unpromoted `.tmp` files — promoted results are never touched.

**2. Run identity**: `run-500.sh` builds a manifest (dataset, mode, occurred_at, model, image:tag, chunk size), hashes it to a run id, and all outputs live under `/results/runs/<run-id>/` with the manifest stored alongside. A resume with a drifted manifest is rejected with exit 1 — never silent reuse, never deletion; deliberate continuation requires explicit `LME_RUN_ID=<id> LME_ALLOW_MANIFEST_DRIFT=1` and is appended to a drift log. `:latest` is gone everywhere: both workloads pin `${LME_IMAGE}:${LME_IMAGE_TAG}` (defaulting to the commit tag the build pushes), and `--skip-build` verifies the tag exists in ACR first. Existing flat-layout results on the live PVC are untouched — the new layout is additive.

**3. Full override rendering**: new `render-common.sh` renders all six manifests with an allowlisted `envsubst` (the embedded runner script's own `${...}` vars survive), with unset-variable and leaked-placeholder guards; namespace, image:tag, port, and model render consistently everywhere, and the ACR build target derives from the same `LME_IMAGE`. The new `render-test.sh` renders with nondefault namespace/image/port/model and asserts every manifest agrees (YAML parses, embedded script passes `bash -n`, no `:latest`, port consistent across server config/containerPort/Service/runner address, no surviving placeholders).

**4. Puller pod**: `trap cleanup EXIT` installed immediately before pod creation, deleting exactly the generated pod with `--ignore-not-found`.

Validation: `bash -n` clean on every touched script including the extracted rendered runner script; all YAMLs parse; render-test passes; a 7-scenario fake-harness simulation (your one-byte repro, retry-after-failure, marker resume with zero harness invocations, SIGTERM-partial, errored entries, manifest-mismatch rejection with outputs preserved, explicit drift acceptance) passed 25/25 checks. `teardown-cluster.sh` and the `LME_CONFIRM_RESULTS_PULLED` guard are byte-for-byte untouched. Like before, I did not run the funded live AKS path.

Ed-Jaramillo (commented):
## Review — LongMemEval AKS rig

Really nice piece of work. The durability model is the standout here: `.tmp` → validate → atomic rename → `.done` marker, with the marker (not file non-emptiness) as the resume signal, is exactly right for a multi-day paid run, and the manifest-based run identity + `render-test.sh` cross-manifest assertions show a lot of care about not silently mixing results or desyncing workloads. The "read before launching" cost section and the teardown safety gate are thoughtful touches too.

I verified the runner's assumptions against the actual `cmd/loom-longmemeval` harness — the detailed-results-is-a-JSON-array check, the `error`/`question_id`/`hypothesis` field names, the filter-by-type-then-offset ordering, the per-type counts, and the presence of `.dockerignore` all line up. So the core logic is sound.

I have three observations, all clustered around the run-identity / resume guarantees the rig advertises. None are blockers — sharing them as things worth weighing before the first real (expensive) launch.

### 1. Image tag is the commit hash, but the build uses the working tree — `run-500.sh`

`GIT_COMMIT="$(git ... rev-parse --short HEAD)"` tags the image with the committed short hash, but `az acr build` uploads the current working tree (dirty or not), and ACR tags are mutable by default. So if an operator has uncommitted local edits (pretty common while iterating on the rig) and launches at HEAD `abc123`, the image is tagged `:abc123` and `LME_RUN_MANIFEST` records `image=...:abc123` — a string identical to a prior *clean* `abc123` run. The manifest check then passes and the new, different-binary run silently **resumes and mixes chunks** from two different builds into one run directory, which is the cross-run contamination the identity design is meant to prevent.

This is the one I'd most want to close before launch since it fails silently on a common condition. A dirty-tree marker on the tag (or hashing the actual build context into the manifest) would restore the guarantee.

### 2. A single deterministically-failing entry blocks its whole chunk — `runner-script.yaml`

Chunk validation is (correctly) all-or-nothing — any entry with a non-empty `error` fails the chunk — but resume granularity is the whole chunk. If one entry errors *deterministically* (e.g. a Bedrock content-filter rejection on that question's text, or an oversized session that fails ingest), `validate_chunk` returns non-zero on every attempt, no `.done` is ever written, and each of the up-to-31 pod restarts (`backoffLimit: 30`) re-runs the entire chunk — re-billing the ~9 healthy entries (~60 min each) up to ~30 times for a chunk that can never complete, while those entries never make it into the results.

Transient failures are handled beautifully by this design; it's specifically the deterministic-poison case that's expensive. Worth a thought on whether a chunk should be able to "give up" on a persistently-failing entry rather than retry the whole batch indefinitely.

### 3. Per-type counts are hardcoded — `runner-script.yaml`

`counts=(133 133 78 70 56 30)` matches `longmemeval_s_cleaned.json` today (I confirmed against `docs/guides/longmemeval-benchmark.md`), so this is a maintainability note rather than a live bug. But the harness already exposes these exact numbers via `loom-longmemeval info`, and CLAUDE.md is explicit about "No hardcodes." If a future dataset revision shifts any count, the loop drifts silently: an over-count drives the final chunk into `offset >= entry count` (a hard error in `main.go:328`, failing all 30 retries), and an under-count stops the loop early and quietly omits entries — producing a wrong published number. Deriving the counts from `info` would keep the loop self-calibrating.

---

Thanks for the thorough PR description and validation notes — they made this easy to review. Happy to talk through any of the above.

ilsiepotamus:
All three observations addressed in `51cee397`. Each one now has a test that fails on the old behaviour rather than an assertion by inspection — I added `deploy/longmemeval/slice-loop-test.sh`, which extracts the runner script from the *rendered* ConfigMap and drives it against a synthetic dataset with a stub harness (no cluster, no Bedrock, no spend).

**1. Dirty tree vs. commit tag** — agreed this was the one that fails silently on a common condition. `lme_build_id()` fingerprints what actually gets uploaded: a clean tree yields the short commit, a dirty one yields `<commit>-dirty-<sha12>` over the uncommitted tracked changes plus the contents of every untracked non-ignored file. Since the tag is in the manifest, a dirty build gets its own run id and can no longer resume a clean build’s chunks; `run-500.sh` prints a warning when it happens. Verified locally that clean → `70eb7f2a`, dirty → `70eb7f2a-dirty-645f3b2a4a67`, and a *different* dirty state → a different fingerprint.

**2. Deterministic poison entry** — chunks now get `LME_MAX_CHUNK_ATTEMPTS` tries (default 3), counted on the PVC so the budget spans pod restarts. Past that the chunk is quarantined with a `.failed` marker naming the failing question ids and how to retry it, and later passes skip it so the remaining chunks finish instead of the run wedging on one bad entry.

It deliberately does not launder a partial run into a passing one: exit stays nonzero and `RUN-INCOMPLETE.txt` lists what is missing, so scoring a run that skipped entries is a loud condition rather than a quiet wrong number. Once only quarantined chunks remain, further restarts are no-op passes that cost nothing and the Job ends `Failed` — the honest outcome. I also stopped deleting the rejected attempt: it moves to `runs/<id>/rejected/`, outside the `s500-*.jsonl` glob the scorer reads, because that output cost real spend and names the poison entry. The test asserts the bound directly — 2 attempts across 4 passes, with the other 4 chunks completing around it.

**3. Hardcoded counts** — removed. `loom-longmemeval info --json` now emits the statistics the text output already computed (`entries`, an ordered `question_types` array, session/turn totals), and the slice loop derives its types and counts from the dataset it is about to run, checking they sum to the entry count. The test uses a dataset whose per-type counts match no hardcoded table (5 + 3, `CHUNK=2`) and asserts the ragged final chunk is sized to what remains and that the loop never runs past the last entry.

One addition in the same spirit as your run-identity point: the derived counts are pinned in `dataset-stats.json` next to `manifest.txt`, so a dataset revised underneath a resume is refused for the same reason a changed configuration is — otherwise deriving counts would just move the drift one level down.

`go test -tags fts5 -race ./cmd/loom-longmemeval`, `render-test.sh`, and `slice-loop-test.sh` all pass; `gofmt` clean. Thanks for the careful second pass — the dirty-tag one in particular would have been very unpleasant to discover after a multi-day run.

Ed-Jaramillo (changes requested):
## PR Review - loom #323

**https://github.com/teradata-labs/loom/pull/323**  
**Title:** feat(longmemeval): AKS rig for the full 500-question LongMemEval_S run  
**Author:** ilsiepotamus  
**Branches:** feat/longmemeval-aks-500 -> main  
**Reviewed SHA:** 51cee39727a9c63dbc5dfe8bc33793d06419dc1d  
**Review started:** 2026-09-10T07:18:29-07:00  
**Review completed:** 2026-09-10T08:27:05-07:00

### Summary

This PR adds an AKS deployment rig for the 500-question LongMemEval_S benchmark, a CGO/FTS5 Docker image containing `looms` and `loom-longmemeval`, machine-readable dataset statistics, and a resumable shell runner. The runner derives per-type chunks from the dataset, validates temporary outputs before promotion, records completion markers on an RWX PVC, bounds repeated chunk failures, and preserves rejected attempts.

The implementation is well covered for the normal first-run path, and the reviewed head plus GitHub's synthetic merge ref both build the LongMemEval package successfully. However, two deterministic failure paths can invalidate a paid run: the runner does not fail closed when PVC persistence commands fail, and `run-500.sh` updates a mounted server ConfigMap without ensuring that the already-running `looms` process reloads it. Additional non-blocking issues affect dataset reproducibility, image identity, transient-failure handling, and regression-test enforcement.

### Evidence Reviewed

**Live GitHub state**

- `gh pr view 323 --repo teradata-labs/loom` and the Pull Requests API for title, author, description, branches, base/head SHAs, commits, mergeability, and update time.
- `gh pr diff 323 --repo teradata-labs/loom` and the Pull Request Files API for the 20 changed paths.
- `gh pr checks 323 --repo teradata-labs/loom`: all listed CI checks completed successfully; `gosec` was neutral/skipping rather than a successful check.
- Reviewed head: `51cee39727a9c63dbc5dfe8bc33793d06419dc1d`. The head did not change during the review.
- The PR API reports base SHA `24e9edc3446cd9777dd3ce829b55769ec6edda87` and merge state `behind`; current `upstream/main` is `2b23b8b87491414bf543b1d5bdb2b9f4d06c3877`. GitHub's synthetic merge commit `480edb8a621c705deb87a42e6ec345c2a4f6545c` was available and was used for targeted merge-result validation.

**Source and integration files inspected**

- `cmd/loom-longmemeval/main.go`
- `cmd/loom-longmemeval/dataset.go`
- `cmd/loom-longmemeval/runner.go`
- `cmd/loom-longmemeval/info_test.go`
- `cmd/loom-longmemeval/runner_test.go`
- `deploy/longmemeval/Dockerfile`
- `deploy/longmemeval/README.md`
- `deploy/longmemeval/lme.env`
- `deploy/longmemeval/namespace.yaml`
- `deploy/longmemeval/pvcs.yaml`
- `deploy/longmemeval/server-config.yaml`
- `deploy/longmemeval/server-deployment.yaml`
- `deploy/longmemeval/runner-job.yaml`
- `deploy/longmemeval/runner-script.yaml`
- `deploy/longmemeval/render-common.sh`
- `deploy/longmemeval/render-test.sh`
- `deploy/longmemeval/run-500.sh`
- `deploy/longmemeval/pull-results.sh`
- `deploy/longmemeval/scale-down.sh`
- `deploy/longmemeval/setup-cluster.sh`
- `deploy/longmemeval/slice-loop-test.sh`
- `deploy/longmemeval/teardown-cluster.sh`
- `.gitignore` and `.dockerignore`
- `cmd/looms/root.go` and `cmd/looms/config.go` for server configuration loading behavior
- `docs/guides/longmemeval-benchmark.md`

**Commands and local verification**

- `CGO_ENABLED=0 go test -tags fts5 -v ./cmd/loom-longmemeval/...` at the reviewed head: passed.
- The same targeted package test at synthetic merge commit `480edb8a621c705deb87a42e6ec345c2a4f6545c`: passed.
- `bash -n deploy/longmemeval/*.sh` at both the reviewed head and synthetic merge: passed.
- `bash deploy/longmemeval/render-test.sh` at both revisions: passed its non-YAML rendering assertions.
- `git diff --check 24e9edc3446cd9777dd3ce829b55769ec6edda87..51cee39727a9c63dbc5dfe8bc33793d06419dc1d`: no whitespace errors.
- `CGO_ENABLED=1 go test -tags fts5 -race ./cmd/loom-longmemeval/...`: could not build because the Windows environment has no `gcc`.
- `bash deploy/longmemeval/slice-loop-test.sh`: exited with `SKIP: jq not available`.

**External documentation inspected**

- Kubernetes Deployment documentation: a rollout is triggered only when the Pod template changes; changing a separate ConfigMap does not itself roll out the Deployment: <https://kubernetes.io/docs/concepts/workloads/controllers/deployment/>
- Kubernetes image documentation: tags can move, non-`latest` tags default to `IfNotPresent`, and digests are immutable: <https://kubernetes.io/docs/concepts/containers/images/>
- Microsoft Azure Container Registry image-tag guidance: unique tags or manifest digests are recommended for deployments, and a Git commit tag is only semi-stable when base images can change: <https://learn.microsoft.com/en-us/azure/container-registry/container-registry-image-tag-version>

**Evidence desired but unavailable**

- No live AKS deployment, Azure Files failure injection, ACR build/pull, Kubernetes rollout, or AWS Bedrock execution was performed. These require cloud infrastructure and, for the benchmark, billable model calls.
- The local environment lacks `jq` and Python/PyYAML, so the rendered YAML parse and extracted real runner state-machine assertions were not executed locally.

### Verification of PR Claims

| # | Claim | Classification | Evidence | Rationale |
|---|---|---|---|---|
| 1 | The reviewed PR head is `51cee39727a9c63dbc5dfe8bc33793d06419dc1d`, and the PR is mergeable but behind `main`. | VERIFIED | Pull Request API and `gh pr view` | Live GitHub state matched the review snapshot; the base branch has advanced since the PR base. |
| 2 | The repository CI checks for the reviewed head pass. | VERIFIED | `gh pr checks 323` | All listed checks were successful except `gosec`, which was neutral/skipping. |
| 3 | The LongMemEval CLI statistics contract and changed package behavior pass targeted tests. | VERIFIED | `go test -tags fts5 -v ./cmd/loom-longmemeval/...` at the head and synthetic merge | The package tests passed in both trees. |
| 4 | All deployment shell scripts are syntactically valid. | VERIFIED | `bash -n deploy/longmemeval/*.sh` | Every deployment script passed Bash syntax validation at the head and synthetic merge. |
| 5 | The rendered manifests and embedded runner pass the new offline tests. | UNKNOWN | `render-test.sh` passed rendering assertions; YAML extraction and `slice-loop-test.sh` were skipped because PyYAML and `jq` were unavailable | The available portions passed, but the claim that all YAML and the extracted state machine were exercised was not independently reproduced. |
| 6 | Chunk outputs are validated, atomically promoted, resumable, and bounded by `MAX_CHUNK_ATTEMPTS`. | SUPPORTED | `runner-script.yaml:44-48, 61-64, 164-249, 260-294`; `slice-loop-test.sh` source | The implementation contains these mechanisms, but the end-to-end state-machine test was skipped locally. |
| 7 | A revised dataset is refused on resume. | CONTRADICTED | `runner-script.yaml:132-142`; `main.go:252-299`; `main.go:113-124, 497-518` | Only `entries` and `question_types` are pinned. A reordered or edited dataset with the same counts/type distribution is accepted, and the downloader caches an existing file without a content hash. |
| 8 | A dirty working tree is separated from a clean build of the same commit. | SUPPORTED | `render-common.sh:49-74`; `run-500.sh:35-62` | The local dirty fingerprint covers tracked diffs and untracked non-ignored files. It does not make the resulting registry tag immutable. |
| 9 | The runner's configured model is the model used by the server. | CONTRADICTED for an existing deployment rerun | `run-500.sh:89-107`; `server-config.yaml:8-21`; `server-deployment.yaml:17-73`; `cmd/looms/root.go:49,131-136` | The script updates a ConfigMap and applies an unchanged Pod template. The server loads configuration during Cobra initialization and no config watcher was found; changing `LME_MODEL` can therefore leave the old process using the old model. |
| 10 | Pilot duration and spend estimates describe observed prior behavior. | AUTHOR-REPORTED | `deploy/longmemeval/README.md:119-134`; PR description | The figures were not independently reproduced and live execution was intentionally not attempted. |

### Merge Recommendation

**REQUEST CHANGES**

The runner can report a successful run after persistence operations fail, and a documented model override can change the run manifest without changing the model used by an already-running server. Both paths can produce materially invalid benchmark output without a corresponding failed Job. The dataset and image identity gaps should also be addressed before relying on the rig for reproducible multi-day results.

### Blocking Findings

#### 1. BLOCKER - Runner ignores PVC persistence failures and can mark incomplete work as successful

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/runner-script.yaml:44, 61-64, 91, 132-142, 221-249, 268-293`
- **Observed behavior:** The embedded script uses `set -uo pipefail`, not `set -e`. `write_atomic` returns the status of `cat`/`mv`, but its calls are not checked. The two result moves and the `.done` marker write are also standalone unchecked commands. The script still prints `chunk complete`, increments `completed`, and exits using `exit "$failed"` even when a persistence command failed; `failed` is set only by harness/validation/quarantine logic.
- **Expected behavior:** A failed write or rename on the results PVC must fail the chunk and the Job, and `.done` must be written only after both promoted result files are known to exist. No run should be reported successful while durable state is uncertain.
- **Evidence:** `write_atomic()` at lines 61-64 has no caller-side status handling. Manifest, dataset-stat, attempt, marker, quarantine, and incomplete-run writes pipe into it without checks at lines 91, 142, 222, 247, 269, and 292. `mv "$tmp_det" "$det"` and `mv "$tmp_out" "$out"` at lines 244-245 are unchecked. The final status is `exit "$failed"` at line 298, and a storage error never sets `failed`.
- **Trigger:** A transient Azure Files rename/write error, a quota/full-volume condition after temporary output creation, or another partial PVC failure during promotion or marker creation.
- **Consequence:** The script can exit zero with missing result files or without a durable completion marker; if one promotion fails while the marker write succeeds, a later pass can skip a chunk whose output is incomplete. The scorer can then omit paid entries or consume a false completion state.
- **Minimal fix:** Make persistence operations fail closed: enable carefully scoped `set -e` or explicitly check every `write_atomic`, `mv`, and cleanup operation that determines state. Promote both files successfully, verify them, and only then write `.done`; return nonzero on any failure.
- **Verification test:** Run the extracted runner against a temporary filesystem seam that fails one result rename and separately fails the marker write. Assert a nonzero exit, no `.done` marker after a failed promotion, and no successful-run summary with missing outputs.

#### 2. BLOCKER - `run-500.sh` does not restart the server when the configured model changes

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/run-500.sh:89-107`; `deploy/longmemeval/server-config.yaml:8-26`; `deploy/longmemeval/server-deployment.yaml:17-73`
- **Observed behavior:** `run-500.sh` applies `server-config.yaml`, then applies a Deployment whose Pod template contains the image, ports, environment, and the fixed ConfigMap name, but not the model value or a ConfigMap checksum. If an existing Deployment keeps the same image/tag and pod-template fields, changing `LME_MODEL` updates only the ConfigMap. `kubectl rollout status` then observes the existing ready Pod rather than proving that a new server loaded the new model. `looms` loads configuration during `cobra.OnInitialize(initConfig)` and no config watcher/reload hook was found.
- **Expected behavior:** Every run must execute against the model and server configuration recorded in its run manifest. A changed model/configuration must cause a fresh server process to load the new ConfigMap, or the script must reject the change while an old process is running.
- **Evidence:** `LME_MODEL` is an operator-configurable input in `lme.env:35` and is rendered into `bedrock_model_id` at `server-config.yaml:11`. The run manifest includes the model at `run-500.sh:74`, so changing it creates a new run identity, but the server rollout path at lines 89-107 does not force a Pod-template change. Kubernetes documents that Deployment rollouts occur only when `.spec.template` changes. The server reads its config at startup in `cmd/looms/root.go:49,131-136`.
- **Trigger:** Complete or interrupt one run, then rerun `run-500.sh` with a different `LME_MODEL` (or a model-specific `lme.env.local`) while keeping the same image tag.
- **Consequence:** The runner records and reports the new model in `manifest.txt` and `RUN_ID`, while the existing `looms` process can continue sending requests to the old model. This silently invalidates the benchmark's model comparison after already incurring cloud spend.
- **Minimal fix:** Tie the Pod template to the rendered server configuration with a checksum/run annotation, or explicitly restart/recreate the server and wait for a new Pod before launching the Job. Define how `/var/lib/loom` state is reset or retained when a new run identity is created.
- **Verification test:** Deploy once with model A, rerun with the same image and model B, and assert that the server Pod UID/ReplicaSet changes and startup logs/config report model B before the runner Job is created. Also verify that a new run does not accidentally reuse state from the previous run.

### Non-Blocking Findings

#### Major

##### 1. Dataset resume protection fingerprints only counts and type distribution

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/runner-script.yaml:132-142`; `cmd/loom-longmemeval/main.go:252-299, 497-518`
- **Observed behavior:** `stats_id` stores only `{entries, question_types}`. It does not include the dataset bytes, question IDs/order, question text, answers, sessions, or dates. `download` skips any existing target file and `downloadFile` writes directly to the final path.
- **Expected behavior:** The run identity and resume check must identify the exact dataset content whose entries produced the completed chunks.
- **Evidence:** A same-size dataset with the same per-type counts but changed question content or ordering produces the same `stats_id`; completed offsets are then reused. The README's broader statement that a revised dataset is refused is therefore true only for revisions that change those summary values.
- **Trigger:** Replace or refresh the cached file with a revision that preserves entry count and type distribution, or retain results while a new download source revision has the same summary statistics.
- **Consequence:** A run can silently combine chunks generated from different questions or answers while appearing to have a matching dataset identity.
- **Minimal fix:** Hash the exact dataset file or a canonical ordered list of question IDs/content and store the hash in the run manifest/statistics. Download to a temporary file and atomically rename it after a successful transfer.
- **Verification test:** Modify question text, answer, ordering, or session content without changing counts/types and assert that resume is refused.

##### 2. The image identity remains a mutable tag, not the built image digest

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/render-common.sh:49-74`; `deploy/longmemeval/run-500.sh:49-75`; `deploy/longmemeval/server-deployment.yaml:30`; `deploy/longmemeval/runner-job.yaml:30,38`
- **Observed behavior:** The script records and deploys `${LME_IMAGE}:${LME_IMAGE_TAG}`. The dirty-tree fingerprint prevents one local dirty tree from sharing a clean commit tag, but `az acr build --image ...:${LME_IMAGE_TAG}` can move the tag and the run manifest records only the tag string. Kubernetes does not specify `imagePullPolicy`, so a non-`latest` tag defaults to `IfNotPresent`.
- **Expected behavior:** A resumable run must identify and launch the exact image bytes used by its completed chunks.
- **Evidence:** Microsoft recommends unique deployment tags or manifest digests and notes that Git commit tags are only semi-stable when base images change. Kubernetes documents that tags can move and digests are immutable; it also documents `IfNotPresent` for non-`latest` tags.
- **Trigger:** Rebuild the same clean commit after a base-image update, retag an ACR image, or use `--skip-build` after the tag has been replaced.
- **Consequence:** A resumed run can use a different binary, or different nodes can use different cached contents for the same tag, while `manifest.txt` and `RUN_ID` still match.
- **Minimal fix:** Resolve the pushed manifest digest and deploy `image@sha256:<digest>` in both the server and runner, recording the digest in the run manifest; alternatively enforce immutable ACR tags and verify the digest before resume.
- **Verification test:** Push two different images under the same tag, launch a resume on a multi-node test cluster, and assert the rendered Deployment/Job use one verified digest.

##### 3. The attempt budget quarantines transient failures as if they were deterministic poison entries

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/runner-script.yaml:221-273`; `deploy/longmemeval/lme.env:53`
- **Observed behavior:** Every nonzero harness exit and every validation failure increments the same `.attempts` counter. After three attempts by default, the chunk is marked `.failed`, skipped on later passes, and the run remains incomplete. No distinction is made between deterministic entry errors and provider/server/PVC failures.
- **Expected behavior:** A temporary provider outage, server restart, or transport failure should be retried with an infrastructure-appropriate policy rather than permanently quarantining healthy entries after the same small budget used for deterministic input failures.
- **Evidence:** The code increments `attempt` before every harness invocation and reaches quarantine solely on `attempt >= MAX_CHUNK_ATTEMPTS`; `lme.env` sets the default to 3. The implementation records the last error but does not classify it.
- **Trigger:** Three consecutive Bedrock throttles, server restarts, or other transient failures while a chunk's entries are otherwise valid.
- **Consequence:** The run becomes incomplete and requires manual removal of `.failed` and `.attempts` even after the infrastructure recovers; healthy neighbors may also be re-billed during the failed attempts.
- **Minimal fix:** Distinguish deterministic per-entry validation failures from infrastructure failures, or use separate retry budgets/backoff policies and do not quarantine on transient transport/provider errors.
- **Verification test:** Make the stub harness fail with a transient transport error for three calls, then succeed; assert that the chunk is retried after recovery rather than permanently quarantined.

#### Minor

##### 4. `RUN-INCOMPLETE.txt` can remain stale after successful recovery

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/runner-script.yaml:278-293`
- **Observed behavior:** The script writes `RUN-INCOMPLETE.txt` when a chunk is quarantined, but there is no success-path removal. If an operator removes the documented `.failed` and `.attempts` markers and a later pass completes the chunk, the old warning remains.
- **Expected behavior:** A run with all chunks completed and no quarantined chunks should not retain an incomplete-run sentinel that the README tells operators to treat as proof of missing entries.
- **Evidence:** The only write is inside `if [ "${#quarantined[@]}" -gt 0 ]`; no `else` removes the file.
- **Trigger:** Recover a quarantined chunk and rerun the Job to completion.
- **Consequence:** Operators or scoring workflows can reject a complete run or report it as incomplete based on stale state.
- **Minimal fix:** Remove `RUN-INCOMPLETE.txt` in the zero-quarantine success path, and add the recovery case to `slice-loop-test.sh`.
- **Verification test:** Quarantine a chunk, clear its markers, complete it, and assert that the sentinel is absent.

##### 5. `lme.env.local` is not ignored by Git

- **Confidence:** HIGH
- **Location:** `.gitignore:38-40`; `deploy/longmemeval/lme.env:1-2`; deployment scripts that source `lme.env.local`
- **Observed behavior:** The operator workflow copies `lme.env` to `lme.env.local`, but `.gitignore` has `.env` and `.env.*` rules that do not match `lme.env.local` under Git's matching behavior. `git check-ignore --no-index deploy/longmemeval/lme.env.local` returns no match. `.dockerignore` separately excludes environment files, so this is a source-control risk rather than a Docker-context leak.
- **Expected behavior:** Local overrides that may contain model, cluster, or bearer-token values should be ignored.
- **Evidence:** The file is sourced by `run-500.sh`, `setup-cluster.sh`, `scale-down.sh`, `teardown-cluster.sh`, and `pull-results.sh`, while the repository ignore rules do not cover its name.
- **Trigger:** An operator creates the documented local override and runs a broad `git add`.
- **Consequence:** Local deployment settings or credentials can be staged accidentally.
- **Minimal fix:** Add `*.env.local` or the narrower `deploy/longmemeval/lme.env.local` pattern to `.gitignore`.
- **Verification test:** `git check-ignore deploy/longmemeval/lme.env.local` should return the matching rule.

##### 6. The new shell tests are not enforced by visible CI and can pass by skipping

- **Confidence:** HIGH
- **Location:** `deploy/longmemeval/slice-loop-test.sh:29-34`; `deploy/longmemeval/render-test.sh:42-61`; repository CI workflow inventory
- **Observed behavior:** `slice-loop-test.sh` exits 0 with `SKIP` when `jq`, Python, PyYAML, or Go is unavailable. `render-test.sh` skips YAML parsing and embedded-script extraction when PyYAML is unavailable. No visible CI check or Just target invokes these two scripts.
- **Expected behavior:** The regression tests for the paid runner should run automatically, or missing prerequisites should fail the test job rather than produce a green no-op.
- **Evidence:** The local slice test exited 0 without running assertions because `jq` was unavailable; the render test passed while explicitly skipping YAML checks. `gh pr checks` listed no LongMemEval shell-test job.
- **Trigger:** CI or an operator runs the scripts in an environment without the optional tools.
- **Consequence:** Changes to manifest YAML or the embedded state machine can merge without the tests that are intended to protect them executing.
- **Minimal fix:** Add a CI job with pinned `jq`, Python/PyYAML, and Go prerequisites, or make the scripts fail when invoked as tests and document an explicit dependency check.
- **Verification test:** Run the scripts in CI and assert that the YAML extraction, Bash parse, and full state-machine assertions execute rather than skip.

### Questions and Assumptions

1. **Is rerunning the rig with a changed `LME_MODEL` or other run configuration supported?**
   - Evidence searched: `lme.env`, `README.md`, `run-500.sh`, the Deployment/ConfigMap templates, and the server config loader.
   - Why it matters: The documentation presents these as operator overrides and the run manifest creates a new identity for them, but the current server Pod is not necessarily restarted.
   - Verdict impact: This does not remove the observed mismatch; if reruns are intentionally unsupported, the scripts should reject them and document the restriction.
   - Resolution: Define and test a restart/state-reset policy for every new run identity.

2. **Can the intended production AKS environment guarantee that `jq`, PyYAML, and the offline shell tests are run in CI?**
   - Evidence searched: visible GitHub check names, deployment scripts, and local execution.
   - Why it matters: The state-machine test currently reports success without exercising its assertions when prerequisites are absent.
   - Verdict impact: It does not change the two blocking code paths, but it affects confidence in future fixes.
   - Resolution: Add the tests to CI with explicit dependencies or change missing dependencies from skip to failure.

3. **Are the PR's ACR tag and cached dataset intended to be externally immutable?**
   - Evidence searched: `run-500.sh`, `render-common.sh`, `runner-script.yaml`, `main.go`, `README.md`, and the ACR/Kubernetes documentation.
   - Why it matters: The code does not verify a registry digest or exact dataset content.
   - Verdict impact: Even if operators currently avoid retagging or replacing cached files, the documented reproducibility guarantee remains weaker than stated.
   - Resolution: Record an image digest and dataset hash, or explicitly state the external immutability requirements and enforce them before resume.

### Suggested PR Comments (posted as inline review comments below; #6 as a separate PR comment)

1. **`deploy/longmemeval/runner-script.yaml:244-248` - fail closed on PVC state errors**

   `write_atomic` and both result `mv` operations are unchecked because the embedded script uses `set -uo pipefail`. If a PVC rename/write fails, the loop can still print `chunk complete`, and `failed` remains zero because it is only set for harness/validation failures. Please check every persistence operation and write `.done` only after both result files are durable; add a failure-injection test that asserts a nonzero Job result and no completion marker.

2. **`deploy/longmemeval/run-500.sh:89-107` - restart the server for changed configuration**

   Applying `server-config.yaml` does not change the Deployment Pod template, so an existing `looms` process can keep the old `LME_MODEL` while the new model is recorded in `RUN_MANIFEST`. Please add a config checksum/run annotation or an explicit restart/recreate and verify the new Pod before launching `lme-runner`; also define how `/var/lib/loom` state is isolated between run identities.

3. **`deploy/longmemeval/runner-script.yaml:132-142` - pin exact dataset content**

   `dataset-stats.json` stores only entry count and type counts. A reordered or edited dataset with the same summary values will pass the resume check and reuse `.done` chunks. Please store a content or canonical-entry hash and test a same-shape content mutation.

4. **`deploy/longmemeval/run-500.sh:49-75` - deploy an immutable image identity**

   The dirty-tree tag fix prevents one local collision, but the manifest still records a mutable ACR tag. Please resolve the pushed digest and use `image@sha256:...` (or enforce immutable tags) in both the server and runner so a resume cannot silently switch binaries.

5. **`deploy/longmemeval/runner-script.yaml:221-273` - separate transient and deterministic retries**

   Every harness/validation failure consumes the same three-attempt budget and then quarantines the chunk. A short Bedrock/server outage can therefore leave a healthy chunk permanently incomplete. Please classify transient infrastructure/provider errors separately or give them a distinct backoff/retry budget.

6. **`.gitignore` - ignore the documented local override**

   Please add `*.env.local` (or the scoped deployment path) so the documented `lme.env.local` override cannot be accidentally staged with local settings or credentials.

### Positive Observations

- The `.tmp` -> validation -> rename -> `.done` lifecycle is a good foundation for resumable paid work when all persistence operations succeed.
- The runner's validation checks exact chunk counts, unique non-empty question IDs, hypothesis types, and detailed-result errors instead of treating a nonempty file as complete.
- The dirty-tree fingerprint is a meaningful improvement over tagging an `az acr build` context only with the committed Git hash.
- The result puller uses a read-only PVC mount and an EXIT cleanup trap, and teardown requires an explicit results-pulled confirmation.
- The targeted Go tests cover the new JSON statistics contract, empty-dataset formatting, historical timestamps, and the time-override abort behavior.

### Review Limitations

- No live AKS, ACR, Azure Files, Kubernetes rollout, or Bedrock run was performed; those require cloud resources and, for the benchmark, billable model calls.
- Race-enabled local tests could not run because the Windows environment lacks a C compiler (`gcc`). The reviewed head's CI unit/build checks were successful.
- `slice-loop-test.sh` exited with `SKIP: jq not available`; `render-test.sh` skipped YAML parsing and embedded-script extraction because Python/PyYAML was unavailable. The available shell syntax and rendering assertions passed.
- `go test -tags fts5 ./cmd/looms/...` on the synthetic merge encountered a Windows file-lock cleanup failure in `TestUpgradeCmd_DryRun`; `cmd/looms` is not changed by this PR, and the targeted LongMemEval package passed on both the PR head and synthetic merge.
- GitHub reports the PR behind current `main`; the synthetic merge ref was available and targeted LongMemEval validation passed there, but a full merged-tree test suite was not completed.
- No source files were modified and no GitHub review, comment, commit, push, or merge action was performed.


---

*This review was performed with GitHub Copilot and Claude Opus 4.7.*

Josh-Schoen:
## PR #323 — feat(longmemeval): AKS rig for the full 500-question LongMemEval_S run

This adds `deploy/longmemeval/` — cluster setup, a resumable chunked runner Job, result-pull/teardown scripts — to actually run the 500-question benchmark on real infrastructure instead of a laptop pilot (which the PR body says lost 4 entries to sleep). Reasonable, scoped to `deploy/` and a couple of `cmd/loom-longmemeval` files; no framework code touched.

**Verdict: approve.** No blockers.

**What this makes possible:** an operator can run the full 500-question LongMemEval_S benchmark unattended on a small, cheap AKS pool (~$0.55/hr, per the PR) with resumability across pod restarts and a safety gate that refuses teardown until results are confirmed pulled.

**Where it could hurt:** the PR body states it's stacked on #320 and "not yet exercised against a live cluster" — both worth knowing before you click merge, but neither is a code-quality problem:
- **The #320 dependency has already resolved.** #320 merged (closed, merged: true) after this PR was opened; the branch merges cleanly against current `main` with no conflicts. The PR body's "retarget to main after it merges" note is just stale — no action needed beyond acknowledging it in the merge.
- **Genuinely unexercised against a live cluster**, as the author says outright — the integration test is the first real `az acr build` + `run-500.sh` run, which spends real money and is explicitly left for the operator. I can't and didn't try to verify cluster behavior; that's honestly out of scope for a static/local review.

**Verified credit:** confirmed #320 is merged via the GitHub API. Confirmed this branch merges into current `main` without conflict. Ran `bash -n` against all 8 shell scripts in `deploy/longmemeval/` — all syntax-clean. Doc-honesty grep against CLAUDE.md's banned-word list on `deploy/longmemeval/README.md` — clean.

**Findings:** none blocking. One "not verified": the Kubernetes YAML manifests (`runner-job.yaml`, `server-deployment.yaml`, etc.) after `envsubst` rendering — I don't have `yq`/`kubectl`/PyYAML available in this environment to parse them, so I couldn't confirm they're valid YAML post-substitution the way the PR's own validation note claims. Worth a manual `envsubst | kubectl apply --dry-run=client -f -` before the first real run if that hasn't already been done.

**Things I was wrong about:** my initial triage called this "not independently reviewable yet" because of the stacked-PR note in the body — that was based on the PR description, which is stale. #320 already merged, so this is reviewable and mergeable as-is.

**Environment gotchas:** none beyond the missing `yq`/`kubectl`/PyYAML locally, noted above as an honest gap rather than a finding about the PR.

Josh-Schoen:
## Correction — I missed an active review before posting

My "approve" above was wrong to post without first checking for existing reviews on this PR — I skipped that step across this whole sweep, and this PR is where it mattered. There's a `CHANGES_REQUESTED` review from @Ed-Jaramillo posted 2 minutes before my comment, on this exact head (`51cee397`), with two HIGH-confidence blockers that supersede my verdict:

1. **Runner can report success on a failed persistence write.** The embedded script uses `set -uo pipefail`, not `set -e`, so `write_atomic` and both result `mv` operations aren't checked — a failed rename/write on the results PVC (e.g. an Azure Files hiccup) can still let the loop print "chunk complete" and write `.done`, because `failed` is only ever set by harness/validation logic, never by a storage error. (`runner-script.yaml:44,61-64,91,132-142,221-249,268-293`)
2. **A model change doesn't restart the server.** `run-500.sh` applies `server-config.yaml` (a ConfigMap) but the Deployment's Pod template doesn't change, so Kubernetes has no reason to roll the Pod — the already-running `looms` process keeps its original `LME_MODEL` even though the run manifest now records the new one. (`run-500.sh:89-107`, `server-config.yaml`, `server-deployment.yaml`)

Both are the kind of thing that fails silently and only shows up after a multi-day paid run finishes with wrong-model or incomplete data — real blockers, not nitpicks.

**Revised verdict: blocked**, pending those two fixes (plus whatever else comes out of a genuine re-review — I have not independently re-verified @Ed-Jaramillo's findings myself, only confirmed the review exists and read its evidence). My original "approve" above should be disregarded.

deploy/longmemeval/runner-script.yaml
```diff
@@ -0,0 +1,298 @@
...
+          reason="harness exited $?"
+          ok=1
+        fi
+
+        if [ "${ok}" -eq 0 ]; then
+          mv "$tmp_det" "$det"
+          mv "$tmp_out" "$out"
+          printf 'run_id=%s entries=%s attempts=%s completed=%s\n' \
+            "${RUN_ID}" "${lim}" "${attempt}" "$(date -u '+%FT%TZ')" | write_atomic "${marker}"
+          rm -f "${attempts_file}"
```
Ed-Jaramillo: **Fail closed on PVC state errors.**

`write_atomic` and both result `mv` operations are unchecked because the embedded script uses `set -uo pipefail`. If a PVC rename/write fails, the loop can still print `chunk complete`, and `failed` remains zero because it is only set for harness/validation failures.

Please check every persistence operation and write `.done` only after both result files are durable; add a failure-injection test that asserts a nonzero Job result and no completion marker.

deploy/longmemeval/run-500.sh
```diff
@@ -0,0 +1,118 @@
...
+        exit 1
+    fi
+    kubectl create secret generic lme-bedrock \
+        -n "${LME_NAMESPACE}" \
+        --from-literal=bearer-token="${LME_BEDROCK_BEARER_TOKEN}"
+fi
+
+echo "=== Deploying server (${LME_IMAGE}:${LME_IMAGE_TAG}, port ${LME_GRPC_PORT}) ==="
+lme_render "${SCRIPT_DIR}/server-deployment.yaml" | kubectl apply -f -
+kubectl rollout status deployment/lme-server -n "${LME_NAMESPACE}" --timeout=300s
```
Ed-Jaramillo, line 131: **Restart the server for changed configuration.**

Applying `server-config.yaml` does not change the Deployment Pod template, so an existing `looms` process can keep the old `LME_MODEL` while the new model is recorded in `RUN_MANIFEST`.

Please add a config checksum/run annotation or an explicit restart/recreate and verify the new Pod before launching `lme-runner`; also define how `/var/lib/loom` state is isolated between run identities.

deploy/longmemeval/runner-script.yaml
```diff
@@ -0,0 +1,298 @@
...
+        else
+          echo "FATAL: refusing to resume run ${RUN_ID}: ${DATASET_FILE} no longer holds the"
+          echo "questions this run's completed chunks answered. Existing outputs are preserved."
+          echo "  stored:  ${stored_stats}"
+          echo "  current: ${stats_id}"
+          exit 1
+        fi
+      fi
+    else
+      printf '%s' "${stats_id}" | write_atomic "${stats_file}"
```
Ed-Jaramillo: **Pin exact dataset content.**

`dataset-stats.json` stores only entry count and type counts. A reordered or edited dataset with the same summary values will pass the resume check and reuse `.done` chunks.

Please store a content or canonical-entry hash and test a same-shape content mutation.

deploy/longmemeval/run-500.sh
```diff
@@ -0,0 +1,118 @@
...
+        exit 1
+    fi
+fi
+
+# Run identity: a manifest of every input that determines what the benchmark
+# measures. Chunk outputs are namespaced by its hash on the results PVC, so a
+# configuration change can never silently reuse chunks from a different run;
+# the runner also verifies the stored manifest before resuming.
+LME_RUN_MANIFEST="dataset=${LME_DATASET_FILE} mode=${LME_MODE} occurred_at=${LME_OCCURRED_AT} model=${LME_MODEL} image=${LME_IMAGE}:${LME_IMAGE_TAG} chunk=${LME_CHUNK}"
+LME_RUN_ID="${LME_RUN_ID:-$(lme_sha256_12 "${LME_RUN_MANIFEST}")}"
```
Ed-Jaramillo, line 92: **Deploy an immutable image identity.**

The dirty-tree tag fix prevents one local collision, but the manifest still records a mutable ACR tag.

Please resolve the pushed digest and use `image@sha256:...` (or enforce immutable tags) in both the server and runner so a resume cannot silently switch binaries.

deploy/longmemeval/runner-script.yaml
```diff
@@ -0,0 +1,298 @@
...
+              printf 'last failure: %s\n' "${reason}"
+              printf 'last attempt output: rejected/%s.jsonl, rejected/%s-detailed.json\n' "${name}" "${name}"
+              printf 'entries the harness reported an error for:\n'
+              errored_entries "${reject_dir}/${name}-detailed.json"
+              printf 'to retry this chunk: rm %s %s\n' "${quarantine}" "${attempts_file}"
+            } | write_atomic "${quarantine}"
+            quarantined+=("${name}")
+            echo "CHUNK QUARANTINED after ${attempt} attempts: ${name} — later passes will skip it so the remaining chunks can finish. See ${quarantine}."
+          else
+            echo "CHUNK FAILED (attempt ${attempt}/${MAX_CHUNK_ATTEMPTS}): ${name} — retried on the next pod restart."
```
Ed-Jaramillo: **Separate transient and deterministic retries.**

Every harness/validation failure consumes the same three-attempt budget and then quarantines the chunk. A short Bedrock/server outage can therefore leave a healthy chunk permanently incomplete.

Please classify transient infrastructure/provider errors separately or give them a distinct backoff/retry budget.

Ed-Jaramillo:
**`.gitignore` — ignore the documented local override**

Please add `*.env.local` (or the scoped path `deploy/longmemeval/lme.env.local`) to `.gitignore` so the documented `lme.env.local` override cannot be accidentally staged with local settings or bearer-token credentials.

_(This comment is scoped to `.gitignore`, which is not in this PR's diff so it can't be posted as an inline review comment.)_

Josh-Schoen (changes requested):
Requesting changes, aligned with @Ed-Jaramillo's review above. My original pass here missed an existing review before posting an approve — corrected in the comment above. The two HIGH-confidence blockers on this head (PVC persistence failures not checked, so a chunk can report success with missing results; a model change doesn't restart the server, so a rerun can silently keep hitting the old model) are real and worth fixing before a funded multi-day run depends on this rig.

ilsiepotamus:
Both blockers and every major/minor fixed in `e9a4c21b`. CI green (18 pass).

**BLOCKER 1 — persistence failures were ignored.** You were right about the mechanism: no `set -e`, and `write_atomic`'s result plus the two promotion renames were unchecked, so a full or flaking Azure Files volume still printed `chunk complete` and could exit zero. Every write and rename is now checked, `write_atomic` cleans up its `.tmp` and returns nonzero, the `.done` marker is written **last** and only after both promoted files are confirmed present, and any persistence failure calls `fatal_pvc`, which stops the pass outright — durable state is uncertain, so writing more results into that volume is the wrong move.

**BLOCKER 2 — a changed model did not restart the server.** Confirmed: `looms` reads config once at `cobra.OnInitialize` with no watcher, so a ConfigMap-only change left the old model serving while the manifest claimed the new one, and `kubectl rollout status` observed the already-ready pod. The rendered config's hash is now stamped on the pod template, so a config-only change rolls a new pod. Verified the hash moves with `LME_MODEL` and is stable for identical config.

**Majors:**
- Run identity pins the dataset's SHA-256, not just `{entries, question_types}` — a revision preserving the summary but changing questions or answers would otherwise resume onto chunks that answered different ones.
- Identity also records the image's immutable digest, and both workloads set `imagePullPolicy: Always`. Unresolvable digests degrade to tag-only with a warning rather than blocking a run.
- The quarantine budget now counts **deterministic** failures only. A harness that could not run at all is infrastructure and says nothing about the chunk's entries; three unlucky Bedrock throttles would otherwise quarantine ~9 healthy ones. Those retries are bounded by the Job's `backoffLimit` instead.

**Minors:** a completed run clears a stale `RUN-INCOMPLETE.txt` (the README tells operators to treat it as proof of missing entries), and `lme.env.local` is gitignored — you were right that the generic `.env` rules never matched it.

**On the tests being a green no-op:** both rig scripts gain `LME_RIG_STRICT`, and `just lme-rig-test` now runs in CI. Nothing invoked them before, and a missing `jq`/pyyaml printed SKIP and went green — for the only regression tests of a runner that spends real money, that is worse than failing. `slice-loop-test.sh` also gains fail-closed and sentinel-cleanup cases; 7 sections, all green.

Both new blockers-fixes are covered by tests that drive the *rendered* runner script, not a copy.

deploy/longmemeval/run-500.sh
```diff
@@ -0,0 +1,142 @@
...
+echo "=== Run identity ==="
+echo "  run id:   ${LME_RUN_ID}"
+echo "  manifest: ${LME_RUN_MANIFEST}"
+echo "  digest:   ${LME_IMAGE_DIGEST}"
+echo "  outputs:  PVC lme-results:/results/runs/${LME_RUN_ID}/"
+
+# Hash the rendered server config so a config-only change (model, port,
+# anything in looms.yaml) rolls the server pod. Computed BEFORE the config is
+# applied, from exactly the bytes that will be applied.
+LME_CONFIG_HASH="$(lme_render "${SCRIPT_DIR}/server-config.yaml" | lme_sha256_12_stdin)"
```
Ed-Jaramillo, line 109: Chicken-and-egg: this is the first `lme_render` call in the script, and it runs while `LME_CONFIG_HASH` is still unset (the export list at :95-99 omits it, `lme.env` doesn't define it), but `lme_render` validates the *full* allowlist — which includes `LME_CONFIG_HASH` — before rendering. Under `set -euo pipefail` the script dies right here: `lme_render: required variable LME_CONFIG_HASH is unset or empty`, before any manifest is applied. Reproduced against this exact head.

One caution on the tempting one-liner (`LME_CONFIG_HASH=bootstrap` just above this line): if `server-config.yaml` ever gains the `${LME_CONFIG_HASH}` placeholder, the dummy value gets baked into the applied config *and* hashed, and neither guard catches it — `bootstrap` isn't `${LME_`. The structural fix belongs in `lme_render` (see comment on render-common.sh).

deploy/longmemeval/render-common.sh
```diff
@@ -0,0 +1,73 @@
...
+# render-common.sh — shared manifest rendering for the LongMemEval rig.
+# Sourced by run-500.sh and render-test.sh (not executed directly).
+#
+# Manifest templates use ${LME_*} placeholders. envsubst runs with an explicit
+# allowlist so shell syntax inside embedded scripts (the runner ConfigMap's
+# slice loop) is never touched, and rendering fails loudly if a required
+# variable is unset or an unknown ${LME_*} placeholder survives.
+
+# Every LME_* placeholder that may appear in a manifest template.
+LME_RENDER_VARS='${LME_NAMESPACE} ${LME_IMAGE} ${LME_IMAGE_TAG} ${LME_GRPC_PORT} ${LME_MODEL} ${LME_DATASET} ${LME_DATASET_FILE} ${LME_MODE} ${LME_CONCURRENCY} ${LME_CHUNK} ${LME_OCCURRED_AT} ${LME_RUN_ID} ${LME_RUN_MANIFEST} ${LME_ALLOW_MANIFEST_DRIFT} ${LME_MAX_CHUNK_ATTEMPTS} ${LME_CONFIG_HASH}'
```
Ed-Jaramillo, line 11: Suggest making `lme_render` require non-empty only for allowlisted variables whose placeholder actually appears in the template being rendered (e.g. `grep -q "\${${var}}" "${tpl}"` before the emptiness check). That keeps both safety properties intact — empty-but-*used* vars are still caught pre-envsubst (envsubst would substitute an empty string and the post-render `${LME_` grep can't see an already-erased placeholder), and unknown placeholders are still caught post-render — while cleanly breaking run-500.sh's bootstrap cycle, since `server-config.yaml` uses only `${LME_GRPC_PORT}`, `${LME_MODEL}`, and `${LME_NAMESPACE}`. The alternative (a dedicated bootstrap-render that skips the hash var) also works but leaves the coupling in place for the next template.

deploy/longmemeval/render-test.sh
```diff
@@ -0,0 +1,128 @@
...
+export LME_DATASET_FILE="test_dataset.json"
+export LME_MODE="ingest"
+export LME_CONCURRENCY="3"
+export LME_CHUNK="7"
+export LME_OCCURRED_AT="false"
+export LME_RUN_ID="runid0123abcd"
+export LME_RUN_MANIFEST="dataset=test_dataset.json mode=ingest occurred_at=false model=test.model-v9 image=testreg.azurecr.io/lme-render-test:cafef00d chunk=7"
+export LME_ALLOW_MANIFEST_DRIFT="0"
+export LME_MAX_CHUNK_ATTEMPTS="4"
+export LME_CONFIG_HASH="cfg123456789"
```
Ed-Jaramillo, line 30: This pre-seed is exactly why CI can't see the run-500.sh bootstrap failure — every rendering test runs with `LME_CONFIG_HASH` already set, but run-500.sh's first render is the one that *computes* it. Please add a case before this line that mimics the bootstrap: with `LME_CONFIG_HASH` unset, assert `lme_render server-config.yaml` succeeds and yields a hash. That's CI-runnable (no kubectl needed) and directly pins this failure mode against regression.

Ed-Jaramillo (changes requested):
Reviewed at e9a4c21, with an in-depth verification pass: the blocker below was independently reproduced from a clean snapshot of this commit (pure-bash repro of `run-500.sh`'s exact state at the failing line).

**Blocker: the documented invocation of `run-500.sh` fails before applying any manifest.** `render-common.sh` puts `LME_CONFIG_HASH` in the global required-variable allowlist and `lme_render` validates every listed variable before rendering anything — but `run-500.sh:109` *computes* that hash by calling `lme_render` on `server-config.yaml` while it is still unset (the export list at :95-99 deliberately omits it). With `set -euo pipefail`, the script dies with `lme_render: required variable LME_CONFIG_HASH is unset or empty (rendering .../server-config.yaml)` at the first render call — before the `kubectl apply` loop at :114.

CI stays green because it never exercises this path: ci.yml:112-113 runs `render-test.sh` and `slice-loop-test.sh`, and both pre-seed `LME_CONFIG_HASH` (render-test.sh:30, slice-loop-test.sh:58). `lme.env` doesn't define it either.

Calibration, to be fair: `LME_CONFIG_HASH=anything bash run-500.sh` works correctly today — :109 reassigns it from the rendered config, and since `server-config.yaml` uses only `${LME_GRPC_PORT}`/`${LME_MODEL}`/`${LME_NAMESPACE}`, the computed hash is independent of the seed. So this is "documented invocation broken, undocumented workaround exists" — still a blocker for a script whose header documents exactly the failing invocation.

Inline comments carry the suggested fix (per-template validation in `lme_render`), why the dummy-seed shortcut is fragile, and the regression test to add. Positive notes: `set -euo pipefail` makes this fail loudly instead of silently hashing empty input into a constant config checksum, and the run-identity/manifest-drift design is genuinely well thought through.

ilsiepotamus:
Rebased cleanly onto current \`main\` at \`156dbe31\` (previous fix commit unchanged: \`e9a4c21b\`, all blockers/majors/minors from the review).

Verified on the pushed head: \`go build -tags fts5 ./...\`, \`go test -tags fts5 -race ./cmd/loom-longmemeval/...\` (all pass), \`gofmt -l\` clean, \`golangci-lint run --build-tags fts5 ./...\` (0 issues), \`buf lint\`/\`buf format --diff\`/\`buf breaking --against main\` all clean, CI green (deploy scripts/yaml unchanged by the rebase).

@Ed-Jaramillo — ready for another look whenever you have a chance.

deploy/longmemeval/server-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: lme-server-config
+  namespace: ${LME_NAMESPACE}
+data:
+  looms.yaml: |
+    llm:
+      provider: bedrock
+      bedrock_model_id: ${LME_MODEL}
+      bedrock_region: us-west-2
+      # bedrock_bearer_token comes from LOOM_LLM_BEDROCK_BEARER_TOKEN (Secret lme-bedrock)
```
Ed-Jaramillo: **Blocker: this env var never reaches `looms`.**

`LoadConfig` calls `viper.AutomaticEnv()` and then `viper.Unmarshal` (`cmd/looms/config.go:1066-1074`). In viper v1.21.0, `Unmarshal` only decodes keys viper already knows (config file, defaults, bound flags, `BindEnv`), and `AutomaticEnv` doesn't register any. `llm.bedrock_bearer_token` is none of those: it only appears in this comment, has no `SetDefault` (`config.go:1154-1173`), and has no flag. The codebase documents the same rule here:

https://github.com/teradata-labs/loom/blob/156dbe31ffadbd45d9e0d8ffda92a7e7ebcacc43/cmd/looms/config.go#L1143-L1144

I reproduced it against the `looms.yaml` that `run-500.sh` renders. `LOOM_LLM_BEDROCK_REGION` overrides as expected, but `LLM.BedrockBearerToken` comes out empty. With no token, the Bedrock client falls back to the default AWS credential chain (`pkg/llm/bedrock/client_sdk.go:101-106`). The pod has no AWS credentials, and startup validation only checks the region, so the server comes up healthy and every call fails.

Declaring the key makes the env var override it (verified):

```suggestion
      # Declared empty so viper binds LOOM_LLM_BEDROCK_BEARER_TOKEN (Secret lme-bedrock) on
      # Unmarshal; an env-only key with no config entry or default is silently dropped.
      bedrock_bearer_token: ""
```

Alternatively, `viper.SetDefault("llm.bedrock_bearer_token", "")` in `setDefaults` fixes it for every `looms` user. Either way, a `LoadConfig` test next to `TestEnvVar_StoragePostgresDSN_NoYAMLKey` would pin it. Please avoid a `${…}` placeholder here: `cmd_serve.go:614` passes the bearer token through without expanding it.

deploy/longmemeval/runner-script.yaml
```diff
@@ -0,0 +1,360 @@
...
+        else
+          failed=1
+          echo "  ${reason}"
+          # Keep the rejected attempt: it cost real Bedrock spend and names
+          # the failing entries. It lands in rejected/ so the scorer's
+          # s500-*.jsonl glob can never pick it up as a result.
+          mkdir -p "${reject_dir}"
+          [ -f "$tmp_out" ] && mv -f "$tmp_out" "${reject_dir}/${name}.jsonl"
+          [ -f "$tmp_det" ] && mv -f "$tmp_det" "${reject_dir}/${name}-detailed.json"
+          if [ "${harness_rc}" -ne 0 ]; then
```
Ed-Jaramillo: **This branch is only reached when the harness exits nonzero, but it exits 0 on provider/transport failures mid-chunk.**

Weave, CreateSession, and CreateAgent errors are recorded per entry (`runner.go:418-546`). `Run` only returns an error for the `occurred_at` abort (`runner.go:171-239`; `runner_test.go:144-151` pins `Unavailable` as not an abort), and `runBenchmark` then exits 0 (`main.go:470-478`). Nonzero exits only happen before any entry runs, e.g. the `ListSessions` health check, which never touches Bedrock.

So a Bedrock throttle or outage (the case `d542fa8` cites), or an `lme-server` restart, is charged at L312-313, and an outage spanning three passes quarantines every incomplete chunk. That's the opposite of `README.md:114-118`. CI can't see it because the stub always exits 0, so nothing exercises this branch.

Could we classify from `-detailed.json` instead? For example, treat the chunk as infrastructure when every errored entry carries a retryable status (`rpc error: code = Unavailable|DeadlineExceeded|ResourceExhausted`). Or the harness could exit with a distinct code on retryable failures. If quarantining these is intended, the README, the comment block above, and `runner-job.yaml:9-13` should say so. A slice-loop case would pin it: the stub marks every entry `code = Unavailable` for `MAX_CHUNK_ATTEMPTS` passes, then succeeds; assert no `.failed` marker.

deploy/longmemeval/runner-job.yaml
```diff
@@ -0,0 +1,95 @@
...
+      initContainers:
+        # One-time (cached on the data PVC): fetch the dataset from HuggingFace.
+        - name: download-dataset
+          image: ${LME_IMAGE}:${LME_IMAGE_TAG}
+          # Any tag other than the floating default pulls IfNotPresent, so a
+          # node with a cached layer under a moved tag would keep serving the
+          # old bytes.
+          imagePullPolicy: Always
+          command: ["/loom-longmemeval"]
+          args: ["download", "--datasets", "${LME_DATASET}", "--data-dir", "/data/longmemeval"]
```
Ed-Jaramillo: Minor: `download` skips any existing file (`main.go:114-118`) and streams straight to the final path without cleanup on error (`main.go:496-521`). A network reset during the first ~285 MB download therefore leaves a truncated dataset cached on `lme-data`, and every later pass dies at `FATAL: could not read dataset statistics` until `backoffLimit`.

There's no spend, but the run can't start until someone deletes the file from a PVC that only the Job mounts. Could the download go to a `.tmp` and be renamed on success, or could the init container validate with `info --json` and remove a bad file before retrying?

deploy/longmemeval/pvcs.yaml
```diff
@@ -0,0 +1,31 @@
+# Template — rendered by run-500.sh (render-common.sh, allowlisted envsubst).
+#
+# lme-results holds benchmark outputs — NEVER auto-delete; these represent
+# days of paid LLM compute. teardown-cluster.sh refuses to run while it exists.
```
Ed-Jaramillo: Nit: `teardown-cluster.sh` doesn't check whether this PVC exists. It refuses until `LME_CONFIRM_RESULTS_PULLED=1`, then deletes the whole resource group, PVC included.

```suggestion
# days of paid LLM compute. teardown-cluster.sh destroys it with the cluster and
# refuses to run until LME_CONFIRM_RESULTS_PULLED=1 (after pull-results.sh).
```

Related: `README.md:84` still describes `dataset-stats.json` as counts only; it pins the dataset SHA-256 now.

Ed-Jaramillo (changes requested):
Re-reviewed at `156dbe3`.

The config-hash bootstrap blocker is fixed. I reproduced the old failure on `d542fa8`, confirmed it's gone here, and confirmed the new `render-test.sh` case fails against the pre-fix `render-common.sh`, so it now covers the path CI couldn't see. A stubbed end-to-end `run-500.sh` also applies all six manifests, with the pod-template checksum matching the hash of the applied config. Thanks for the structural fix.

Requesting changes for two issues that predate this round (I missed both in earlier passes):

1. **Blocker:** the Bedrock key injected as `LOOM_LLM_BEDROCK_BEARER_TOKEN` never reaches `looms`' config, so the documented launch can't authenticate to Bedrock. One-line fix suggested inline.
2. The deterministic-only retry budget still charges provider/transport failures that happen mid-chunk, because the harness exits 0 on them. A Bedrock outage can therefore quarantine healthy chunks, which is the opposite of what the README describes.

There's also a minor issue (an interrupted dataset download stays cached) and a doc nit, both inline.

For calibration: both fail loudly, with no silently wrong numbers and little spend. But the first means the documented launch can't produce a single result, and the second stops the unattended resume the rig exists for.

Not verified: I didn't run anything on a live cluster or against Bedrock. `jq`, PyYAML, and `gcc` weren't available locally, so for the rig suites and `-race` I relied on this SHA's CI logs (both suites ran in strict mode and passed).

PR #324, opened 2026-08-17
fix(bedrock): extract Converse toolUse input via MarshalSmithyDocument (TER-770)

ilsiepotamus:
## Root cause (TER-770)

`ChatConverse` extracted a toolUse block's input with `json.Marshal(b.Value.Input)`. The smithy `document.Interface` concrete types keep their value in an **unexported** field and implement `MarshalSmithyDocument`, not `json.Marshaler` — so `json.Marshal` serializes `{}` and **every tool argument from every Converse-routed model (DeepSeek, Qwen, Meta, …) was silently dropped**.

Downstream effect, reproduced live on preprod pp75 with `deepseek.v3.2` and `qwen.qwen3-235b`:

- any tool with required params rejects the call (`invalid_input`),
- the model re-issues the same call turn after turn until the per-turn cap (the "continuous loading loop" in TER-770),
- no-arg tools (`list_skills`) work, which made the skill system look at fault instead of the transport,
- Anthropic models are unaffected (they never route through `ChatConverse`), so this presented as "skills fail to load on models other than Claude".

Persisted evidence from the repro session: 11 consecutive `manage_skills` calls, each with a Bedrock-assigned `tooluse_…` id and a name but **no input**, each answered `Tool error: invalid_input - action must be 'load' or 'list', got ""`, ending with the model saying "I'm having trouble with the tool format."

## Fix

New `toolUseInputToMap` helper: `MarshalSmithyDocument()` → `json.Unmarshal` into the map (keeps standard `encoding/json` number semantics, matching every other provider path). Nil / non-object inputs degrade to an empty map so the tool's own validation reports the problem.

## Tests

- `TestToolUseInputToMap_ExtractsArguments` — real smithy document round-trips args.
- `TestToolUseInputToMap_PlainJSONMarshalDropsArguments` — pins the failure mode the fix removes (fails loudly if an SDK bump ever changes it).
- nil-document and non-object-input degradation cases.

`go test -race ./pkg/llm/bedrock/` green, `go vet` clean.

### DIFF
--- a/pkg/llm/bedrock/converse.go
+++ b/pkg/llm/bedrock/converse.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/bedrockruntime"
+	"github.com/aws/aws-sdk-go-v2/service/bedrockruntime/document"
 	bedrocktypes "github.com/aws/aws-sdk-go-v2/service/bedrockruntime/types"
 	llmtypes "github.com/teradata-labs/loom/pkg/llm/types"
 	"github.com/teradata-labs/loom/pkg/shuttle"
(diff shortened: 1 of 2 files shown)


bsahitya (approved):
Approving. Verified locally (checked out PR head, `go vet` clean, `go test -tags fts5 -race ./pkg/llm/bedrock/` green).

Root cause is correctly diagnosed: the smithy `document.Interface` concrete type keeps its value in an unexported field and implements `MarshalSmithyDocument` rather than `json.Marshaler`, so `json.Marshal(doc)` serialized `{}` and dropped every Converse-routed tool argument. `toolUseInputToMap` using `MarshalSmithyDocument()` → `json.Unmarshal` is the correct accessor and matches the number semantics of the other provider paths.

The fix degrades safely (nil document and non-object input both fall back to an empty map, leaving the tool's own validation to report missing args). `TestToolUseInputToMap_PlainJSONMarshalDropsArguments` is a good regression pin that will fail loudly if a future SDK bump changes the marshalling behavior.

PR #325, opened 2026-08-17
fix(llm): surface reasoning content on LLMResponse.Thinking (TER-770)

ilsiepotamus:
## What

Stacked on #324 (same file, same response-parsing seam — merge that first; GitHub will retarget this to `main` automatically).

Reasoning models (DeepSeek, Qwen, …) return their trace as `reasoningContent` blocks on Bedrock Converse and as `message.reasoning_content` through OpenAI-compatible gateways (LiteLLM). Both paths silently dropped it, so a reasoning-heavy turn could surface as an **empty response** with no trace of what the model did — the second of the TER-770 secondary findings.

- `ChatConverse`: new `ContentBlockMemberReasoningContent` case → `reasoningContentText` helper extracts `ReasoningText` (redacted reasoning is skipped — encrypted bytes, no readable text) → `LLMResponse.Thinking`.
- OpenAI client: `ChatMessage.ReasoningContent` (`json:"reasoning_content,omitempty"` — response-only; a serialization test pins that outgoing requests stay byte-identical) → `LLMResponse.Thinking` in `convertResponse`.

Deliberately **never** into `Content`: reasoning is not the user-facing answer and must not enter the replayed conversation. A reasoning-only turn still has empty `Content` and the existing empty-answer handling above this layer stays in charge of what the user sees — the trace just becomes observable (`llm.thinking.length` span attribute, persisted Thinking).

Out of scope: `ConverseStream` (known tool-serialization issues, unused by the cloud — `bedrockConverseProvider.ChatStream` falls back to non-streaming `ChatConverse`).

## Tests

- `TestReasoningContentText_{TextBlock,RedactedBlock,Nil}` (bedrock)
- `TestClient_ConvertResponse_{ReasoningContent,ReasoningOnly}` + `TestChatMessage_ReasoningContentOmittedFromRequests` (openai)

`go test -race` green on both packages, `go vet` clean.

### DIFF
--- a/pkg/llm/bedrock/converse.go
+++ b/pkg/llm/bedrock/converse.go
@@ -138,6 +138,7 @@ func (c *Client) ChatConverse(ctx context.Context, messages []llmtypes.Message, 
 
 	// Extract response content
 	var contentText string
+	var thinkingText string
 	var toolCalls []llmtypes.ToolCall
 
 	if output.Output != nil {
@@ -148,6 +149,9 @@ func (c *Client) ChatConverse(ctx context.Context, messages []llmtypes.Message, 
 				switch b := block.(type) {
 				case *bedrocktypes.ContentBlockMemberText:
 					contentText += b.Value
+
+				case *bedrocktypes.ContentBlockMemberReasoningContent:
+					thinkingText += reasoningContentText(b.Value)
 
 				case *bedrocktypes.ContentBlockMemberToolUse:
 					// Extract tool call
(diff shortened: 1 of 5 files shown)


bsahitya (dismissed):
Approving. Verified locally on the PR head (`go vet` clean, `go test -tags fts5 -race ./pkg/llm/bedrock/ ./pkg/llm/openai/` green).

Reasoning is correctly routed to `LLMResponse.Thinking` and never `Content` — confirmed `Thinking` exists on the response type and is consumed only for observability (`llm.thinking.length` span attribute in `instrumented_provider.go`), so a reasoning-only turn leaves `Content` empty and the existing empty-answer handling stays in charge.

Both extraction paths look right: the Bedrock `reasoningContentText` helper skips redacted (encrypted) reasoning, and the OpenAI `reasoning_content` field is response-only with `omitempty` (pinned by `TestChatMessage_ReasoningContentOmittedFromRequests` so outgoing requests are byte-identical). The out-of-scope note on `ConverseStream` is accurate — `ChatStream` delegates to `Chat` → `ChatConverse`, so that streaming path is unused.

Note this is stacked on #324; merge that first as described.

bsahitya (approved):

PR #327, opened 2026-08-17
MCP 2026-07-28 migration: spec, briefs, and full client-side implementation

ilsiepotamus:
## Summary

Implements the client side of the MCP 2026-07-28 migration (stateless core, MRTR, subscriptions, deprecations), plus the migration spec and per-phase implementation briefs it executes.

**Design record:** `docs/architecture/mcp-2026-07-28-migration.md` (verification-corrected rev 2, three diagrams) and `docs/planning/mcp-2026-07-28/` (per-phase briefs with pinned decisions D1–D4, ratified 2026-08-17).

### Protocol / negotiation
- `Connect` negotiates revisions via `server/discover` with initialize fallback; every outbound client path switched (manager, docker manager, teradata backend, fabric factory, `InstrumentedClient`)
- Per-server `protocol_version` pin (`auto`/`legacy`/explicit); fallback widened to bare HTTP 404/405/501; JSON-RPC error bodies in HTTP 4xx delivered as typed protocol errors
- Required request-metadata headers (SEP-2243): `MCP-Protocol-Version`, `Mcp-Method`, `Mcp-Name` (tools/call, prompts/get, resources/read), and `x-mcp-header` → `Mcp-Param-*` mirroring with Base64 sentinel encoding; invalid annotations exclude the tool from `tools/list` per spec

### MRTR (SEP-2322) client driver
- `input_required` results drive a bounded retry loop (default 5 rounds) at the one choke point all request paths share; `requestState` echoed opaquely; retry params rebuilt from originals each round
- Handler-equipped clients advertise the `elicitation` capability; `manager.NewElicitationInputHandler` adapts the HITL approval surface; nil handler = typed fail-fast

### Broken-stream recovery (§7.5 client half)
- Transport synthesizes `CodeStreamLost` when an SSE response stream ends without its final response (fixes a pre-existing hang-until-timeout)
- Stateless `tools/call` carries `_meta[com.teradata.loom/idempotencyKey]`; one re-issue with the same key

### subscriptions/listen client
- `Subscribe` + `subscriptionId` demux, graceful-closure vs stream-loss distinction, manager watch loop (refetch-on-reopen, backoff, clean exit when unsupported)
- Fixes a pre-existing JSON-RPC violation: notifications are dispatched, no longer answered with MethodNotFound

### Deprecations
- §9.1 unreachable code deleted (sampling plumbing, LogNotification, resumption buffer); §9.2 spec-visible surface frozen with `Deprecated:` markers (staticcheck-enforced), removal no earlier than 2027-07-28

## Test plan
- All new behavior covered by table-driven tests against scripted fakes and httptest at the wire level, including the spec's own Base64 encoding examples as goldens
- `go test -tags fts5 -race ./pkg/mcp/...` clean, `-count=5` on client/manager; race detector caught and fixed one real subscription race during development
- `just check` green (matches GitHub CI)

Server side (dual-mode admission, discover handler, looms dedupe, sealed requestState, listen handler, PRM, conformance matrix) follows in subsequent PRs per the briefs.


Ed-Jaramillo (changes requested):
# PR 327 Review: MCP 2026-07-28 Migration

**PR:** https://github.com/teradata-labs/loom/pull/327  
**Reviewed by:** GitHub Copilot and Claude Fabble  
**Recommendation:** REQUEST CHANGES

## Summary

The design direction is sound, but the submitted implementation is not yet interoperable with conforming 2026 servers and regresses some legacy clients. CI and `go test -tags fts5 -race ./pkg/mcp/...` pass, but the tests use self-consistent fake transports rather than official discovery payloads or live long-lived SSE streams.

## Blocking findings

### 1. BLOCKER: Discovery does not match the 2026 wire contract

**Location:** `pkg/mcp/client/connect.go:174-214`, `pkg/mcp/protocol/version.go:150-155`, `docs/architecture/mcp-2026-07-28-migration.md:94,101,287`, `docs/planning/mcp-2026-07-28/phase-1-client-exposure.md:38`

The published [2026-07-28 Discovery specification](https://modelcontextprotocol.io/specification/2026-07-28/server/discover) says the request "carries no body parameters beyond the standard `_meta`" and shows `supportedVersions` in the response, with server identity under `_meta["io.modelcontextprotocol/serverInfo"]`. This code sends `{}`, expects `protocolVersions`, and expects `serverInfo` at the top level. A conforming server either rejects the request or produces an empty version list, preventing every stateless connection.

**Smallest fix:** Stamp the probe with `StampMeta`, model `supportedVersions`, read server identity from result `_meta`, and test against the official discovery example. Correct the architecture diagram, design text, and phase-1 test scenario so later phases do not reintroduce the same wire shape.

### 2. BLOCKER: Streamable HTTP buffers the long-lived subscription stream

**Location:** `pkg/mcp/transport/streamable_http.go:166-210`

`io.ReadAll(resp.Body)` waits for the SSE response to close before starting `handleSSEStream`. Since `subscriptions/listen` intentionally stays open, `Client.Subscribe` never returns or delivers notifications over Streamable HTTP.

**Smallest fix:** Transfer the live response body to the stream goroutine and close it there. Add an `httptest` that keeps the stream open after acknowledgment.

### 3. BLOCKER: Legacy fallback drops required post-initialization headers

**Location:** `pkg/mcp/transport/streamable_http.go:114-121,176-187`, `pkg/mcp/client/client.go:246-266`

Discovery is now the first POST, so `started` becomes true before initialization and the `Mcp-Session-Id` returned by a legacy initialize response is discarded. The following initialized notification also lacks the negotiated `MCP-Protocol-Version`. Strict 2025 session servers will reject that notification.

**Smallest fix:** Capture session IDs from the initialize response regardless of probe order and attach the negotiated protocol version to the initialized notification.

### 4. BLOCKER: Explicit legacy revision pins do not send the pinned revision

**Location:** `pkg/mcp/client/connect.go:96-128`, `pkg/mcp/client/client.go:181-188`

Legacy negotiation and explicit pins discard the selected revision because `Initialize` always sends `protocol.ProtocolVersion` (`2024-11-05`). For example, pinning `2025-03-26` against a server supporting both revisions negotiates 2024 and then fails the pin check.

**Smallest fix:** Add an internal initialize path accepting the requested revision and use it for discovered and pinned legacy versions.

### 5. BLOCKER: The approved deletion decision does not account for downstream source compatibility

**Location:** Deleted API in:

- `pkg/mcp/client/client.go` (base lines 86, 507-510)
- `pkg/mcp/client/instrumented_client.go` (base line 709)
- `pkg/mcp/protocol/types.go` (base lines 205-257)
- `pkg/mcp/transport/resumption.go`

Migration spec §9.1 explicitly directs these symbols to be deleted as unreachable code with "no observable behavior change." That is true for repository-internal runtime paths, but the claim does not cover downstream Go importers. Removing exported `SamplingHandler`, `SetSamplingHandler`, sampling/log types, and `StreamResumption` is source-incompatible for any downstream user of those symbols. Repository-local caller searches cannot establish that exported APIs are unused outside this module.

**Smallest fix:** If exported `pkg/` APIs are covered by v1 compatibility, restore these symbols with `Deprecated:` markers until a major release or the documented removal window. If they are explicitly outside the compatibility guarantee, record that maintainer decision and release policy in §9.1.

### 6. MAJOR: Tool-list reconciliation has a missed-change window

**Location:** `pkg/mcp/manager/watch.go:52-75`

The watcher refreshes tools before opening and acknowledging the subscription. A tool change between those operations is neither in the fetched list nor delivered as a notification. The cache remains stale until another change notification or reconnect; if neither occurs, that can be the remainder of the stream's lifetime.

**Smallest fix:** Open the subscription, wait for acknowledgment, then refetch.

## Non-blocking findings

### 7. MAJOR: Subscription error state is published after channel closure

**Location:** `pkg/mcp/client/subscriptions.go:118-136`

`Notifications` closes before `sub.err` is set. Consumers can observe closure, read `Err() == nil`, and misclassify `MethodNotFound` as graceful termination.

**Smallest fix:** Set terminal error state before closing externally visible channels.

### 8. MAJOR: Stdio subscription cancellation is not sent to the server

**Location:** `pkg/mcp/client/subscriptions.go:158-163`

Context cancellation only removes local state. On stdio, the protocol requires `notifications/cancelled`; otherwise the server retains the subscription.

**Smallest fix:** Send the cancellation notification using the listen request ID for stdio subscriptions.

### 9. MAJOR: Per-server stop and removal leave watchers running

**Location:** `pkg/mcp/manager/manager.go:198-207,283-328`

`StopServer` and `RemoveServer` close the client but do not stop that server's watcher. It continues retrying until the entire manager stops.

**Smallest fix:** Track and cancel watchers per server, then wait for the watcher before closing or removing the client.

### 10. MAJOR: Header-annotation validation changes legacy and stdio behavior

**Location:** `pkg/mcp/client/tools.go:48-75`

Invalid `x-mcp-header` annotations are filtered in every revision and transport. The published [`x-mcp-header` specification](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#x-mcp-header) says Streamable HTTP clients **MUST** reject invalid definitions, while clients using other transports such as stdio **MAY** ignore the annotations entirely. The current unconditional filtering therefore changes behavior outside the 2026 Streamable HTTP path.

**Smallest fix:** Gate validation and filtering on negotiated mode and transport header capability.

### 11. MINOR: Empty MRTR request state is treated as absent

**Location:** `pkg/mcp/protocol/mrtr.go:38-61,78-91`

An explicitly present empty `requestState` is treated as absent, although the schema permits an empty string.

**Smallest fix:** Preserve field presence separately, such as with `*string`.

### 12. MINOR: The implementation brief describes the wrong header behavior

**Location:** `docs/planning/mcp-2026-07-28/phase-1-client-exposure.md:21,45`

The brief says header-marked arguments are removed from the body and emitted as `x-mcp-header-*`. The published [`x-mcp-header` specification](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#x-mcp-header) requires retaining the body value and mirroring it as `Mcp-Param-*`.

**Smallest fix:** Correct the completed brief to match the protocol and implementation.

## Questions and assumptions

- Finding 5 assumes exported APIs under `pkg/` remain source-compatible throughout v1. If maintainers explicitly exclude these packages from that guarantee, its severity becomes a release-policy question rather than a compatibility blocker.
- Is MRTR intentionally library-only in this PR? `NewElicitationInputHandler` has no production call site, so managed clients currently retain fail-fast behavior.
- How is duplicate execution communicated for third-party servers that ignore `com.teradata.loom/idempotencyKey` when the mandatory stream-loss reissue occurs?

## Ready-to-post PR comments

### Discovery wire contract

> **BLOCKER:** The published [2026-07-28 Discovery specification](https://modelcontextprotocol.io/specification/2026-07-28/server/discover) says the request carries standard `_meta`, the response field is `supportedVersions`, and server identity is returned under result `_meta`. This implementation sends `{}` and expects `{protocolVersions, serverInfo}` at the top level, so conforming servers cannot connect. Please align the request/result types, add an official-schema golden test, and correct the same shape in `mcp-2026-07-28-migration.md:94,101,287` and `phase-1-client-exposure.md:38`.

### Long-lived SSE handling

> **BLOCKER:** `io.ReadAll` buffers the SSE body before starting the parser. A `subscriptions/listen` response intentionally remains open, so `Subscribe` blocks forever over Streamable HTTP. Please hand the live body to the stream goroutine and add a persistent-stream integration test.

### Legacy session and version headers

> **BLOCKER:** `started` now refers to the discovery probe rather than initialization, causing the initialize response's session ID to be ignored. The initialized notification also lacks the negotiated version header. Please preserve both headers through legacy fallback.

### Explicit legacy pins

> **BLOCKER:** Explicit legacy pins still send `2024-11-05` in `InitializeParams`. A server supporting both 2024 and the pinned 2025 revision therefore negotiates 2024 and fails the pin. Please initialize with the selected or pinned revision.

### Exported API removals

> **BLOCKER / compatibility-policy confirmation:** I see that migration spec §9.1 explicitly orders these symbols deleted as unreachable code with "no observable behavior change." That is true for repository-internal runtime paths, but downstream Go importers can observe the removal as a compile failure. If exported `pkg/` APIs are covered by v1 compatibility, please restore the sampling, logging, and resumption symbols with deprecation markers until the removal window or a major release. If they are not covered, please record that compatibility decision in §9.1.

### Tool-list reconciliation

> **MAJOR:** Refetching before subscription acknowledgment leaves a race where a tool-list change is missed. The cache then remains stale until another change notification or reconnect, potentially for the stream's remaining lifetime. Please establish and acknowledge the subscription first, then perform the reconciliation fetch.

### Subscription error publication

> **MAJOR:** Subscription cleanup closes the notification channel before publishing the terminal error. A consumer can therefore read `Err() == nil` during abnormal termination. Please finalize error state before channel closure.

### Stdio cancellation

> **MAJOR:** Canceling a stdio subscription only removes client state; the server never receives `notifications/cancelled`. Please propagate cancellation using the listen request ID.

### Per-server watcher lifecycle

> **MAJOR:** `StopServer` and `RemoveServer` do not terminate the associated watcher. Please track watcher cancellation per server and wait for it before closing or removing the client.

### Header annotations on legacy transports

> **MAJOR:** The [`x-mcp-header` specification](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#x-mcp-header) requires rejection for Streamable HTTP clients but says clients using other transports, such as stdio, may ignore the annotation. This unconditional validation hides tools outside the 2026 Streamable HTTP path. Please gate filtering on the negotiated revision and transport capability.

### Empty request state

> **MINOR:** `requestState: ""` is currently indistinguishable from an omitted field, so the client rejects or fails to echo a schema-valid opaque state. Please preserve field presence independently of its value.

### Planning brief correction

> **MINOR:** This brief says header-marked arguments are removed from the body and sent as `x-mcp-header-*`. SEP-2243 retains the body value and mirrors it as `Mcp-Param-*`. Please correct the completed brief so later server work follows the actual contract.

## Positive observations

- Centralizing MRTR and stream-loss handling in `sendRequest` is a good architectural seam.
- Typed protocol and HTTP errors, plus notification dispatch, improve failure visibility.
- The new tests are table-driven and race-clean; the main gap is real wire-level interoperability coverage.

ilsiepotamus:
Thank you for this review — every finding held up under verification against the published specification, the vendored 2026-07-28 schema, and the code. All 12 findings are addressed (blockers, majors, and minors), each in its own commit with regression tests. Point by point:

### 1. Discovery wire contract — fixed in 32866d50
Confirmed against both the published Discovery page and the vendored schema (`DiscoverRequest.params` is required, with required `_meta`; `DiscoverResult.supportedVersions` is required; identity under `_meta["io.modelcontextprotocol/serverInfo"]`). The probe now stamps `params._meta` at the preferred revision — which also satisfies the header/body match rule that would otherwise draw `HeaderMismatch` — and `DiscoverResult` is remodeled on the official schema (`resultType`, `supportedVersions`, `capabilities`, `instructions`, `ttlMs`, `cacheScope`, `_meta`), with `ServerInfo()`/`SetServerInfo()` accessors. A golden test decodes the specification's verbatim response example so the field names cannot drift again, and a probe-shape test asserts the stamped `_meta` keys. The architecture doc (§4.1 diagram, §5.2, §7.1) and the phase-1 brief now describe the same shape, so later server phases implement the real contract.

### 2. Buffered SSE stream — fixed in 25e7d4fc
The live response body is handed to the stream goroutine and parsed incrementally; body ownership transfers with it. A watcher closes the body to unblock the parser's blocking read on transport close or request-context cancellation — the latter is exactly how a 2026-07-28 client cancels an HTTP subscription (closing the stream is the signal), and deliberate teardown no longer synthesizes `CodeStreamLost`. Content-Type is parsed as a media type so `text/event-stream; charset=utf-8` is recognized. httptest coverage: events delivered while the stream stays open (deadlocks on the old code), abrupt-close synthesis, `Close()` unblocking, cancellation without synthesis, parameterized content types.

### 3. Legacy session and version headers — fixed in dad8d75d
Session adoption is no longer gated on "first POST": the transport adopts an `Mcp-Session-Id` whenever a legacy server offers one and none is held (2026 servers never send it), which also re-adopts a fresh session after `ErrSessionExpired`. The initialized notification now carries the negotiated `MCP-Protocol-Version` header. Tests cover the probe-then-initialize ordering at the wire level and the notification header at the client level.

### 4. Explicit legacy pins — fixed in 6873989b
`initializeWithVersion` carries the requested revision: pins request the pinned revision (your 2025-03-26-vs-dual-revision-server scenario is now a test that fails on the old code), discover-selected legacy revisions request exactly what negotiation chose, and the public `Initialize` now requests the newest legacy revision this client speaks (2025-11-25) per legacy negotiation rules — the previous hardcoded 2024-11-05 was also silently under-negotiating every dual-revision legacy server.

### 5. Exported API deletions — resolved in 372772ff (restored + policy recorded)
You're right that repo-local reachability can't cover external importers of a published v1 module. The policy is now recorded in the migration spec as a "compatibility boundary" amendment to §9.1: delete-now applies only to unexported code and internal wiring; **exported** symbols get the §9.2 freeze regardless of internal reachability. Restored accordingly with the staticcheck-enforced `Deprecated:` markers and the 2027-07-28 removal floor: the sampling types + `LogNotification` (protocol), `SamplingHandler`/`SetSamplingHandler` + dispatch (client and instrumented client — functional for legacy connections, exactly as before), and `StreamResumption` (transport, retained unwired since its former write path was observable to nobody). The immediate brief carries an amendment note so the deletion instructions aren't re-executed as written.

Restoring the dispatch exposed a latent routing bug that predates this PR: `receiveLoop` probed for a JSON-RPC *response* before a *request*, so any id-bearing server-initiated request parsed as a response and was silently dropped — servers never even received the `MethodNotFound` answer they're owed. Routing now follows JSON-RPC shape (a `method` member means request/notification; only method-less messages are responses), with dispatch tests for both the handler and no-handler paths.

### 6. Tool-list reconciliation window — fixed in f8e52aa1
The watcher now subscribes first and runs the reconciliation fetch on `notifications/subscriptions/acknowledged` — the race-free point, since the server sends nothing on the subscription before the ack and notifies every change after it. A failed reconciliation tears the subscription down and re-establishes with backoff rather than trusting a stale cache. The test asserts the ordering (`listen` strictly before any `tools/list`).

### 7. Error published after channel close — fixed in f8e52aa1
The terminal error is set before anything externally visible closes; a consumer that drains `Notifications` to closure and immediately reads `Err()` now always sees the abnormal cause (regression test does exactly that with a `MethodNotFound` ending).

### 8. Stdio cancellation — fixed in f8e52aa1
On transports where closing the response stream is not the cancellation signal, cancelling a subscription now sends `notifications/cancelled` referencing the listen request's ID (with a fresh bounded context, since the subscription's own context is already cancelled). On Streamable HTTP nothing is sent — per the subscriptions spec, closing the SSE stream is the signal there, and with finding 2 fixed the request context's cancellation actually closes it. The new `transport.RequestHeaderCarrier` optional interface distinguishes the transports; tests cover both (cancelled sent with the right `requestId` on stdio, nothing sent on Streamable HTTP).

### 9. Watcher lifecycle — fixed in b5fe2421
Per-server `watcherHandle` (stop + done): `StopServer`, `RemoveServer`, and `Stop` close the handle and wait for the watcher to exit before closing the client (watchers never take the manager mutex, so the wait can't deadlock), and remove the registration — which also fixes a duplicate-watcher bug on stop→re-add of the same name. Lifecycle tests for all three paths.

### 10. Header-annotation scope — fixed in f6a94d19
Validation, exclusion, and `Mcp-Param-*` mirroring now apply only when the connection is stateless **and** the transport carries request headers (`RequestHeaderCarrier`); stdio and legacy-negotiated connections ignore the annotations entirely and never hide tools. Tests cover the MUST-reject path on stateless Streamable HTTP, the ignore path on stdio, the ignore path on legacy-over-Streamable-HTTP, and no-mirroring off Streamable HTTP.

### 11. Empty requestState — fixed in 1f4ea296
`RequestState` is now `*string`: present-but-empty parses (it satisfies "at least one of `inputRequests` or `requestState`"), is echoed back verbatim on the retry as the spec's MUST requires, and absent stays absent. Tests cover parse, echo, and omission, plus an end-to-end client retry with `requestState: ""`.

### 12. Phase-1 brief header behavior — fixed in 8c45c6a2
The brief now states the real SEP-2243 contract (value stays in the body, mirrored as `Mcp-Param-{name}` with the Base64 sentinel encoding, server-validated for agreement), matching the implementation, so later server work follows the actual contract.

### Your questions

- **MRTR library-only?** Intentionally, in this PR. `Config.MRTR` is programmatic (`yaml:"-"`) and `NewElicitationInputHandler` is the adapter; the production wiring from looms' HITL approval surface is follow-up work in the server-side phases. Fail-fast (`InputRequiredNotSupportedError`) is the deliberate default for headless contexts that cannot answer elicitations — silently guessing answers would be worse.
- **Idempotency key on third-party servers.** The re-issue after a lost stream is at-least-once by nature — the 2026 revision removed resumption, so re-issuing as a new request is the only recovery the protocol offers, dedupe or no dedupe. The vendor key upgrades that to effectively-once on servers that honor it (Loom's own dedupe lands in the server-side phases); on servers that ignore it, the risk is exactly what the revision's model already implies. The client makes the event visible by logging the re-issue at Warn with the idempotency key. Only one re-issue is ever attempted, and only in stateless mode.
- **Compatibility policy** — answered under finding 5: exported `pkg/` API is treated as v1-compatible surface; the decision and its rationale are recorded in §9.1.

### Two adjacent interop gaps fixed while in there

- **Bodiless 202/204 notification acknowledgments** (c29aee1e): a conforming server answers an accepted notification POST with 202 and no body — and therefore no Content-Type — which the Content-Type switch rejected as unexpected.
- **400 fallback for strict legacy session servers** (d10a82fa): the transport spec's backward-compatibility rule says a 400 whose body is not a recognized modern JSON-RPC error identifies a legacy server. Strict 2025 session servers (the official TypeScript SDK's among them) reject any pre-initialize POST with a bare 400 ("session required"), so the auto-negotiating probe previously hard-failed against them instead of falling back to the handshake. Modern 400s (UnsupportedProtocolVersion/HeaderMismatch bodies) still never fall back.

### Verification
`just check` green (lint + full test suite with `-tags fts5 -race` + build); `-count=5 -race` clean on `pkg/mcp/client` and `pkg/mcp/manager`; golden tests pin the discover response and header-encoding examples from the published specification. One honest caveat on your interoperability point: these fixes are still validated against scripted fakes and httptest at the wire level, now shaped by the official examples; the official-SDK interop suite lives in the server-side branch that stacks on this one and is being re-based onto these corrections (its server-side discover handler emits the old shape and is being fixed to the real contract there).

Ed-Jaramillo (changes requested):
## Follow-up review

> Review performed with GitHub Copilot, powered by **GPT-5.6 Sol** (model ID: `gpt-5.6-sol`).

### Summary

All 12 original findings are addressed at PR #327 head `d10a82f`. However, I found one new compatibility blocker.

### Merge recommendation

**REQUEST CHANGES**

### BLOCKER — legacy stdio fallback remains too narrow

**Location:** `pkg/mcp/client/connect.go:83-91,161-177`; `pkg/mcp/client/client.go:114,144-160,509-548`

`isLegacyServerSignal` falls back only for `MethodNotFound`. The [MCP stdio specification](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/stdio#backward-compatibility) explicitly requires fallback for any non-modern error or timeout, commonly including `-32602 InvalidParams`. I reproduced a legacy server returning `-32602 "initialize first"`; `Connect` failed without sending `initialize`. Silent legacy servers also cannot recover because `RequestTimeout` is defaulted but never applied, leaving the probe waiting until the caller context is exhausted.

This is merge-blocking under the PR's stated dual-era compatibility goal: it causes connection-level failure for a class of legacy stdio servers the specification explicitly requires clients to handle. It is less severe than a security or data-loss blocker, but it is still a concrete compatibility regression in the primary negotiation path. If Loom intentionally supports only legacy servers that return `-32601`, this can instead be treated as a documented limitation and downgraded to MAJOR.

**Smallest fix:** Apply transport-specific fallback rules, treat non-modern stdio errors as legacy, and run discovery with a bounded child timeout that leaves the parent context available for initialization. Add `-32602` and no-response regressions.

### Prior findings

Discovery shape, live SSE streaming, legacy headers, revision pins, exported APIs, watcher reconciliation, error ordering, stdio cancellation, watcher lifecycle, annotation scope, empty `requestState`, and documentation are all fixed with regression coverage. Focused FTS5/race suites pass.

### Concise review comment

> **BLOCKER:** `isLegacyServerSignal` only recognizes `MethodNotFound`, but the published stdio compatibility rule says fallback must not depend on one error code: legacy servers commonly return `-32601`, `-32602`, or remain silent. A server returning `-32602 "initialize first"` currently fails `Connect` without attempting `initialize`. Please classify every non-modern stdio response as legacy and give discovery a bounded timeout that preserves context for the fallback handshake.

ilsiepotamus:
Thanks for the follow-up — the new blocker is confirmed and fixed in a97b6ee3. You're right on all three counts: the stdio page's rule is a hard MUST NOT ("the fallback MUST NOT be keyed to one specific error code: legacy servers respond to unknown pre-`initialize` requests with implementation-defined errors (commonly `-32601` or `-32602`) or not at all"), and `Config.RequestTimeout` was indeed defaulted in `NewClient` and then never read — dead config.

What changed, exactly as the binding pages prescribe:

- **Error classification**: any JSON-RPC error from the probe other than the recognized modern codes (`-32022 UnsupportedProtocolVersion`, `-32020 HeaderMismatch`, `-32021 MissingRequiredClientCapability`) is now a legacy signal on every transport; the modern codes never fall back (with no mutual modern version that's a hard error naming the server's advertised list, never a silent downgrade). Your reproduced `-32602 "initialize first"` case is a regression test, alongside `-32000` and `-32603`.
- **Bounded probe**: the discover probe now runs under a child context bounded by `Config.RequestTimeout` (wired for real; default 30s), so the caller's context is still alive for the fallback handshake when a legacy server never answers. Probe silence is classified as legacy **on stdio only**, per the transport-specific rules — on Streamable HTTP the era signal is status-based, so a hung POST stays a hard error rather than masking an outage; and caller-context expiry is never a legacy signal (regressions for all three).
- **Scope of RequestTimeout**: deliberately bounds only the negotiation probe, and its doc comment now says so — tool calls and MRTR elicitation rounds legitimately outlive any fixed timeout and remain caller-context-bounded.

The architecture doc (§4.1) and the phase-1 brief now state the complete fallback rule set (non-modern JSON-RPC errors; bare 404/405/501; 400 without a modern body; stdio probe silence) instead of the original narrow trigger, so the rule can't regress from the docs either.

`just check` green on both branches, full `-race` suite clean, and `feat/mcp-2026-server` (#328) is rebased on top with the SDK interop suite passing.

Ed-Jaramillo:
**Final-pass blocker: late probe responses can panic the client.**

The latest commit fixes the stdio fallback issue, but the new bounded probe timeout exposes a close/send race in `pkg/mcp/client/client.go:528-534,650-661`.

When discovery times out, `dispatchAndWait` removes the pending entry and closes `respChan`. Meanwhile, `handleResponse` may already have copied that channel under `pendingMu.RLock`; because it releases the lock before sending, a response arriving at the deadline can execute `respChan <- resp` after cleanup closes the channel. This races under `go test -race` and can panic with `send on closed channel`.

The channel does not need to be closed after its map entry is removed, so removing `close(respChan)` is the smallest fix. Alternatively, synchronize lookup/send with cleanup/close. Please add a regression that delivers a response while cancellation removes the pending entry.

All previous review findings are otherwise resolved.

ilsiepotamus:
Recording a review finding relayed out-of-band (the reviewer's final pass caught it via stress testing but GitHub permissions prevented posting it here): **a timing race between request teardown and a late server response could panic the client.**

The sequence: a request's context expires and `dispatchAndWait`'s teardown deleted its pending entry and closed the response channel; a late server response that had grabbed the channel reference under the read lock just before the delete then sent into it — send on a closed channel panics, and a `select` `default` clause does not protect a send on a *closed* channel. The window is a few instructions wide, which is why CI never tripped it.

Fixed in 228adc7c:

- **The reported race**: the response channel is deliberately never closed. It's buffered, its only receiver has already returned when teardown runs, and an orphaned unclosed channel is simply garbage-collected — the close was pure liability. A 2000-iteration stress loop that reliably reproduced the panic on the old code (within a few hundred iterations under `-race`) now guards it as a regression test.
- **Same-class audit** across the package found two more closes reachable by unwaited senders, fixed the same way: `StreamableHTTPTransport.Close` closed the message channel that `Send` still delivers into on the caller's goroutine (in-flight request racing `Close` → same panic; now guarded by a concurrent Send-vs-Close stress test), and the frozen legacy `HTTPTransport.Close` closed the events channel the SSE callback can still write. The correct single-owner closes (stdio reader, subscription notifCh under its mutex, the notifications channel behind the waited receive loop) are untouched and documented.

`just check` green, full `-race` suite clean, `feat/mcp-2026-server` (#328) rebased on top with the SDK interop suite passing.

Ed-Jaramillo (approved):
Thank you, Ilsun — I completed the final pass on `228adc7`. The reported late-response race is fixed, and the same-class transport audit looks correct. I reran the affected client and transport packages under the race detector and stress-ran both new regressions 25 times; everything passed. No remaining findings from me. Approved.

PR #328, opened 2026-08-18
MCP 2026-07-28 migration: server side — dual-mode core, ownership, dedupe, subscriptions, MRTR, PRM

ilsiepotamus:
## Summary

Server side of the MCP 2026-07-28 migration, stacked on #327 (client side). Retarget to `main` once #327 merges.

### Phase 2 — dual-mode core (`f675f346`)
- `server/discover` (advertising only implemented revisions), `_meta` middleware (`RequestMetaFromContext`), result stamping (never overwrites handler `resultType`), stateless `ping` → MethodNotFound
- Transport dual-mode admission: stateless requests bypass the session layer entirely; SEP-2243 header-body validation rejects disagreement with `HeaderMismatch` (-32020) in an HTTP 400 routable by request id; legacy admission byte-for-byte unchanged

### Session ownership (`0513cde8`) — security
- The transport session never gated admission (header-less POSTs reach any handler), so ownership is enforced at the state owner: SQLite sessions gain `user_id` (via `SessionStore.initSchema`, which owns that schema — see the brief amendment for why not a numbered migration), all five store methods predicate on ctx identity, and the upsert is ownership-guarded against cross-user row overwrite
- gRPC layer closes the unenforced paths (`Weave`/`StreamWeave` routing, `GetConversationHistory`, `SubscribeToSession`) behind one documented predicate; wrong-owner is byte-indistinguishable from not-found (asserted); identity plumbing moved to `pkg/types` (postgres delegates)
- First-ever test coverage for these checks, `-race` clean

### D1 — weave idempotency dedupe (`8b6e6dcc`)
- Bridge forwards `_meta[com.teradata.loom/idempotencyKey]` as gRPC metadata; looms dedupes `Weave`/`StreamWeave` per (caller, key) with terminal-result-only joins and a 10-minute replay window — a spec-mandated re-issue after stream loss joins the in-flight run instead of double-executing

### Phase 5 — subscriptions/listen (`13671f80`)
- Listen streams with ack-first semantics and filter honoring; `publishNotification` gives server-initiated notifications their first HTTP delivery path (fixes the notifyCh fill-and-warn bug)
- **Fixes a latent transport bug:** the client buffered whole SSE bodies, so streamed progress was never live and listen streams blocked forever; SSE now parses incrementally with a body-close watchdog

### Phase 3 — CacheableResult + ordering (`bc5ac3e7`)
- `ttlMs`/`cacheScope` (default `private`) on list/read results; tools/resources sorted at the response boundary — byte-stable serialization asserted (prompt-cache economics)

### Phase 4 — server MRTR + sealed state (`1365f33e`)
- Handlers pause via typed `protocol.InputRequiredError` (no signature migration); retries deliver `inputResponses`/`requestState` via context; `internal/mcpstate` seals state per decision D2 (AES-256-GCM, HKDF domain-separated, principal+expiry bound, uniform failure) with replica-hop interop asserted
- Zero-fakes E2E: complete MRTR cycle over real HTTP with sealed state verified

### Phase 7 — PRM (`995ad39c`)
- RFC 9728 metadata at `/.well-known/oauth-protected-resource` on auth-configured deployments (`LOOM_MCP_RESOURCE_URL`, documented)

## Test plan
- Every phase ships its brief's scenario table; two new zero-fakes E2E tests exercise Loom's client against Loom's server (stateless negotiation, subscriptions, full MRTR)
- The race detector caught and fixed two real races during development (subscription err field, dedupe expiry sweep)
- `just check` green on every commit (matches GitHub CI)

Remaining (tracked in `docs/planning/mcp-2026-07-28/_index.md`): Phase 8 conformance matrix + SDK interop CI (next), optimizer confirmation consumer, sealer deployment wiring.


Ed-Jaramillo (changes requested):
Requesting changes on the logical PR #328 delta (`228adc7...3859059`).

I did a final assumption pass against the current head, the official 2026-07-28 Streamable HTTP, authorization, stdio, and subscriptions specifications, and the ratified local D1/D2 briefs. All 15 findings below remain valid.

## Blockers

1. **`pkg/mcp/transport/streamable_http_server.go:190` — modern admission is fail-open.** Admission is selected from body `_meta` alone and `IsStatelessVersion` is an open-ended date comparison, so future versions execute and a legacy-shaped body with modern headers bypasses validation. The transport requires exact version support and required/matching `MCP-Protocol-Version`, `Mcp-Method`, applicable `Mcp-Name`, and recognized `Mcp-Param-*` headers, with HTTP 400 and `-32020`/`-32022` on failure.

2. **`pkg/mcp/transport/streamable_http_server.go:160` — no `Origin` validation.** Streamable HTTP requires validating `Origin` on every incoming connection and returning HTTP 403 when a present Origin is invalid. Add an explicit trusted-Origin policy before body/session processing.

3. **`pkg/server/multi_agent.go:532` — blank identities are ownership wildcards.** Normal `Chat`/`Weave` session creation does not stamp `UserID`, while `sessionAccessibleBy` treats a blank caller or owner as globally accessible. This permits cross-user access and, with optional auth, anonymous traversal. Stamp ownership from context and make single-tenant compatibility an explicit mode.

4. **`pkg/server/multi_agent.go:749` — `Weave`/`StreamWeave` bypass or fall through authorization.** A supplied `agent_id` skips ownership lookup, and a failed lookup falls through to the explicit/default agent with the foreign `session_id`. Resolve and authorize every existing session before agent selection; an inaccessible session must not become a create/fallback path.

5. **`pkg/agent/session_store.go:474` — SQLite ownership does not cover child data or delete side effects.** `SaveSession` trusts `session.UserID` over the authenticated context; child reads/writes predicate only on `session_id`; and `DeleteSession` runs filesystem cleanup and hooks without checking whether an owned row was deleted. Make context identity authoritative, guard child operations through the owned parent, and require one affected row before external cleanup.

6. **`cmd/loom-mcp/main.go:302` — accepted tokens are not bound to the advertised MCP resource.** PRM advertises `LOOM_MCP_RESOURCE_URL`, but JWT audience validation independently defaults to `authenticated`. The MCP authorization specification requires accepting only tokens issued specifically for this resource. Bind validation to the canonical PRM resource and fail startup on inconsistent auth/PRM configuration.

7. **`pkg/mcp/client/client.go:477` — stream-loss reissue is not exactly-once outside one Weave process.** Every stateless method is retried after stream loss, but only `tools/call` receives a key and looms consumes it only for `Weave`/`StreamWeave`; workflows, session/UI/artifact mutations, and arbitrary `RawRequest` methods can execute twice. D1 also requires a durable `run_dedupe` table, not the process-local map shipped here. Gate retry on end-to-end idempotency or implement operation-wide durable dedupe.

8. **`pkg/server/multi_agent.go:1142` — terminal delivery failure poisons a completed dedupe result.** Successful execution sets `dedupeFinal`, but the deferred `finish` records the final `stream.Send` error as the run error. A retry then receives the delivery failure instead of the completed result. Cache execution outcome independently from response delivery.

## Major

9. **`pkg/mcp/server/meta.go:121` — valid successful results can panic during stamping.** JSON `null` unmarshals into a nil map, so a nil result or `"_meta": null` panics on assignment. Allocate/validate both maps after unmarshal and add regressions for both shapes.

10. **`pkg/mcp/server/subscriptions.go:92` — subscription IDs collide across HTTP clients.** JSON-RPC IDs are client-scoped, but active subscriptions use one server-global map keyed only by request ID. Two clients using ID `1` collide. Use a server-unique internal key or scope the registry to the originating connection.

11. **`pkg/mcp/server/server_stream.go:45` — stdio advertises notification capability but cannot establish modern subscriptions.** `subscriptions/listen` exists only on the HTTP streaming path; stdio returns `MethodNotFound` and can emit untagged legacy notifications even though the stdio binding requires acknowledged, tagged subscriptions and cancellation. Implement stdio subscriptions or make capabilities transport-specific.

12. **`pkg/server/weave_dedupe.go:82` — dedupe admission is unbounded and O(N).** Every new key scans the full map under one mutex; there is no capacity, principal quota, or key-size limit, and in-flight entries never expire. Add bounded admission and indexed/periodic expiry; the durable D1 store needs a unique key and expiry index.

13. **`internal/mcpstate/sealer.go:54` — request-state rotation violates ratified D2.** D2 requires sealing with a new key ID while retaining old keys through their expiry horizon. `Sealer` holds one key and rejects all prior key IDs immediately. Use a keyring keyed by key ID.

## Minor

14. **`pkg/mcp/server/server.go:201` — modern unknown methods return HTTP 200.** The transport requires HTTP 404 plus the JSON-RPC `MethodNotFound` body. Map `-32601` to 404 on both JSON and SSE-capable stateless paths.

15. **`Justfile` (`check` recipe) — interop checks do not prove their stated guarantees.** CI has a separate MCP interop job, but `just check` claims to match CI while omitting `interop`. The interop test also comments that it checks cache fields and modern negotiation but asserts neither. Add the assertions and include `interop` in `check`, or narrow the claims.

The migration work is substantial, and the latest patch resolves the earlier discovery and `requestState`-presence issues. The remaining admission, Origin, tenant-isolation, OAuth audience, and exactly-once defects should block merge. The branch also currently conflicts with merged `main` and still displays PR #327's history.

ilsiepotamus:
All 15 findings verified against the published transport/authorization/stdio/subscriptions specifications and the ratified briefs — every one held up — and all 15 are addressed. The branch is also rebased onto merged main, so the PR now shows only the server-side delta. Point by point:

### Blockers

1. **Fail-open modern admission — fixed (e64ea2a9).** Admission now routes on body *or* header declaring a modern revision (closing the legacy-shaped-body-under-modern-headers bypass) and validates strictly: the declared revision must be an exact member of the configured supported set (default `{2026-07-28}`; futures are rejected 400/`-32022` with the supported list), and requests must carry matching `MCP-Protocol-Version` and `Mcp-Method`, plus `Mcp-Name` matching the body value **after Base64-sentinel decoding** for tools/call, prompts/get, and resources/read (`protocol.DecodeHeaderValue` is new). Malformed `Mcp-Param` sentinel payloads are rejected as invalid characters; well-formed ones are ignored — no tool served here declares `x-mcp-header` annotations, so no `Mcp-Param` header is *recognized* by this server, which is the spec's forwarding rule for unrecognized params. Notification POSTs carry no header requirements in this revision and get version-support validation only. Conformance fixtures updated to send conformant requests.

2. **Origin validation — fixed (e64ea2a9).** A present-but-invalid `Origin` is rejected HTTP 403 before any body or session processing. Default policy admits loopback origins only; `AllowedOrigins` configures an explicit allowlist (`"*"` opts into any); absent Origins (non-browser clients) pass.

3. **Blank-identity wildcards — fixed (2cde191c).** Ownership is stamped at the single central creation point (`Memory.GetOrCreateSessionWithAgent`) from the authenticated context, covering Chat, Weave, StreamWeave, and workflow paths — not just the CreateSession RPC. Single-tenant permissiveness is now the explicit mode: `SetEnforceSessionOwnership(true)` (wired to `config.Server.Auth.Enabled` in looms) makes blank identities never match — anonymous callers see nothing, and unowned legacy rows are not world-readable.

4. **Weave/StreamWeave authorization ordering — fixed (2cde191c).** Existing sessions are resolved and authorized before *any* agent selection: `findSessionOwner` distinguishes does-not-exist (safe to create) from exists-but-foreign, which surfaces as NotFound — indistinguishable from missing — on both RPCs, including with an explicit `agent_id` (the exact bypass), and an explicit agent that disagrees with the owning agent is FailedPrecondition. Regression: foreign session + explicit agent_id → NotFound, never a fallback execution.

5. **SQLite ownership depth — fixed (2cde191c).** The authenticated context is authoritative on save (a caller presetting `session.UserID` cannot claim another owner — regression proves the attacker keeps the row, the "victim" never sees it); child reads/writes (messages, tool executions, eviction, folds, snapshots) are guarded through the owned parent — foreign writes fail as not-found, foreign reads are empty; and `DeleteSession` requires exactly one affected row before filesystem cleanup or hooks — a foreign delete is now a complete no-op (regression asserts hooks don't fire). `LoadAgentSessions`/`LoadMessagesForAgent` turned out to be dormant interface surface with no runtime callers; documented as such.

6. **PRM audience binding — fixed (01b88166).** When PRM advertises `LOOM_MCP_RESOURCE_URL`, JWT audience validation is bound to that canonical resource; an explicit `LOOM_SERVER_AUTH_SUPABASE_AUDIENCE` must equal it or startup fails naming both values. Without PRM, prior behavior stands.

7. **Stream-loss re-issue scope — fixed on the client (1a4efac6); dedupe scope documented.** Re-issue is now gated on end-to-end idempotency: idempotent-by-nature reads (`server/discover`, `tools/list`, `prompts/*`, `resources/*`) and keyed `tools/call` retry transparently; everything else — `RawRequest` mutations included — surfaces the typed `CodeStreamLost` to the caller, who owns the retry decision. On the durable-table point: the phase index records the deliberate deviation — the in-memory dedupe is correct for the deployed topology (N stateless bridge replicas in front of **one** stateful looms, which is the state owner), and the durable `run_dedupe` store rides decision D3 (multi-replica looms) where it becomes load-bearing. If you read D1 as requiring the table for single-replica looms too, happy to pull it forward — it's a migration plus a store swap behind the same `begin/finish` seam.

8. **Delivery failure poisoning a completed result — fixed (2cde191c).** StreamWeave's dedupe finish now caches the execution outcome independently of response delivery: once the run completed, the result stays joinable even when the final `stream.Send` failed — that failure is exactly why a re-issue is coming.

### Major

9. **Nil stamping panics — fixed (441fc433).** JSON `null` results and handler-set `"_meta": null` both stamp instead of panicking, with regressions for both shapes.

10. **Subscription ID collisions — fixed (441fc433).** The registry key is a server-unique sequence; the client-facing identity remains the request id, tagged per stream in `_meta` (streams never mix). Regression: two clients both using id 1 hold concurrent acknowledged subscriptions and both receive published changes.

11. **stdio subscriptions — implemented (e004b882).** The stdio Serve loop now implements the binding: connection-scoped registration (duplicate ids per connection rejected; ids reusable after cancel), acknowledgment strictly before that subscription's notifications, tagged interleaved delivery, and `notifications/cancelled` referencing the listen id as the cancellation signal (no response). Untagged legacy notifications are era-gated: they flow only after a legacy `initialize` handshake; under the modern era the server MUST NOT send unrequested types, so the legacy channel drops with a debug log. Wire-level tests run Serve over real pipes.

12. **Unbounded dedupe — fixed (1a4efac6).** Admission is bounded (256-byte key cap, 4096 entries) with amortized sweeps (at most once a minute) plus per-entry O(1) expiry on the hit path — a stale completed result is never joined. On saturation, after evicting expired and then oldest-completed entries, new keys run without dedupe (the at-least-once semantics of key-less requests); in-flight entries a joiner may be waiting on are never evicted.

13. **Sealer rotation per D2 — fixed (01b88166).** Sealing always uses the current key; unsealing resolves the blob's key-ID prefix against a ring that retains previous keys (`NewSealerWithPrevious`; deployment convention `LOOM_MCP_STATE_SECRET_PREVIOUS`), so an MRTR retry straddling a rotation stays valid through its own TTL. Regression covers pre-rotation unseal, new-key rejection by the old sealer, and rejection once a key leaves the ring.

### Minor

14. **Unknown modern methods → 404 — fixed (e64ea2a9).** Handler-produced `-32601` maps to HTTP 404 (version/header/capability failures to 400) on the stateless path, and the SSE path engages only for methods that actually stream (`tools/call`, `subscriptions/listen`), since it must commit its status before the handler runs. Matrix expectations updated (`ping` and `resources/subscribe` under stateless `_meta` now assert 404).

15. **Interop honesty — fixed (latest commit).** The suite now asserts what its comments promised — `ttlMs`/`cacheScope` surviving official-SDK decoding, and negotiation landing on `2026-07-28` stateless — and `just check` includes the `interop` recipe, so "matches GitHub CI" is true again.

### Verification

`just check` green (now including the SDK interop suite), full `-tags "fts5 interop" -race` across `pkg/mcp/...`, `pkg/agent`, `pkg/server`, `pkg/shuttle`, and `internal/mcpstate`; the strict-admission, Origin, ownership, dedupe-bounds, keyring, and stdio-subscription behaviors each carry dedicated regressions referencing their finding numbers.

Ed-Jaramillo (commented):
Follow-up verification pass on the current head. First off: the corrections adopted in `e3bc20c` (discover stamping, `requestState` presence) verified clean — none of the findings below originate there, and the dual-mode admission design, the sealed-state work, and the SDK interop suite are genuinely solid pieces. This pass reconfirms a few findings from my standing review with new detail, adds six new ones, and closes out three suspicions that turned out to be fine.

## Sharpened detail on standing findings

- **Finding 3 (blank-identity wildcard) is reachable with auth fully enabled**, not just on pre-migration rows: sessions auto-created by `Weave` are never stamped with an owner in memory (only the `CreateSession` RPC stamps `UserID`, `multi_agent.go:2803`), so `sessionAccessibleBy` (`multi_agent.go:533`) admits every authenticated caller to them for the process lifetime — both `GetConversationHistory` and `DeleteSession` pass the gate.
- **Finding 9's failure mode is a panic, not a null passthrough**: `json.Unmarshal([]byte("null"), &obj)` succeeds and leaves the map nil, so `stampResult` (`meta.go:127`) panics on the `resultType` assignment — connection abort on HTTP, process crash on stdio. The `if result != nil` guard at `handlers.go:109` admits exactly this case for any third-party `ResourceProvider` returning `(nil, nil)`.
- Findings 10 (subscription ID collision) and 12 (dedupe growth) reproduce unchanged on this head.

## New findings

1. **SQLite ownership predicate covers five methods; the rest of the store is unscoped** (`pkg/agent/session_store.go:982` and siblings). `LoadMessagesForAgent`, `LoadMessagesFromParentSession` (`:1094`), `SearchMessages` (`:1716`), `SearchMessagesByAgent` (`:1840`), `GetStats` (`:1595`), and `loadMessagesLocked` (`:735`) all query with no `user_id` filter — the postgres twin filters each of these (`postgres/session_store.go:557`, `:590`). Cross-user reads via `LoadMessagesForAgent` and both FTS5 search paths reproduce in a quick repro test. Adjacent to standing finding 5 but distinct query paths.

2. **Dedupe caches cancellation errors for the full 10-minute window** (`pkg/server/weave_dedupe.go:58`). `finish()` stamps `expiresAtNano` unconditionally, and in `StreamWeave` the deferred `finish(dedupeFinal, weaveErr)` (`multi_agent.go:906`) runs when the owner's stream dies — so a client disconnect caches the cancellation error, and the spec-mandated re-issue after stream loss (the exact scenario D1 exists for) joins the cached failure instead of re-executing. Sibling of standing finding 8, different trigger. Suggest either not caching cancellation/transient errors, or ratifying error replay in the D1 brief with a test.

3. **Legacy streaming clients get MRTR errors as an opaque `isError` blob** (`pkg/mcp/server/server_stream.go:74`). The `&& meta.Stateless()` gate short-circuits for non-stateless clients, downgrading `InputRequiredError` to `CallToolResult{IsError: true}` carrying the raw internal error string — while the synchronous path (`server.go:228`) returns a proper `-32600` JSON-RPC error for the identical condition.

4. **`ParseRetryInput` swallows the unmarshal error** (`pkg/mcp/protocol/mrtr.go:99`). A retry with malformed `inputResponses` parses as a zero `RetryInput`, indistinguishable from an initial call — the handler re-pauses and the client re-elicits input. Bounded by `MaxRounds`, but an explicit `InvalidParams` would fail fast instead of silently looping.

5. **Dropped notifications leave no client-visible gap signal** (`pkg/mcp/server/subscriptions.go:159`). The drop-on-overflow policy is documented as intentional with refetch-on-reconnect as recovery — but the client is never told a drop occurred, so it doesn't reconnect and serves stale data until `ttlMs` expiry.

6. **`ValidateRequest` failures respond with `"id": null`** (`pkg/mcp/server/server.go:183`). Unmarshal has already succeeded at that point, so `req.ID` is available; JSON-RPC 2.0 requires echoing a detectable id, and `null` breaks correlation for multiplexing clients. Pre-existing line, noting it since it sits beside this PR's insertion.

## Verified clean, for the record

- **Header-less POST admission**: pre-existing behavior matching the transport's documented localhost-only contract; the new state-owner enforcement is the right layer for it.
- **`dedupeScope` separator collision**: not reachable — `validateUserID` rejects control bytes on both identity ingress paths.
- **Empty-userID postgres `Internal` error**: unreachable through looms; the interceptor always injects a non-empty default.

Priority-wise, the two ownership items (the `Weave`-session stamping gap and the unscoped store methods) are the ones I'd tackle first — everything else is contained. Happy to pair on any of these.

ilsiepotamus:
Verified every item in this pass against the current head and the specs. One calibration note first: the four "sharpened standing" reproductions reference the pre-rebase tree (the `228adc7...` delta — e.g. `multi_agent.go:2803`, the old `meta.go:127` shape). Those were fixed in the round-1 force-push (`21e5f6bc`) and each has a passing regression on this head — details per item below. The **six new findings all verified real** (one had a twist worth calling out), and all six are fixed. Thank you for these — the hostile-parent variant of #1 and the cancellation-caching in #2 are exactly the class of bug that survives casual review.

### Standing findings — state on the current head

- **Finding 3 (blank-identity / Weave stamping):** Ownership stamping was centralized in round 1 at the single session-creation funnel `Memory.GetOrCreateSessionWithAgent` (`pkg/agent/memory.go:415`), stamped from the authenticated context; `multi_agent.go` no longer has any direct creation site, so Weave/StreamWeave/Chat/workflow sessions are all owned. `TestSessionsStampedAtCreation` and `TestWeaveForeignSessionIsNotFound` pass on this head, and `SetEnforceSessionOwnership(true)` (wired to auth-enabled) makes blank identities never match.
- **Finding 9 (null-shape panic):** Both shapes are guarded since `441fc433` — `json.Unmarshal("null")` leaving a nil map (`meta.go`, obj-nil guard) and handler-set `"_meta": null` (meta-nil guard). `TestStampResultNilShapes` covers both, including the typed-nil provider result, which marshals to `null` and lands in the same guard.
- **Findings 10, 12:** registry keys are server-unique (`subSeq`) since `441fc433`; dedupe admission is bounded (256-byte keys, 4096 entries, amortized sweep + O(1) hit-path expiry) since `1a4efac6`. `TestSubscriptionIDsDoNotCollideAcrossClients` and `TestDedupeBoundedAdmission` pass on this head.

### New findings — all real, all fixed

1. **Unscoped SQLite store reads — fixed (8a9a7ea9).** Confirmed unscoped on this head: `SearchMessages` (both branches — the all-sessions branch had no ownership predicate at all), `SearchMessagesByAgent`, `LoadMessagesForAgent`, and `GetStats`. All now scope through `sessions.user_id`, mirroring the postgres twin; `loadMessagesLocked` carries a defense-in-depth predicate so the helper can never become a cross-user read if a future caller forgets to guard. The twist: `LoadMessagesFromParentSession` was child-guarded in round 1, **but your instinct was right anyway** — `parent_session_id` is caller-supplied at save time, so an attacker's own child linked to a foreign parent read the parent's coordinator/shared messages. The parent lookup and the message read are now both owner-predicated (a hostile linkage reads as parentless). Regressions cover every path and were proven red pre-fix; `TestSQLiteHostileParentLinkageDoesNotLeak` is the interesting one.

2. **Cancellation cached for the dedupe TTL — fixed (4df0ca6d).** Exactly as described: the deferred finish cached `context.Canceled` for ten minutes, so the re-issue joined the failure D1 exists to recover from. Chose your first option: `finishAndRelease` releases entries whose outcome is a cancellation or deadline (bare or gRPC-coded) — the re-issue re-executes as a fresh owner — while deterministic execution errors stay joinable (replaying them is correct dedupe), and joiners already waiting still observe the error through the closed channel. No error replay was ratified into D1; the brief's semantics are unchanged for durable outcomes.

3. **Legacy streaming MRTR downgrade — fixed (a4b41489).** The streaming path now returns the same explicit `-32600` ("requires an MRTR-capable client") the synchronous path returns, instead of an `isError` result carrying the raw internal error string.

4. **`ParseRetryInput` error swallow — fixed (a4b41489).** The parser now distinguishes absent retry members (not a retry, parses clean) from present-but-malformed ones, which surface as an error carried on the request meta; both dispatch paths (synchronous and streaming) answer it with `InvalidParams` before any handler runs. No more silent re-elicit loop.

5. **Invisible notification gaps — fixed (8eb20a71).** Agreed the drop policy defeated its own recovery story. Overflow now terminates the subscription: the HTTP listen stream closes, and on stdio — where the connection outlives the subscription — the server sends `notifications/cancelled` for the listen id (the binding's cancellation signal, used symmetrically), after which the id is reusable for the re-subscribe. The client's documented recovery (re-subscribe, refetch, `ttlMs` keeps it cheap) now actually triggers. The phase-5 brief's drop-oldest line is revised to match, and regressions cover both transports.

6. **`ValidateRequest` errors with `"id": null` — fixed (a4b41489).** The id is echoed; `ValidateRequest` never faults the id itself (it checks the jsonrpc version and method only), so the echo is always safe.

### Verification

`just check` green on the head (lint, full `-tags fts5 -race` suite, build, SDK interop), each fix carries a regression referencing its finding number, and the three ownership regressions were verified red against the pre-fix code.

Ed-Jaramillo (commented):
Re-verified the whole fix round against the current head (`8eb20a71`), and this is genuinely close. **All ten findings from the last pass are fixed**, each with a targeted regression, and the corrections are real rather than nominal — the central ownership stamping in `GetOrCreateSessionWithAgent`, the cancellation-release semantics in the dedupe layer, and the overflow-terminates-the-subscription redesign are all the right calls. The sealer keyring, the per-request JWT audience binding, and the interop suite (now gating `just check`) also come through clean. Thank you for the thorough turnaround on these.

Adversarially testing the fixes themselves surfaced three items — one I'd treat as a blocker, one should-fix, one nit. To be clear up front on the first one: it lives entirely inside code this PR introduced, and it only became reachable *because* of the fix, so it's a second-order finding rather than anything pre-existing.

### 1. Blocker — store-only sessions bypass the new Weave authorization gate (`multi_agent.go:549`)

The ownership fix authorizes Weave/StreamWeave through the new `findSessionOwner`, which inspects only in-memory sessions. A session that lives only in the store — after a restart, after memory eviction, or simply not yet loaded into this process — is invisible to that gate, so authorization is skipped for it.

Concrete path, `enforceOwnership=true`:
1. Victim's session exists only in the store (e.g. post-restart), owned by `victim`.
2. Attacker calls `Weave{SessionId: victimID}`. `findSessionOwner` finds nothing in memory → `exists=false` → the ownership check is skipped. Weave falls through to agent selection and `GetOrCreateSessionWithAgent` creates a fresh in-memory session under `victimID` stamped `UserID="attacker"`. The attacker's call succeeds.
3. The victim then calls `Weave{SessionId: victimID}` for their own session, hits the attacker-owned shadow, and gets `NotFound` — locked out of their own session.

No victim *data* leaks — the store upsert/read guards from `2cde191c` hold, so the persisted row stays victim-owned — but it's an attacker-controlled denial-of-service / session-ID squat in exactly the auth-enabled mode the fix targets. Verified with a temporary red probe.

The encouraging part: the PR already has the right pattern a few hundred lines away — `DeleteSession` (`multi_agent.go:3101`) falls back to `s.sessionStore.LoadSession` when the session isn't in memory. Giving `findSessionOwner` the same owner-scoped store fallback closes this.

### 2. Should-fix — cancellation laundering defeats the dedupe release on the unary path (`multi_agent.go:915`)

The release fix (`4df0ca6d`) is right, but there's a path where it doesn't get a chance to fire. Agent errors are wrapped as `status.Errorf(codes.Internal, "agent execution failed: %v", err)` before `finishAndRelease` classifies them, and `%v` discards the error chain. So on the **unary** Weave path (`bridge_handlers.go`), a caller disconnecting mid-run produces a wrapped `context.Canceled` that reaches the classifier already looking like a durable `Internal` — and gets cached for the full ten-minute window, which is the exact scenario D1's release logic exists to avoid. The streaming path's own `stream.Context().Done()` branch usually wins on disconnect, so this mainly bites the unary path; a mid-stream `Send` failure on connection closure (mapped to `codes.Unavailable`) is a related variant. Classifying before wrapping — or preserving the chain with `%w` — restores the intended behavior.

### 3. Nit — stdio subscription pump leaks its child context on self-termination (`stdio_subscriptions.go:155`)

`subCtx, cancel := context.WithCancel(ctx)` is only cancelled via the connection-level `cancelAll`. When the pump exits on its own via the overflow (`:197`) or send-failure (`:213`) path, the defers remove the map entries but never call `cancel`, so the child context stays registered on the Serve context until the connection tears down. Bounded per connection, but it accumulates across repeated overflow/re-subscribe cycles on a long-lived stdio connection. A single `defer cancel()` in the goroutine fixes it (idempotent with the map-routed cancels).

### Verified clean

The sealer's rotation/nonce/size-bound handling, the exact-match per-request JWT audience check, and the interop assertions all held up under scrutiny; the full `./pkg/mcp/...` tree and both ownership packages pass under `-race`. One observation, not a finding: the sealed-state path isn't wired to a production handler in this PR yet, so its unseal side isn't network-reachable here — noting it only so it's on the radar for whichever PR does wire it.

Net: item 1 is the one I'd want closed before merge, and it's a small, well-scoped change since the pattern already exists in the file. 2 and 3 are quick. Happy to pair on any of them.

ilsiepotamus:
All three verified real and fixed in `906e5b99` — the squat scenario reproduced exactly as you described it, and it's a genuinely good second-order catch: the gate I added created the very blind spot it was meant to close.

1. **Store-only session squat — fixed.** `findSessionOwner` now falls through to the store, but your `DeleteSession`-pattern suggestion needed one extension to fully close the hole: the owner-scoped `LoadSession` can resume the caller's own post-restart session (agent resolved via `getAgent`, since stored sessions record config names, not registry GUIDs), but on a miss it *cannot* distinguish exists-but-foreign from absent — and that distinction is the whole game for a create-on-missing flow. So the store gained `SessionExists`, a deliberately unscoped existence-only primitive (documented as an authorization check, not a data read; both SQLite and postgres implement it): foreign → NotFound with nothing created, absent → safe to create, and store errors fail closed as inaccessible. A stored session bound to an agent this process no longer registers is FailedPrecondition rather than a silent resume under whatever agent is default. Regression: `TestWeaveStoreOnlySessionCannotBeSquatted` covers attacker-NotFound-with-no-shadow-minted, owner resume, and the unregistered-agent case.

2. **Cancellation laundering — fixed, both ways.** The wrap sites now use `wrapAgentError`, which preserves cancellation/deadline as their own gRPC codes (also simply the correct codes for callers — `Internal` for a client disconnect was wrong independent of dedupe). And because laundering can happen in layers we don't control, `finishAndRelease` is now context-aware: a dead request context is never a durable outcome, whatever the error looks like — which covers your `Unavailable` mid-stream-Send variant too. Regressions: `TestDedupeReleasesLaunderedCancellation` (a `%v`-flattened Internal with a dead context is released; the re-issue re-executes) and `TestWrapAgentErrorPreservesCancellationCodes`.

3. **stdio pump context leak — fixed.** `defer cancel()` in the pump goroutine; idempotent with the map-routed cancels.

On the observation: agreed — the sealed-state unseal side is not network-reachable in this PR. Wiring the sealer into the deployment mains (env secret sourcing per the Phase 4 brief) is tracked as the open Phase-4 item and will be its own PR, where the unseal path gets its wire-level coverage.

`just check` green on the head (lint, full `-tags fts5 -race` suite, build, SDK interop).

Ed-Jaramillo (commented):
Verified `906e5b99` against all three round-3 items, adversarially and under `-race`. **All three are fixed** — this clears everything I had open across the whole review.

- **Store-only session squat (blocker) — fixed.** `findSessionOwner` now takes a context and, on an in-memory miss, falls back to an owner-scoped `LoadSession` plus an unscoped `SessionExists` existence probe. An attacker's `Weave` against a victim's store-only session now returns `NotFound` *before* any in-memory session can be minted, so there's no shadow session and no lockout. It fails closed on a store error and releases `s.mu` before the DB call, so no lock is held across I/O. `TestWeaveStoreOnlySessionCannotBeSquatted` drives the real store-only path (writes the victim session directly, never in memory) and asserts no shadow session is created. Nice reuse of the `DeleteSession` store-fallback pattern, and the SQLite/postgres implementations match.
- **Dedupe cancellation laundering — fixed.** Belt-and-suspenders: `wrapAgentError` uses `errors.Is` before flattening so wrapped cancellations surface as `codes.Canceled`/`DeadlineExceeded`, and `finishAndRelease` additionally releases whenever `ctx.Err() != nil` — which also covers the `codes.Unavailable` mid-stream `Send` variant. Genuine deterministic errors with a live context still cache and join, and the regression uses a real `%v`-flattened coded error rather than a bare cancellation.
- **stdio subscription context leak — fixed.** A single `defer cancel()` now covers every pump exit path, idempotent with the connection-level `cancelAll`.

A few non-blocking observations for whenever it's convenient, none of which I'd hold merge for: the legitimate resume path now does a full `LoadSession` in the gate and `ag.Chat` loads the session again (a lighter owner+agent lookup would avoid the double read); `SessionExists` has no direct unit test (it's covered indirectly through the server test); and it's worth being aware that with a store configured, every `Weave` carrying a session id now incurs a store round-trip and fails closed during a DB outage — correct, just new behavior.

From my side this is good to merge. Thanks for the fast and careful turnaround across all three rounds — the ownership model in particular is in solid shape now.

Ed-Jaramillo (approved):
Round-3 fixes verified — approving.

PR #329, opened 2026-08-18
mcp-npd: LLM-layer correctness, batch-first tool surface, explicit-only graph memory

mrg-ray:
32 commits on top of develop; develop is fully contained (no conflicts). This is the loom side of the tera CLI + data-eng-bench campaign: agent-runtime changes measured on that bench, LLM-layer cost/thinking correctness, and the batch-first builtin tool surface.

## Behavior changes (default-on — each one deliberate)

- **Automatic graph-memory extraction + recall injection removed** (e4a1e09d). Extraction fired on every user message and `injectGraphMemoryContext` joined it before turn 1 — a 10–30s synchronous barrier per message, plus cadence re-fires. Measured on an ade-bench sweep: 520 bookkeeping calls, ~66 min model time, $4.53 for graphs nothing recalled. Graph memory is now written solely through the explicit `graph_memory` tool.
- **Compression profile thresholds move** (1f3293b8): critical 70→80 (data_intensive), 75→90 (balanced), 85→92 (conversational); data_intensive warning 50→45. Compaction fires later. Docs updated to match.
- **Session `task_list` joins the base tool set** (df580cd0, 7ef7bc72, 9c1d88a6): always-on base tools are now load_pattern, query_tool_result, recall, task_list; batch updates, output_contract, `description` accepted as title alias.
- **Skill instructions outrank ROM operating rules** (87deb6b5); ROM START_HERE drops the size-fear line (aa8b71b3).
- **Cost figures change because they were wrong** (829c854a, 90d3bac7, 21748ad1, 7c3ee87d, 8f1b949c, d2688fbf, parts of 1f3293b8): opus-5 was billed at sonnet's card (substring fallback hit before the catalog — catalog is now consulted first, incl. un-suffixed profile ids); cache tiers billed separately across gateway clients (openai, azure, gemini, ollama paths); a gateway-reported cost header wins over the local estimate; OpenAI cache reads discount by model generation (0.10x gpt-5/o-series, 0.50x gpt-4 era); gateway-proxied Claude ids price at Anthropic rates, never GPT's; Bedrock third-party (GLM, DeepSeek) + fable-5 profile ids catalogued. An uncatalogued model now reports cost 0 with a warning instead of a silently invented rate.
- **Batch-first advertised schemas on the SQL builtins** (80c0c250, ac7bf311): `execute_query` advertises only `statements[{label,sql}]` — per-statement read-only gate, isolation, labeled sections, row cap 50 with "+N more" footer, 40-char cell elision, per-call render budget shared across the batch. `table_profile` takes `tables[]` (max 20/call), per-table isolation. Legacy inputs (`sql` string, bare-string entries, `table`) are coerced, so existing callers keep working; only the model-visible contract changes.
- **Shell/file output hygiene** (80c0c250): shell stdout/stderr are ANSI-stripped and capped (12KB, 4KB head + 8KB tail); wildcard file reads skip build-artifact dirs (`target/`, `dbt_packages/`, `logs/`) unless the path names them exactly.
- **Context-compilation follow-ups** (21b61a68, 253d0dbc, 6a626eb9, 27a28a9c, 7ef7bc72, parts of 1f3293b8): stub previews are metadata-shaped (schema, not a data slice); the cache breakpoint never parks on a text-empty message; the workload profile selects the relief marks; a till-NOW cache marker rides as a 4th breakpoint on the litellm path (native clients stay capped at 3); rung-0 relief sweeps/evicts/folds the current turn behind the last assistant message; the transient tail rides user-role (fixes gateway cache hoisting).

## Additive

- **Thinking**: level plumbed through the native Anthropic and Bedrock clients (4ec18088); on adaptive models (4.6+/5-family) the level maps to `output_config.effort`, low→max — auto/empty omit the field (a38ce907); litellm wire support for extended thinking: adaptive request, both observed response shapes, split-shape block completion, verbatim in-turn replay, settle strip, thinking blocks never persisted (1f3293b8).
- **Bedrock non-Anthropic models route through Converse** (057f8594): the previous path sent Anthropic-format payloads whose tool arguments never decoded (smithy documents don't json.Marshal — arguments were always `{}`), so no working behavior existed to change; decode now goes through `UnmarshalSmithyDocument`.
- **`dbt_build` builtin** (f4f38d40): builds seeds → whole project (never a selector) and reports the grain of freshly authored models.
- **File tools** (7ef7bc72): `file_read` gains glob reads and project-wide search (gitignore-aware — new dependency `sabhiram/go-gitignore`); `edit_files` added; `file_write` extended.
- **Shell sandbox guard** (df580cd0): `LOOM_SANDBOX_DIR` is honored exactly as the tool's Suggestion advertises — an anchored workspace outside the data dir no longer rejects every call.
- **Skills**: skill-declared MCP tools mount directly at activation (f1d15f48, example in `skills/teradata-sql-analytics.yaml`); fold-unmount contract under test — refcounted skill tools, base-tool immunity, a real fold fires deactivation (96098b7a, 1c8c3503); base-subset surface proven (6 tools in context, not the full 47) (a599f763).
- **Fabric**: multi-database duckdb DSN — comma-separated attach, qualified-only (ac7bf311).
- **Telemetry**: client-side latency split stamps on the openai stream path (06fbd786).

## Tests / docs / infra

- New `test/context-optimiser` rig (mechanisms, rig, rung0) exercising the relief machinery end-to-end (1f3293b8).
- Server test logger survives background workers (7adc2897); broadcast/workflow-reply tests updated alongside.
- Two tests that still pinned pre-change behavior updated to the current design — graph-memory suppression (extraction never auto-enables) and the base-tool set (e8d68666); one pre-existing sub-millisecond duration flake fixed at the mock, which also flakes on main (1307644a).
- Four architecture/guide docs refreshed to the current thresholds and design; `minL1Messages` documented as a vestigial parsed field that drives nothing (1f3293b8).

## Verification

- Full suite green: 93/93 packages (`go test -tags fts5 ./...`).
- Tag `bench-2026-08` on the fork (mrg-ray/loom) marks this exact tree as the source of the published bench numbers.


ilsiepotamus (changes requested):
Overall this PR measures well — I ran it against LongMemEval (temporal-reasoning, 30 questions, Opus 4.6 agent + judge): **PASS 29/30 (97%) vs the main-lineage baseline's 24/30 (80%)**, zero pipeline errors, and the explicit-only graph memory change did not hurt recall. The remaining five question types are running now. This review is about one architectural item, not quality.

## Requesting changes: `task_list` duplicates `task_board`

The new `task_list` tool (`pkg/agent/task_list_tool.go`) overlaps the existing `task_board` in purpose — an agent-facing checklist for multi-step work — and in the **default profile both surface to the model simultaneously**:

- `task_list` is registered unconditionally at agent construction (`pkg/agent/agent.go:238` — "Registered by default")
- `task_board` is suppressed only under `tools.minimal` (`cmd/looms/cmd_serve.go:127-131` returns nil suppressions otherwise) and still auto-registers via progressive disclosure (`agent.go:1289`) whenever the task subsystem is wired — which a standard `looms serve` always does

So a default agent carries two checklist tools with different prompt vocabularies ("Your working checklist for multi-deliverable tasks" vs "You have a task_board tool for decomposing goals…"), different state models (session-ephemeral vs persisted kanban), and no reconciliation between them. Costs: tool-choice ambiguity for the model, doubled prompt/context spend, and transcripts where work is tracked in two places that never agree. The minimal-profile intent (batch-first surface) is sound — but the default profile is where most agents run.

## Proposal: absorb `task_list`'s genuinely good ideas into `task_board`

`task_list` adds four things `task_board` lacks (verified against the current action set: decompose/ready/claim/update/close/create/list/show/add_dep/board):

1. **Write-once acceptance criteria + `output_contract`** per item — additive fields on the task item (proto first, per repo convention)
2. **Push-rendered checklist**: the budgeted transient-tail block re-rendered every step (task_board is pull-only via list/show/board) — implementable as a renderer over board state with the same ~600-token budget
3. **Batch transitions** — an `updates` array on the update action
4. **Session-scoped ephemerality** — a `scope: session` board mode (not persisted, GC'd with the session, exempt from context relief)

All four are additive to `task_board`; none requires a second tool. I'm happy to make the `task_board` changes myself if that unblocks this PR — the task system's proto and store are mine to extend.

If there's a constraint I'm missing that makes a separate tool necessary (e.g., the NPD deployment must not link the task subsystem at all), the minimum ask is: make the two mutually exclusive in *every* profile, not just `tools.minimal`, so no model ever sees both.

A fuller review of the rest of the diff is in flight; I'll follow up with anything it finds.

ilsiepotamus:
Follow-up to my review with two additions:

**1. The duplication is deeper than tool overlap.** Both of `task_list`'s signature concepts already exist in the task system's proto: `Task.acceptance_criteria` (task.proto field 6, "Verifiable completion conditions") and `OutputPolicy` (collaboration.proto:58) — which models the output contract as a validated JSON Schema plus LLM-evaluated criteria, validator agents, and retry policy: a strict superset of `task_list`'s freeform `output_contract` string.

**2. The absorption is now concrete: #330** adds the two behaviors `task_board` genuinely lacked — write-once `acceptance_criteria` through the tool (`ACCEPTANCE_CRITERIA_LOCKED` on goalpost-moving, idempotent identical re-sends) and batch `updates` transitions with per-entry results. Tested against the real task store.

That leaves two `task_list` behaviors unabsorbed, and they're the interesting design conversation: **session-scoped ephemerality** and the **budgeted push-rendered checklist block**. The transient-tail render is a genuinely good idea — my ask is that it read from *board state* (e.g., tasks claimed by the current session) rather than a parallel tool with divergent state. If NPD needs boards that die with the session, a `scope: session` board mode seems like the right shape; happy to build that too if it unblocks dropping `task_list` from this PR.

ilsiepotamus:
Settling the two open items rather than leaving them hanging — #330 now also carries **`RenderSessionChecklist`** (`pkg/agent/task_board_render.go`): the session's working checklist rendered *from board state* per step — in-progress items with their acceptance criteria, pending/blocked titles, terminal counts, same ~600-token budget as your transient tail. Point the tail mechanism at it and `task_list` has no remaining job:

- criteria / output contract → already in the proto (`Task.acceptance_criteria`, `OutputPolicy`)
- write-once + batch transitions → #330
- push-rendered budgeted block → `RenderSessionChecklist` in #330
- session-scoped ephemerality → falls out of the view semantics: the block only ever shows the session's items (created-by or claimed-by); the persistence underneath is crash recovery, not a leak

With that, the concrete ask for this PR: drop `task_list_tool.go`, wire the transient tail to `RenderSessionChecklist`, and the rest of the PR stands as-is — the benchmark numbers are genuinely good.

ilsiepotamus:
**Retargeted this PR to `main`** (was `develop`). Rationale: `develop` is currently a strict 11-day-old snapshot of main (its HEAD `dc1c465e` is the merge-base; zero commits main lacks) — merging feature work into it would manufacture lineage divergence and strand the converse correctness fixes away from where everyone builds.

Heads-up for the rebase onto main — these files changed on both sides since `dc1c465e` and will likely conflict:

```
go.mod / go.sum
pkg/agent/agent.go
pkg/agent/context_compilation.go (+ test)
pkg/agent/types.go
pkg/llm/bedrock/client_sdk.go (+ tests)
pkg/shuttle/executor.go
pkg/types/types.go
```

Happy to help untangle the bedrock/client_sdk side if useful — the smithy-document decode fix and cache-token accounting in this PR are things main wants regardless.

ilsiepotamus (changes requested):
Overall this is a strong PR — the LLM-layer cost corrections, thinking plumbing, batch-first SQL surface, context-compilation follow-ups, and the Bedrock Converse fix all look right to me, and I'm fine with them shipping as default-on. (`task_list` feedback is being discussed in a separate thread — deliberately not covered by this review.)

Requesting changes on **one item only: the explicit-only graph memory commit (e4a1e09d)**.

## The problem

The commit removes automatic extraction + recall injection *unconditionally* — it deletes the call sites rather than gating them. Direction decision for loom: **automatic graph memory stays the default**; explicit-tool-only becomes a **config mode**. Downstream (tera-cloud) seeds long-lived user agents that depend on automatic capture + recall for cross-session memory, tunes it through five per-agent knobs that map onto `GraphMemoryConfig`, and surfaces the results in a memory-browsing UI — all of which would go silently dead at the next pin bump. Models essentially never call `graph_memory` unprompted, so "the tool still exists" doesn't preserve the product behavior.

The measured pain is real, but it's separable from the existence of the subsystem:

- The 10–30s pre-turn barrier is specifically `graphExtractionWG.Wait()` inside `injectGraphMemoryContext` — recall synchronously joining in-flight extraction LLM calls. That join can go without removing injection (recall reads a slightly stale graph; extractions land for the next turn).
- The extraction cost ($4.53 / 520 calls / ~66 min on the bench sweep) is already opt-out-able **today** via config: `graph_memory.enable_extraction: false` (`config_loader.go` pointer-bool, defaults true). The bench agent configs can pin explicit-only behavior without loom losing the mode.

## Requested shape

1. **Keep the good part**: drop the `graphExtractionWG.Wait()` from `injectGraphMemoryContext` (keep `FlushGraphMemoryExtraction` for shutdown/tests).
2. **Reinstate** the `NewAgent` extraction-init block and the three removed call sites (user-message fire, tool-cadence, injection call in `runConversationLoop`), gated as before by `EnableExtraction`.
3. **Add the missing knob**: `bool enable_recall_injection` on `loom.v1.GraphMemoryConfig` (proto first, buf regen), defaulted `true` in `DefaultGraphMemoryConfig()` and as a pointer-bool in `GraphMemoryConfigYAML` — the exact pattern `enable_extraction` already uses. Gate `injectGraphMemoryContext` on it alongside `Enabled`.
4. **Bench configs** set `enable_extraction: false` + `enable_recall_injection: false` to keep the published numbers reproducible — config, not code deletion.
5. **Tests**: revert the three `builtin_tool_suppression_test.go` assertions to expect the extractor enabled under default config, and add the mode matrix: default (both on), extraction-off, injection-off, explicit-tool-only (tool registered, no automatics). Plus one test pinning that injection no longer joins extraction.
6. This also brings the currently-dead surface back to life: `extractGraphMemoryAsync`, the cadence fields, and every recall knob (`ContextBudgetPercent`/`MaxContextTokens`/`MinSalienceThreshold`/`MaxRecallCandidates`/`BoostAmount`) — otherwise this PR ships exactly the kind of "parsed field that drives nothing" it documents `minL1Messages` as.

Known remaining sync cost when injection is enabled: the `extractSearchQuery` LLM distillation call before turn 1. That's pre-existing behavior, now config-gated; optimizing it (heuristic/embedding search) can be a separate follow-up.

Everything else in the PR: 👍

ilsiepotamus:
## Comprehensive review — the rest of the diff

This is the fuller pass promised in my first review. Findings below were **verified first-hand** (not just static analysis); I've marked confidence explicitly. Two areas are deliberately **excluded** because they already have dedicated review threads: `task_list` duplication (→ absorbed in #330) and explicit-only graph memory (→ my separate CHANGES_REQUESTED with the requested config-mode shape). Please don't re-litigate those here.

### Blocking correctness (verified)

1. **The PR's own rung-0 fold test fails as shipped.** `LOOM_CONTEXT_OPTIMISER=1 go test -tags fts5 -run TestRung0 ./test/context-optimiser/` → `FAIL: TestRung0_FoldLastResort — fold(t) never fired`. The commit deleted the compressor-less §5.4.5 fold fallback, so an embedded agent built without a prompt registry (compressor only wired when `a.prompts != nil`) has no working fold valve — context fills with unevictable text and the session wedges on `context_exhausted`. **I ran this; it's red on the branch.** (`pkg/agent/context_compilation.go:932`)

2. **The four new builtin tools are never registered, and the prompts point at names that don't exist.** `builtin.All()` still returns only the original ten tools; `NewExecuteQueryTool/NewTableProfileTool/NewEditFilesTool/NewDbtBuildTool` are constructed nowhere outside their own files/tests. Meanwhile `prompts/tools/file.yaml:42` says "use edit_files" and `prompts/tools/shell.yaml:12` says "sql_query for data probes" — and **no tool is named `sql_query`** (it's `execute_query`). So agents get live instructions to call tools that aren't in their toolset. ~1.4k lines of new SQL/dbt tooling is currently dead code. (`pkg/shuttle/builtin/registry.go:20`)

3. **Bedrock Converse breaks numeric tool arguments** — and this is a regression in the same commit that fixed the empty-input bug. `UnmarshalSmithyDocument` decodes JSON numbers as `smithy document.Number` (confirmed in smithy-go document.go:76: "for arbitrary-precision numbers"), but every downstream tool asserts `params["x"].(float64)` (`execute_query.go:118` row_limit, `task_list_tool.go:158/176` id, etc.). String/object args are now correctly delivered (a real fix over the old `"{}"`), but **numeric args fail the type assert** for every non-Anthropic Bedrock model now routed through Converse. (`pkg/llm/bedrock/converse.go:170`)

4. **`file_write` default mode silently changed `create` → `overwrite`.** develop: `WithDefault("create")`, `mode := "create"`, FILE_EXISTS guard. This branch: `WithDefault("overwrite")`, `mode := "overwrite"`. Every existing caller passing `{path, content}` without `mode` now clobbers instead of failing safe — e.g. two workflow stages writing the same path silently lose the first artifact. (`pkg/shuttle/builtin/file_write.go:167`)

5. **`task_list` mutates `Session.Tasks` without a lock** (compounds the #330 discussion, but it's a race regardless of that outcome). `if sess.Tasks == nil { sess.Tasks = types.NewTaskList() }` is an unsynchronized check-then-set, and `agent.go` renders `session.Tasks` — while `Messages` on the same struct is lock-guarded. Concurrent same-session Weaves, or coordinator background workers racing a user turn, will trip `-race` and can install two TaskLists losing one. Zero-tolerance-for-races repo. (`pkg/agent/task_list_tool.go:137`)

### Additionally flagged (from the review pass; NOT independently verified — author should confirm)

- **Thinking not wired past anthropic/bedrock**: `factory.go:262` — openai/litellm clients and `cmd_serve.go` never set `ThinkingLevel`, so the new gateway-thinking machinery is unreachable via `looms serve`.
- **Thinking API constraints**: non-adaptive path may send `budget_tokens ≥ max_tokens` and `temperature≠1` (Anthropic 400s), and the openai client sends `{type:adaptive}` to all Claude models incl. pre-4.6 that reject it. (`anthropic/client.go:259`)
- **Thinking signature corruption** on openai reassembly — relocated reasoning crammed into the first signature block; replayed blocks risk 400 invalid-signature mid-tool-use. (`openai/client.go:376,968`)
- **Pricing fallback duplicated 4×** with the copies already diverging (openai copy has no fable case; bedrock switches lack a fable arm) → Fable mis-billed ~4-5x or $0. The exact incident class these changes cite as motivation. (`openai/client.go:635`)
- **duckdb `GetMetadata` uses Go `%q`** for SQL identifiers → breaks qualified names in the new multi-attach mode and is injectable. (`fabric/factory/duckdb.go:195`)
- **shell output hard-capped at 12KB** (4KB head + 8KB tail) while the schema still advertises `max_output_bytes` up to 1MB, and the offloaded copy is pre-truncated so the elided middle is unrecoverable. (`shell_execute.go:548`)
- **`previewMeta` full-JSON-parses every stub on every render** under the memory write lock — unbounded per-call parsing replacing develop's 160-byte scan. (`context_compilation.go:332`)

### Benchmark (LongMemEval, Opus 4.6 agent + judge, ingest, 30q/type)

| Type | PASS | FAIL | Avg |
|---|---|---|---|
| temporal-reasoning | 29/30 (97%) | 0 | 87.8 |
| knowledge-update | 29/30 (97%) | 0 | 85.2 |
| single-session-user | 29/30 (97%) | 0 | 90.2 |
| single-session-assistant | 30/30 (100%) | 0 | 91.1 |
| **multi-session** | **25/30 (83%)** | **5** | **79.2** |
| single-session-preference | *(running)* | | |

Genuinely strong across the board — **except multi-session, the one category that leans hardest on cross-session recall, which has the only hard failures in the whole sweep (5).** That's empirical weight behind the graph-memory review: with automatic recall injection removed, answers that require pulling facts from earlier sessions are exactly where this regresses. I'll attach the five failed question IDs once the sweep finishes.

Net: the LLM-layer *intent* is right and the benchmark numbers are good, but items 1–4 are shipping bugs (one a red test, one a live tool-not-found, one a numeric-args regression, one a silent-clobber default) that should land before merge.

ilsiepotamus:
Adding one required item to the graph-memory thread (the CHANGES_REQUESTED on the explicit-only commit): **the docs now contradict the code.**

`docs/architecture/memory-systems.md:1639` still states:

> **Integration**: Graph memory context is injected as a system message at each conversation turn, alongside the segmented memory layers.

That is the exact behavior this PR removed. The PR *does* edit `memory-systems.md` (8 lines) — but all of it is compression-profile threshold text; the graph-memory section was left asserting the old auto-injection behavior. A reader is now told the opposite of what the code does.

This is the same honesty failure the PR itself calls out elsewhere (crediting the `minL1Messages` → "vestigial parsed field that drives nothing" doc fix). Per the repo's documentation standard it needs to hold for the headline change too.

**Required as part of the graph-memory rework:** whatever shape that commit lands in (I've asked for automatic-default + an `enable_recall_injection` config mode), reconcile memory-systems.md §Integration and the related passages in `docs/guides/memory-management.md` / `docs/reference/agent-configuration.md` so they describe the actual injection behavior and the new knob — rather than a per-turn injection that no longer happens by default.

(Related but tracked under the correctness thread, not here: the new tools and `ThinkingLevel` are undocumented, which is fine only because they're currently unregistered/unwired — documenting them as working would be a second contradiction.)

ilsiepotamus:
## Benchmark complete — all six types

| Type | PASS | PARTIAL | FAIL | Avg |
|---|---|---|---|---|
| single-session-assistant | 30/30 (100%) | 0 | 0 | 91.1 |
| temporal-reasoning | 29/30 (97%) | 0 | 0 | 87.8 |
| knowledge-update | 29/30 (97%) | 1 | 0 | 85.2 |
| single-session-user | 29/30 (97%) | 1 | 0 | 90.2 |
| **multi-session** | **25/30 (83%)** | 0 | **5** | 79.2 |
| **single-session-preference** | **13/30 (43%)** | **16** | 1 | 76.7 |

Two soft spots, and they are **not the same failure mode** — worth separating before drawing conclusions:

**multi-session (5 hard FAILs) — recall-shaped, corroborates the graph-memory review.** All five failures are cross-session *aggregation* questions, and all five undercount:

- `0a995998` (45): "pick up 2 items" — missed shopping-list entries across sessions
- `6d550036` (45): "led 1 project" — undercounted projects
- `46a3abf7` (37.5): "2 tanks" — undercounted
- `88432d0a` (45): baking instances across two sessions — incomplete
- `7024f17c` (41.25): yoga — partial recall

These are the purest test of *complete* cross-session retrieval, and the pattern is consistent: the model answers from what's in-context and undercounts what isn't. That's exactly what removing automatic recall injection would produce — direct evidence for the config-mode ask on the graph-memory commit.

**single-session-preference (43% PASS, but 16 PARTIAL / only 1 FAIL) — NOT a recall story, and I'm not attributing it to this PR.** Preference is a *single-session* category — it doesn't exercise cross-session memory at all, so it's orthogonal to the graph-memory change. The PARTIAL-dominated shape ("preference partly respected" rather than "forgotten") more likely reflects judge-rubric strictness for preference adherence. **This needs a baseline run on the same harness before it means anything** — I don't have a per-type preference baseline to compare against, so I'm flagging it as inconclusive, not as a regression.

Net: the multi-session result is real signal for the graph-memory thread; the preference number is a "needs a baseline" asterisk, not a finding.

ilsiepotamus:
## Correction to my benchmark analysis

I now have the per-type main-lineage baseline (same harness, oracle, 30/type) I said earlier I was missing. It changes one of my conclusions, so — correcting on the record:

| Type | Baseline (main, auto graph memory) | PR #329 (explicit-only) | Δ |
|---|---|---|---|
| temporal-reasoning | 90% | 97% | +7 |
| knowledge-update | 97% | 97% | 0 |
| multi-session | **83%** | **83%** | **0** |
| single-session-assistant | 100% | 100% | 0 |
| single-session-preference | 50% | 43% | −7 (within noise) |
| single-session-user | 97% | 97% | 0 |

**Retracting my earlier claim that the 5 multi-session failures corroborate the graph-memory change.** They don't: the baseline *with* automatic graph memory scored the identical 25/30 on multi-session, so explicit-only did not regress it — those failures are pre-existing, not introduced by this PR.

And the reason it can't be evidence either way: these runs are `--mode ingest`, which replays all sessions into the context window, so the facts are in-context and the graph-memory *recall* path is barely exercised. That masks the exact capability the graph-memory review concerns — it's the wrong instrument for that question.

Net change to the review:
- The graph-memory CHANGES_REQUESTED **stands on its own** (downstream dependency on automatic capture+recall; delete-vs-gate; automatic-stays-default) — it never rested on these numbers.
- The benchmark's actual verdict on this PR: **temporal +7, everything else flat, zero regressions** — the LLM-layer work measures cleanly. That's a point in the PR's favor, and I'd previously buried it under a multi-session concern that wasn't real.

The five correctness findings and the task_list/doc threads are unaffected.

ilsiepotamus:
## Official metric — the canonical comparison

Ran the LongMemEval **official autoeval** (`evaluate_qa.py`'s verbatim anscheck prompt) over all 180 PR-329 hypotheses, judged by **Azure gpt-5.2** — the same GPT-5-class judge behind the team's tracked "Official metric" column. This supersedes the per-type reading in my earlier benchmark comments, which used Loom's internal judge (the harsher "Internal PASS" methodology) and mismatched columns.

| Type (oracle, 30 each) | Baseline (Official) | PR #329 (Official) | Δ |
|---|---|---|---|
| temporal-reasoning | 28/30 (93%) | 29/30 (97%) | +1 |
| knowledge-update | 29/30 (97%) | 29/30 (97%) | 0 |
| multi-session | 25/30 (83%) | 23/30 (77%) | −2 |
| single-session-assistant | 30/30 (100%) | 30/30 (100%) | 0 |
| single-session-preference | 23/30 (77%) | 24/30 (80%) | +1 |
| single-session-user | 28/30 (93%) | 29/30 (97%) | +1 |
| **Aggregate** | **163/180 (90.6%)** | **164/180 (91.1%)** | **+1** |

**Read:** on the canonical metric PR #329 is a wash-to-marginally-ahead — +1 question overall, three types up 1, one down 2, two flat. All within single-run noise at n=30; no type shows a clear regression.

Two corrections to my earlier posts, both from finally having the official judge:
- **Preference is fine.** My internal-judge run showed a scary 43% PASS; the official metric puts it at **80% (24/30), +1 vs baseline**. Exactly the "internal-judge strictness, needs a baseline, don't attribute to the PR" caveat I flagged — confirmed. The 43% was a judge-rubric artifact, not a regression.
- **Multi-session is the one soft spot on the official metric (−2)**, but at 2 questions it's noise, and I'm explicitly *not* re-raising the graph-memory corroboration claim I retracted — these ingest-mode runs still don't exercise the recall path either way.

Net: the benchmark is not a reason to hold this PR. The correctness findings (fold test red, unregistered tools, converse numeric args, file_write default) are the substantive blockers; the graph-memory and task_list threads stand on their own arguments.

ilsiepotamus (changes requested):
This is a large, ambitious branch (32 commits, 81 files, ~7,300 lines) bundling LLM cost-correctness fixes, a batch-first SQL tool surface, a real Bedrock decode bug fix, and a default-on removal of automatic graph-memory extraction. Standalone it builds clean and its race tests pass — the code itself isn't broken. But there are four things to resolve before this is mergeable:

**1. Stale — does not merge cleanly today.** The merge-base is 151 commits behind current `main` (which now includes the MCP 2026-07-28 migration, the HITL event-gate, and #330's task_board absorbing checklist semantics). `git merge-tree` against current `main` shows 19 genuine content conflicts: 4 in `pkg/agent/agent.go`, 1 in `pkg/agent/context_compilation.go`, 8 in `pkg/llm/openai/client.go`, and 3 across the Bedrock client/converse files. This needs a real rebase, not a fast-forward.

**2. `task_list` duplicates `task_board`.** This branch adds a new `task_list_tool.go`, registered unconditionally at init, while `task_board_tool.go` (from #141/#330, already on main) is registered separately via `checkAndRegisterTaskBoardTool()`. That's two parallel task-tracking tool surfaces exposed to the model at the same time. Please fold `task_list` into `task_board` or explain why both need to coexist — "reuse existing functions before writing new ones" is a hard rule here, not a suggestion.

**3. Cost-calc slice duplicates #299.** #329's `openai/client.go` already contains the `x-litellm-response-cost`-header-authoritative logic and the `claude-opus-4-1`/`claude-opus` substring pricing fix — the same fix #299 makes independently. #299's own PR body cites this branch's HEAD commit directly, and reads like a deliberate, cleaner extraction of that slice. I'm approving #299 as the landing path for that fix — please drop the overlapping cost-calc changes from this branch during the rebase rather than carrying both.

**4. Graph-memory-explicit-only needs an explicit decision, not just a correctness pass.** Removing automatic extraction/recall injection is a legitimate, deliberate change to a feature that's currently default-on elsewhere in the codebase — real product behavior, not a bug fix. The code itself is clean, but bundling a default-behavior flip into a 7,000-line PR means it doesn't get its own decision. Please split this into its own PR so it can be reviewed and approved on its own terms.

The Bedrock Converse fix (`UnmarshalSmithyDocument` — smithy documents never `json.Marshal`, so tool args previously always decoded as `{}`) looks solid and could land on its own once split out. Docs read as accurate with no marketing language.

Requesting changes: rebase onto current main, resolve the task_list/task_board duplication, drop the cost-calc slice in favor of #299, and split the graph-memory default-off change into its own PR.

PR #330, opened 2026-08-18
feat(agent): task_board absorbs checklist semantics — write-once criteria, batch transitions

ilsiepotamus:
## Summary

Companion to the #329 review: rather than a second checklist tool (`task_list`) surfacing alongside `task_board`, this absorbs the genuinely new behaviors into `task_board`.

Discovered while reviewing #329: both of `task_list`'s signature concepts already exist in the task system — `Task.acceptance_criteria` (field 6) and `OutputPolicy.output_schema` (a validated JSON-Schema superset of a freeform output contract, with LLM-evaluated criteria, validator agents, and retry policy). What the task system genuinely lacked:

- **Write-once acceptance criteria via the tool**: `acceptance_criteria` is now settable at `create` (or on `update` while still empty), then immutable — changing non-empty criteria is refused with `ACCEPTANCE_CRITERIA_LOCKED` and a cancel-and-re-create instruction. Criteria define "done"; moving goalposts mid-task defeats them. Identical re-sends stay idempotent for safe retries.
- **Batch transitions**: `update` accepts an `updates` array applied independently with per-entry results — finishing the current task and claiming the next is one call, and one bad entry can't roll back its siblings.

The remaining two `task_list` behaviors — session-scoped ephemeral boards and the budgeted push-rendered checklist block — are proposed as follow-ups on #329: the transient-tail renderer is a good idea that belongs *over board state*, not attached to a parallel tool with its own divergent state.

## Test plan
- New: write-once lifecycle (set at create; set-while-empty; locked-after-set with dedicated error code; idempotent identical re-send) and batch semantics (independent application, partial-failure isolation) — table-tested against the real migrated SQLite task store
- `go test -tags fts5 -race ./pkg/agent/ ./pkg/task/` green


ilsiepotamus:
Fresh full review (independent reviewer + my verification, key regressions re-run) — the write-once single-writer semantics and session attribution were solid, but the review empirically proved a **blocker**: the advertised batch workflow (`update status=done` to finish, `update status=in_progress` to claim) bypassed `CloseTask`/`ClaimTask` entirely — dependents stayed BLOCKED forever (A blocks B; update A→done; B never becomes ready), claims never released, and two agents could pseudo-claim the same task past the WIP limits. All fixed in `c4f55bc3`:

- **Lifecycle routing**: status transitions in `update` now route through the real primitives — `done`→CloseTask, `cancelled`→new `Manager.CancelTask` (composes existing primitives: terminal write, history, event, dependent-unblock, parent auto-complete — no completion memory), `in_progress`→ClaimTask (atomic + WIP limits, and ownership-aware: "already in progress" succeeds only for the claim holder — a naive no-op let a losing session believe it owned the task, caught by the race test), `open` on a claimed task→ReleaseTask. The locked-criteria recovery text now describes a real cancel path instead of instructing the model to mark mistakes DONE (which cascaded parent auto-completion).
- **Write-once is now store-enforced**, not check-then-act: `TaskStore.SetAcceptanceCriteria` does one guarded UPDATE on both stores; concurrent writers — exactly one wins, the loser gets `ACCEPTANCE_CRITERIA_LOCKED`. ⚠️ **Interface note for out-of-repo implementers** (loom-cloud): `TaskStore` gains this one method.
- **Renderer**: session-scoped server-side (`ListTasksOpts` gains SessionID/Statuses/NewestFirst on both stores; this build's SQLCipher lacks JSON1, so sqlite uses an anchored LIKE-escaped match on the exact JSON fragment), live statuses only, newest-first, truncation note when the window clips. It stays **unwired** by design — doc comments now say so explicitly (the #329 follow-up consumes it).
- Minors: real error codes restored (INVALID_PARAMETER/NOT_FOUND), all-fail batches report `Success=false` with structured per-entry errors, non-array/empty `updates` rejected, batch capped at 20, budget writer stops at the truncation marker and counts the header, task titles/criteria newline-escaped so one task can't fabricate checklist lines, and the dead-conditional test (claim on an IN_PROGRESS-created task) now actually exercises claimed membership.

Gates: build, `-race` across pkg/agent + pkg/task + both stores (race tests also `-count=20`), gofmt, golangci-lint — all green.

pkg/storage/sqlite/task_store.go
```diff
@@ -273,6 +273,41 @@ func (s *TaskStore) UpdateTask(ctx context.Context, t *task.Task, _ []string) (*
 	return t, nil
 }
 
+// SetAcceptanceCriteria atomically sets a task's write-once acceptance
+// criteria. The guard lives in the UPDATE itself (no read-then-write
+// window): the write succeeds only while the stored criteria are empty or
+// already equal to the given value (idempotent retries). When different
+// non-empty criteria exist, an error wrapping task.ErrAcceptanceCriteriaLocked
+// is returned.
+func (s *TaskStore) SetAcceptanceCriteria(ctx context.Context, taskID, criteria string) (*task.Task, error) {
```
Ed-Jaramillo, line 282: The write-once guard here is genuinely atomic for two concurrent criteria writers — nice. One gap worth a follow-up, though: `UpdateTask` above ignores its `fields` parameter and overwrites every column, including `acceptance_criteria`. So a field-only `task_board` update (e.g. appending notes) that read the task *before* a concurrent criteria write will write the stale empty value back — silently erasing the criteria and re-opening the write-once lock, with no error anywhere.

This is the pre-existing full-row lost-update pattern, not something this PR introduced, so I don't think it blocks. But since `SetAcceptanceCriteria` is now the dedicated guarded writer, could we drop `acceptance_criteria` from `UpdateTask`'s SET list (or make `UpdateTask` honor `fields`)? All current callers write back the value they read, so the change should be safe. A test racing `SetAcceptanceCriteria` against a field-only `UpdateTask` would pin it. Same applies to the postgres store.

pkg/agent/task_board_tool.go
```diff
@@ -304,47 +319,285 @@ func (t *TaskBoardTool) executeClaim(ctx context.Context, input map[string]inter
...
+			}
+			return nil, fmt.Errorf("task %s is already in progress (claimed by session %q)",
+				existing.ID, existing.ClaimedBySession)
+		}
+		return t.manager.ClaimTask(ctx, existing.ID, t.agentID, sid)
+	case loomv1.TaskStatus_TASK_STATUS_OPEN:
+		if existing.ClaimedBySession != "" {
+			// Returning a claimed task to OPEN is a release: the claim
+			// fields must be cleared or the task can never be re-claimed.
+			return t.manager.ReleaseTask(ctx, existing.ID, existing.ClaimedBySession)
```
Ed-Jaramillo, line 595: Question on intent: this passes the *holder's* `existing.ClaimedBySession` to `ReleaseTask`, so the store's session-conditional release always matches regardless of who called — any session can kick the holder off via `update status=open` and then claim the task. That contrasts with the deliberately strict ownership check on `in_progress` re-assertion just above.

There's precedent (`executeClose` closes anyone's task, and the guide documents open-as-release), and the release is recorded in history, so this may well be the intended trade-off. If it isn't, restricting release-via-open to the claim-holding session (with a dedicated error otherwise) would make the two paths consistent.

pkg/task/manager.go
```diff
@@ -281,6 +281,67 @@ func (m *Manager) CloseTask(ctx context.Context, taskID, reason string) (*Task,
...
+
+	m.recordHistory(ctx, taskID, "cancelled", oldStatus, StatusName(cancelled.Status), oldAssignee, oldSession)
+	m.publishEvent(ctx, TopicTaskUpdated, cancelled, "")
+
+	// CANCELLED is terminal: tasks blocked on this one can proceed.
+	m.tryUnblockDependents(ctx, taskID)
+
+	// Settle the parent if this was the last non-terminal child.
+	if cancelled.ParentID != "" {
+		m.tryAutoCompleteParent(ctx, cancelled.ParentID)
```
Ed-Jaramillo, line 323: Cancelling the last non-terminal child now triggers `tryAutoCompleteParent`, which treats any terminal status as settling and closes the parent as **DONE** — with the reason "auto-completed: all N child tasks are done" and a completion memory — even when every child was cancelled. That records an accomplishment that didn't happen.

The helper predates this PR; what's new is that a cancel can reach it. Worth a follow-up to settle parents as CANCELLED when no child finished DONE?

pkg/storage/postgres/task_store.go
```diff
@@ -275,6 +275,52 @@ func (s *TaskStore) UpdateTask(ctx context.Context, t *task.Task, _ []string) (*
 	return t, nil
 }
 
+// SetAcceptanceCriteria atomically sets a task's write-once acceptance
+// criteria. The guard lives in the UPDATE itself: the write succeeds only
+// while the stored criteria are empty or already equal to the given value
+// (idempotent retries). When different non-empty criteria exist, an error
+// wrapping task.ErrAcceptanceCriteriaLocked is returned.
+func (s *TaskStore) SetAcceptanceCriteria(ctx context.Context, taskID, criteria string) (*task.Task, error) {
```
Ed-Jaramillo, line 283: The postgres side of the new store surface (this method plus the `SessionID`/`Statuses`/`NewestFirst` filters in `ListTasks`) has no test coverage — there's no postgres task_store test file at all, so these paths are compile-checked only. The SQL reads correct to me (JSONB `->>` on the JSONB column, numbered-parameter reuse is valid, int32 params match the INTEGER status column), but a pg integration test mirroring `task_store_criteria_test.go` would close the gap. Fine as a follow-up.

Ed-Jaramillo (approved):
Reviewed at c4f55bc. Approving with non-blocking comments.

This is a solid change: routing `update` status changes through the real lifecycle operations (CloseTask / CancelTask / ClaimTask / ReleaseTask) fixes a genuine pre-existing defect — the old path wrote status raw and skipped claim release, `closed_at`, and dependent unblocking — and the PR proves it with targeted regression tests rather than asserting it. The write-once guard as a single conditional UPDATE is the right mechanism for the TOCTOU concern, and the concurrency claims are backed by racing-goroutine tests against the real migrated store.

Verification I ran locally at head: `go test -tags fts5 -race -count=1` on `./pkg/storage/sqlite/`, `./pkg/task/`, and `./pkg/agent/` — all green except one pre-existing Windows-only TempDir flake (`TestSkillsCutover_AC6...`) that fails identically on main, so it's not from this PR. CI is fully green.

Four non-blocking observations are inline. Two smaller notes not worth an inline thread:

- `Manager.CancelTask` is a read-modify-write full-row update, unlike store `CloseTask`'s single conditional UPDATE — a cancel racing another writer can lose that writer's update, and cancelling an already-DONE task silently flips it to CANCELLED. This matches the existing looseness of `CloseTask`, so it's consistent rather than regressive, but a store-level conditional `CancelTask` would be stronger.
- In `applyTaskUpdate`, field changes persist before the status transition runs, so an entry combining `notes` + `status` whose transition fails will re-append the notes on retry. Low stakes, just worth knowing for retry guidance.

None of these block merge.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

PR #331, opened 2026-08-18
feat(longmemeval): generation-free conversation-replay mode

ilsiepotamus:
## What

Adds a `conversation` run mode to the LongMemEval harness that replays each haystack **turn-by-turn into one continuous agent session**, instead of feeding whole sessions as single blob messages (as `multi-session`/`ingest` do).

User turns go in as user messages; the dataset's **original assistant turns are recorded verbatim (generation-free)**, never regenerated. This preserves the benchmark's ground truth for the 51 `single-session-assistant` questions (whose gold answers quote the original assistant) while driving the memory pipeline the way a live chat would.

## Why

The blob-per-session ingest modes structurally **bypass two of Loom's core long-conversation subsystems**:
- **Salience** (`pkg/memory/salience.go`) is access/recency-driven (`base·decayRate^days` + boosts on touch). A whole session collapsed into one extraction event gives the graph almost no salience signal; turn-by-turn replay accumulates it as a real conversation would.
- **Compression** (`SegmentedMemory.ReleasePressure`, fired at context compilation) never triggers when each session is a one-shot exchange. A continuous replayed thread lets it fire as the window fills.

So today's `ms-on`/`ms-off` numbers measure *extraction + retrieval over a static transcript*, not memory under realistic accumulation. This mode closes that gap. (Context: this is the honest apples-to-apples upgrade — it is **not** generative replay, which would let the agent invent the assistant's turns and silently break the gold answers for the ~10% assistant-evidence questions.)

## How (proto-first)

- **`WeaveRequest.replay_assistant_message`** (field 12): when set, the agent records the text verbatim as the assistant turn and **skips the LLM call**, but still runs context compilation, compression, graph extraction, and salience. Gated by the same `server.allow_time_override` switch as `occurred_at` (FAILED_PRECONDITION when disabled); typically paired with `occurred_at` to anchor each turn.
- **Agent**: `WithScriptedResponse` threads the turn through context; the conversation loop substitutes it for the provider call *after* relief/compression has already run, so the no-tool-calls path finalizes the turn. Only generation is skipped.
- **Harness**: `ModeConversation` walks each session's alternating turns as `(user, assistant)` pairs and replays them generation-free, then asks the question in the same thread. A leading orphan assistant turn is skipped; a trailing bare user turn generates (never inventing assistant content).

## Tests

- `pkg/server/replay_test.go`: gate disabled-by-default, generation-free (no LLM call, scripted text persisted verbatim), `occurred_at` co-anchoring, empty-override fall-through, single-agent mirror.
- `cmd/loom-longmemeval/conversation_test.go`: turn pairing and anomaly handling via a recording client.
- `pkg/server` and `pkg/agent` full suites pass with `-race`; `buf lint`/`buf breaking` clean (field addition is non-breaking); `gofmt` + `golangci-lint` clean.

## Notes

- Stacked on #320 (`feat/longmemeval-occurred-at`) — needs its `occurred_at` plumbing. Merge after #320.
- No behavior change for live traffic: the field is unset on normal requests and gated off by default.


ilsiepotamus:
Fresh full review as a delta on #320: **no findings** — this is a clean, minimal design. The scripted-turn substitution mirrors the `occurred_at` context-key pattern exactly, rides the same `allow_time_override` gate (FAILED_PRECONDITION when disabled), preserves the full memory pipeline (relief/compression, compilation, extraction, salience) and skips only generation; proto field 12's comment says precisely what the code does. Since stacked PRs get no CI, I ran the gates locally at `cc8e3fad`: `go build -tags fts5 ./...` and `-race` suites for pkg/server, pkg/agent, and cmd/loom-longmemeval — all green. Merge after #320 lands (GitHub will retarget automatically).

pkg/server/multi_agent.go
```diff
@@ -712,6 +712,12 @@ func (s *MultiAgentServer) Weave(ctx context.Context, req *loomv1.WeaveRequest)
 	if err != nil {
 		return nil, err
 	}
+	// Generation-free replay: a scripted assistant turn substitutes for the LLM
+	// call while the memory pipeline still runs (see applyReplayAssistant).
+	ctx, err = applyReplayAssistant(ctx, req, s.allowTimeOverride)
```
Ed-Jaramillo: **Gate gap (basis of the requested changes)**: this call correctly gates unary `Weave`, but `MultiAgentServer.StreamWeave` (~line 873, outside this diff) applies only `applyOccurredAt` — not `applyReplayAssistant`. The other three Weave entry points gate both, so on that one path `replay_assistant_message` is silently ignored:

- gate **off** (default): a streaming request carrying the field proceeds normally instead of returning `FAILED_PRECONDITION` — inconsistent contract across entry points;
- gate **on**: a streaming replay persists a *generated* assistant turn where the scripted one should be — silently inventing assistant content in a replayed conversation, with no error signal.

The #320 precedent covered all four entry points for `occurred_at`, so this looks like an oversight. Fix is the same two-line block used in `Server.StreamWeave`; a mirror of `TestWeave_ReplayAssistant_DisabledByDefault` for `MultiAgentServer.StreamWeave` would pin it.

pkg/server/occurred_at.go
```diff
@@ -48,3 +48,26 @@ func applyOccurredAt(ctx context.Context, req *loomv1.WeaveRequest, allow bool)
...
+// pipeline (context compilation, compression, extraction, salience). See
+// agent.WithScriptedResponse.
+//
+// Generation-free replay lets a caller supply the assistant's side of the
+// conversation, so — like occurred_at — it is opt-in per server and gated by
+// the same server.allow_time_override switch; requests carrying the field are
+// rejected with FAILED_PRECONDITION unless allow is true.
+func applyReplayAssistant(ctx context.Context, req *loomv1.WeaveRequest, allow bool) (context.Context, error) {
+	msg := req.GetReplayAssistantMessage()
+	if msg == "" {
```
Ed-Jaramillo, line 72: Tiny gap: the server gates on `msg != ""` here, but the loop substitutes only when `strings.TrimSpace(scripted) != ""` — so a whitespace-only `replay_assistant_message` on an enabled server is accepted as a replay request and then silently generates a real LLM turn, persisting invented content where the caller expected verbatim replay. Unlikely with LongMemEval data, but trimming here (and treating whitespace-only as empty or invalid) would align the two checks. Non-blocking.

proto/loom/v1/loom.proto
```diff
@@ -694,6 +694,22 @@ message WeaveRequest {
...
+  //
+  // This replays a pre-written conversation turn-by-turn while preserving the
+  // ground-truth assistant content (never inventing the assistant's side), so
+  // the memory system is exercised under realistic accumulation. It is used by
+  // the LongMemEval conversation-replay harness and is typically paired with
+  // occurred_at to anchor each turn at its historical time.
+  //
+  // Gated by the same server.allow_time_override switch as occurred_at
+  // (FAILED_PRECONDITION when disabled). Live conversations leave it unset.
+  string replay_assistant_message = 12;
```
Ed-Jaramillo, line 716: Reusing `server.allow_time_override` means operators who enabled it for #320's `occurred_at` timestamp anchoring have now also opted into caller-supplied assistant content — the flag's name no longer describes its scope. It's documented here and default-off, so no objection to shipping as-is, but is a rename (e.g. `allow_replay_overrides`) worth doing before the flag has external users? Cheaper now than after adoption. Non-blocking.

Ed-Jaramillo (changes requested):
## Review — generation-free conversation replay

Since no CI runs on this stacked branch (checks trigger for main-based PRs), I verified everything locally at head `cc8e3fa`:

- All 7 new tests pass with `-tags fts5 -race` (gate-off rejection with the agent provably not running, verbatim scripted persistence with zero LLM calls, `occurred_at` co-anchoring, empty-field fall-through, harness turn pairing/anomalies).
- `cmd/loom-longmemeval` fully green; `pkg/server` and `pkg/agent` failures have a **zero PR-vs-main delta** on my machine (known Windows-local baseline, all in unrelated files).
- `buf lint` + `buf breaking --against main` clean — field 12 is a clean additive change. `gofmt` clean.
- Code inspection confirms the scripted substitution sits after relief/compression and context compilation in `runConversationLoop`, so "only generation is skipped" is structurally true, not just claimed.

The design is sound and the test quality is genuinely good (the mock-returns-"pong" verbatim-persistence test actually distinguishes scripted from generated content; the security default is pinned on both server types).

**Requesting changes for one gap**: `MultiAgentServer.StreamWeave` applies `applyOccurredAt` but not `applyReplayAssistant`, so `replay_assistant_message` is silently dropped on that entry point — neither rejected when the gate is off (inconsistent with unary) nor applied when it's on (a streaming replay would persist a *generated* turn where the scripted one should be, which is exactly the ground-truth corruption this field exists to prevent). Two-line fix mirroring `Server.StreamWeave`, plus a mirrored gate test. Details inline.

Two non-blocking minors inline as well (whitespace-only replay text; `allow_time_override` scope/naming). Also noting: this branch is based on a stale #320 commit (`475bc7b`; #320 head is now `4d04f1d`), so a rebase — and this code's first-ever CI run — are still ahead after #320 merges.

🤖 Reviewed with the help of Claude Code (Fable 5)

ilsiepotamus:
Round-1 feedback addressed in `56114385`.

**Blocker — StreamWeave gate gap**: `MultiAgentServer.StreamWeave` now applies `applyReplayAssistant` immediately after `applyOccurredAt`, matching the other three entry points. Pinned by two new tests: `TestMultiAgentStreamWeave_ReplayAssistant_DisabledByDefault` (gate off → `FAILED_PRECONDITION`, agent never runs) and `TestMultiAgentStreamWeave_ReplayAssistant_UsesScriptedTurn` (gate on → scripted text persisted verbatim with zero LLM calls, proving the streaming path substitutes rather than generates).

**Whitespace-only replay**: the gate now rejects a whitespace-only `replay_assistant_message` with `INVALID_ARGUMENT` instead of accepting it and silently generating — aligned with the loop's non-blank substitution check. `TestWeave_ReplayAssistant_WhitespaceOnlyRejected` pins it.

**Flag scope/naming**: good catch, and it surfaced something better than a rename — `server.allow_time_override` already shipped in released v1.4.0 (it landed on main via #308), so renaming would break existing config. Instead, `replay_assistant_message` now has its own switch, `server.allow_assistant_override` (default off on both server types): operators who enabled timestamp anchoring have NOT implicitly accepted caller-supplied assistant content, which was the real exposure behind your question. `TestWeave_ReplayAssistant_TimeOverrideDoesNotUnlockReplay` pins the separation. Proto field-12 comment, config example, and the benchmark guide updated; conversation mode now documents needing both flags.

Validation: full `pkg/server`, `cmd/looms`, `cmd/loom-longmemeval` suites pass with `-tags fts5 -race`; `buf lint` + `buf generate` clean; `gofmt` clean.

Rebase onto #320's head deferred until #320 merges, per your note — that rebase also picks up #320's hardened FailedPrecondition-based abort matching, which covers the new flag's rejection message automatically.

ilsiepotamus:
Ping on this one — round-1 feedback was addressed back in `18255ae9` (StreamWeave gate gap, separate `server.allow_assistant_override` flag, whitespace-only rejection, all with tests). The branch has since been rebased onto `main` now that #320 has merged, so CI runs on it for the first time. Ready for a re-look whenever you have a moment.

ilsiepotamus:
Correction to my note just above: CI has already run on this branch and is fully green (all checks passed) — not still ahead. Nothing pending here beyond the re-review.

Josh-Schoen:
Acknowledging the problem this solves: `ingest`/`multi-session` blob-per-session replay structurally starves salience and compression of the accumulation signal they need, and generative replay would invent assistant content for the ~10% of questions whose gold answers quote the original assistant verbatim. A generation-free, turn-by-turn `conversation` mode is the right shape for that — it lets the memory pipeline run under realistic pressure without touching ground truth.

**Verdict: approve.**

**What this makes possible**: the LongMemEval harness can now measure memory behavior (salience accumulation, mid-conversation compression) under conditions that actually resemble a live chat, instead of only measuring extraction+retrieval over a static transcript. The turn-pairing logic handles the real messiness of the dataset (orphan leading assistant turns, trailing unanswered user turns) without crashing or inventing content.

**Where it could hurt**: nothing for a normal deployment — I traced every caller of `replay_assistant_message` end to end and it's set in exactly one place in the tree (`cmd/loom-longmemeval/runner.go`), gated by a config flag that defaults off and is separate from `allow_time_override`. The one thing worth flagging for the record: your first commit on this branch had `MultiAgentServer.StreamWeave` silently skip the gate entirely (neither rejecting nor honoring the field) — your second commit caught and fixed it with tests before this ever reached review. Good catch on your own pass, and worth calling out explicitly in the PR description as "found and fixed during self-review," since a future reader of just the final diff won't see that a StreamWeave-shaped RPC is exactly the kind of second entry point that's easy to forget when adding a gate like this.

**Verified by execution, not just reading**:
- `buf lint` / `buf breaking --against main` clean; `buf generate` against the real BSR plugins produces a byte-identical `gen/` tree (no hand-edited generated code).
- `go test -tags fts5 -race ./pkg/agent/...` and `./pkg/server/...` — both full suites pass, including all 12 replay-specific tests with real `=== RUN`/`--- PASS` lines (not a zero-match false "ok").
- `go test -tags fts5 -race -v ./cmd/loom-longmemeval/...` passes, including the turn-pairing and anomaly-handling tests.
- `gofmt`/`go vet` clean on every touched file.
- Confirmed all 4 `WeaveRequest`-accepting RPCs (`Server.Weave/StreamWeave`, `MultiAgentServer.Weave/StreamWeave`) call the gate identically — not just the ones the description calls out.

**Findings**:
1. (non-blocking, next-contributor impact) `pkg/server/replay_test.go` has a direct streaming end-to-end test for `MultiAgentServer.StreamWeave`'s gate but not for the single-agent `Server.StreamWeave`'s — same code shape, same risk class as the bug your second commit fixed on the multi-agent side. Worth a ~15-line mirror test for symmetry, since that's exactly the kind of entry point that got missed once already on this branch.
2. (informational, framework-consumer impact) `agent.WithScriptedResponse` is unrestricted at the `pkg/agent` layer, same as the existing `WithOccurredAt`. Not new, not this PR's scope — just noting the enforcement boundary is `pkg/server`, not `pkg/agent`, for anyone embedding the package directly.

**Things I was wrong about**: I went in expecting to find an unresolved gap where the scripted-response path was reachable without the gate somewhere in the shared serving code — that's exactly what the task brief flagged as the central risk. It was real, but only in this branch's first commit; you found and fixed it yourselves before I looked at it. I also expected to have to caveat whether `cmd/loom-longmemeval`'s tests actually run in CI — they do (`ci.yml`'s test job scope is `./pkg/... ./internal/... ./cmd/...` on this branch), so that's a non-issue.

**Environment note for future reference**: `buf generate`'s plugins here are all `remote:` (BSR-hosted), so a real regeneration check works without `protoc-gen-go-grpc`/`protoc-gen-grpc-gateway`/`protoc-gen-openapiv2` installed locally — only network access to the BSR is needed.

Josh-Schoen:
One process note, not a verdict change: @Ed-Jaramillo's `CHANGES_REQUESTED` review from 8/24 is formally still on record (GitHub doesn't auto-clear review state on a follow-up commit) even though its one blocker was fixed in `18255ae9` with tests, and the author pinged for re-review on 9/10. If branch protection here requires resolving outstanding reviews before merge, that re-review/dismissal is the remaining formal step — separate from the "approve" above, which was about the code itself.

Josh-Schoen (dismissed):
Approve. Traced every caller of the new replay_assistant_message field end to end — set in exactly one place in the tree, gated by a config flag that defaults off. Confirmed the normal (non-replay) serving path is byte-for-byte unchanged. buf lint/breaking clean, gen/ reproduces byte-for-byte, go test -race clean across pkg/agent and pkg/server plus all 12 replay-specific tests. Full findings in the comment above. Note: an earlier CHANGES_REQUESTED review on this PR (8/24) was substantively addressed by a later commit with tests, but is formally still on record pending that reviewer's re-review — separate from this review of the code itself.

cmd/loom-longmemeval/runner.go
```diff
@@ -390,6 +405,29 @@ func (r *Runner) weave(ctx context.Context, sessionID, query, agentID string, oc
...
+// pipeline. Anchors both rows at occurredAt. Requires
+// server.allow_assistant_override (replay) and server.allow_time_override
+// (anchoring) on the target server.
+func (r *Runner) replayTurn(ctx context.Context, sessionID, userContent, assistantContent, agentID string, occurredAt time.Time, result *EntryResult) error {
+	req := buildWeaveRequest(sessionID, userContent, agentID, occurredAt)
+	req.ReplayAssistantMessage = assistantContent
+
+	resp, err := r.client.Weave(ctx, req)
+	if err != nil {
+		return err
```
Ed-Jaramillo, line 441: **MAJOR-001** — On a server with `server.allow_assistant_override` left at its documented default `false`, this call returns `FAILED_PRECONDITION`. Unlike `weave`, `replayTurn` does not set `result.grpcCode`, and `Runner.Run` only aborts for `isTimeOverrideRejection`, which matches `allow_time_override` failures. Conversation mode can therefore write an all-error JSONL and exit 0, allowing automation to accept an invalid benchmark.

**Requested action:** Capture the replay error status (`status.Code(err)`) exactly as `weave` does, extend the fail-fast classification to recognize the assistant-override rejection with an actionable message, and add a regression test asserting non-zero exit on `FAILED_PRECONDITION` from replay.

*This review was performed by Copilot using GPT-5.6 and Claude Code Opus 4.6*

Ed-Jaramillo (changes requested):
## PR Review — feat(longmemeval): generation-free conversation-replay mode

**Reviewed SHA:** `18255ae94cbb5216fe7bd047de9b6e7ac4b00656`

---

### Re-review context

This is the re-review requested by @ilsiepotamus on 9/10 and flagged by @Josh-Schoen's [process note](https://github.com/teradata-labs/loom/pull/331#issuecomment-5621809966): the prior `CHANGES_REQUESTED` review from 8/24 is formally still on record.

**Status of the 8/24 findings — all three are resolved in `18255ae9`:**
- **StreamWeave gate gap** (blocker): `MultiAgentServer.StreamWeave` now applies `applyReplayAssistant` immediately after `applyOccurredAt`, matching the other three entry points. Pinned by two new tests. **Resolved.**
- **Whitespace-only replay handling**: the server now rejects whitespace-only `replay_assistant_message` with the same `FAILED_PRECONDITION` as empty. **Resolved.**
- **Flag reuse security boundary**: replay is no longer piggy-backed on `server.allow_time_override`. A separate `server.allow_assistant_override` flag (default `false`) gates it independently. **Resolved.**

The 8/24 blocker is cleared. However, this fresh re-review of the full `18255ae9` head surfaces one new finding (MAJOR-001 below) in the conversation-mode runner path that was not part of the original review scope.

---

### Positive Observations

- The separate assistant-content gate (`server.allow_assistant_override`) avoids implicitly granting caller-supplied assistant text to operators who only enabled historical timestamp anchoring — a deliberate and well-justified security boundary.
- The gate is wired through both unary and streaming paths for both single-agent and multi-agent servers, with focused tests including whitespace-only input.
- Scripted assistant substitution occurs after context compilation and relief/compression, preserving the intended memory pipeline while avoiding the normal response-generation call. This is exactly the right layer to cut.
- The PR updates the proto contract and generated artifacts together, and the exact reviewed head has successful proto, race-enabled unit-test, build, lint, fuzz, coverage, interoperability, and security-related CI checks.

### Review Limitations

- No local tests or builds were run against `18255ae9`. Exact-head CI was verified through the commit check-runs API, so the lack of local execution does not make the snapshot or the blocking finding indeterminate.
- The local repository's `origin` is a fork; the target repository was explicitly queried through `teradata-labs/loom` and exact review refs. No evidence from the newer `main` branch was mixed into the reviewed diff.
- No GitHub comments, reviews, labels, branches, commits, or other remote state were changed prior to this review.

---

### Merge Recommendation

**REQUEST CHANGES**

The reviewed head introduces the requested conversation-replay path and its separate assistant-content gate, but the benchmark runner can treat a server-side `allow_assistant_override` rejection as a successful run. It can write error-filled results and exit zero because `replayTurn` drops the gRPC status and the existing fail-fast check only recognizes timestamp-override failures. This can produce an invalid benchmark result that automation accepts, so the runner needs a targeted fix and regression test before merge.

### Summary

The base snapshot did not contain the `replay_assistant_message` contract or the LongMemEval conversation mode. The reviewed head adds proto field 12, generated Go/OpenAPI updates, a separately opt-in `server.allow_assistant_override` setting, single-agent and multi-agent unary/streaming enforcement, scripted assistant-turn substitution in the agent loop, and a runner that replays user/assistant pairs in one session.

The implementation persists the user turn and runs the normal context, relief/compression, extraction, and salience pipeline, then substitutes the supplied assistant text instead of making the normal response-generation call. The runner handles pairing and orphan/trailing turns and the question in the same session. These behaviors are supported by the added server and harness tests and by successful exact-head CI.

One head-only reliability defect remains: the new `replayTurn` path does not preserve the gRPC status needed by `Runner.Run`'s existing configuration-rejection abort logic. With the documented/default assistant gate disabled, conversation mode can therefore report completion despite every replay request failing.

### Evidence Reviewed

- Live PR metadata and final head recheck: title, author, branches, base SHA, reviewed head SHA, open PR state, `mergeable=true`, and `mergeable_state=behind`.
- Complete exact-head diff: 15 files changed, 775 insertions, 20 deletions; `git diff --check` exited 0.
- Changed source and contract files inspected at the reviewed SHA:
  - `proto/loom/v1/loom.proto`, `gen/go/loom/v1/loom.pb.go`, `gen/openapiv2/loom/v1/loom.swagger.json`
  - `pkg/server/occurred_at.go`, `pkg/server/server.go`, `pkg/server/multi_agent.go`, `pkg/server/replay_test.go`
  - `pkg/agent/agent.go`, `pkg/agent/scripted_response.go`
  - `cmd/loom-longmemeval/runner.go`, `cmd/loom-longmemeval/conversation_test.go`, `cmd/loom-longmemeval/main.go`
  - `cmd/looms/config.go`, `cmd/looms/cmd_serve.go`
  - `docs/guides/longmemeval-benchmark.md`
- Exact-head CI: all checks passed (`Unit Tests (1.26)`, `Build ubuntu/windows/macos-latest`, `Proto Lint & Breaking Changes`, `Go Lint`, `Coverage Requirements`, `Fuzz Tests`, `MCP SDK Interop`, `Security Scan`, `CodeQL`, `Check Version Drift`). `gosec` concluded `neutral` with skip behavior.

### Verification of PR Claims

| Claim | Status | Rationale |
|---|---|---|
| Add a generation-free LongMemEval conversation mode | ✅ VERIFIED | The mode replays turns in order and supplies scripted assistant content through the agent context. |
| Replay assistant content is separately opt-in from timestamp anchoring | ✅ VERIFIED | Uses `server.allow_assistant_override`, default false, documented separately. |
| All server entry points enforce the assistant replay gate | ✅ VERIFIED | Single-agent and multi-agent unary/streaming paths apply the validation helper; tests cover all paths. |
| Replayed assistant turns are persisted verbatim without normal response generation | ✅ VERIFIED | Substitution occurs after context compilation and relief/compression; assistant turn is still persisted. |
| Conversation pairing and malformed-input handling are covered | ✅ VERIFIED | Tests cover user/assistant pairing, leading orphan assistant turns, trailing user turns, and question generation. |
| CI passes for the reviewed head | ✅ VERIFIED | All named checks point to the immutable reviewed SHA. |
| Replay uses the existing `server.allow_time_override` gate alone | ❌ CONTRADICTED | The final head intentionally uses a separate assistant-content gate, not the time-override gate alone. |

### Blocking Findings

#### MAJOR-001 — Conversation benchmark can exit successfully after assistant replay rejection

- **Severity:** MAJOR
- **Confidence:** 9/10
- **Location:** `cmd/loom-longmemeval/runner.go:408-420`, with related handling at `runner.go:190-252` and `main.go:400-436`.
- **Observed behavior:** `replayTurn` calls `r.client.Weave` directly and returns an error without assigning `result.grpcCode`. `Runner.Run` only calls `isTimeOverrideRejection`, which recognizes `occurred_at`/`allow_time_override`, not `replay_assistant_message`/`allow_assistant_override`. The per-entry errors therefore do not cancel the run. `runBenchmark` writes the results and returns nil when `Runner.Run` returns nil, so a conversation run can produce error rows and process exit status 0.
- **Expected behavior:** A conversation run that cannot perform replay because `server.allow_assistant_override` is disabled should fail fast with an actionable non-zero result, just as the existing timestamp-override configuration failure does.
- **Trigger:** Run `--mode conversation` against a server with the documented default `server.allow_assistant_override: false`.
- **Minimal fix:** Make `replayTurn` record `status.Code(err)` exactly as `weave` does, and extend the runner's fail-fast classification to recognize the assistant override rejection with an actionable message.
- **Verification test:** Add a runner-level regression test with a fake `LoomServiceClient` that returns `status.Error(codes.FailedPrecondition, "...allow_assistant_override...")` for replay. Assert that `Runner.Run` returns an abort error and the benchmark command returns non-zero.

### Questions and Assumptions

- **Assumption:** `server.allow_assistant_override: false` is a supported default, not an environment-specific invariant. The default is explicitly set false in `config.go`, the example config comments it out as false, and the guide requires enabling it.
- **Question:** The PR description mentions 51 assistant-evidence questions, while the benchmark guide lists 56 `single-session-assistant` questions. The available evidence did not reconcile whether these refer to different filtered subsets.
- **CI note:** `gosec` is `neutral`/skipping while `Security Scan` and `CodeQL` are successful for the reviewed SHA.
- **Branch status:** GitHub reports `mergeable_state=behind` at the final snapshot.

---

*This review was performed by Copilot using GPT-5.6 and Claude Code Opus 4.6*

ilsiepotamus:
MAJOR-001 fixed in `15b1e466`. Thanks for catching it — a benchmark that writes error rows and exits zero is the worst failure mode this harness has, and your trace through `replayTurn` → `isTimeOverrideRejection` → `runBenchmark` was exactly right.

Two halves, either of which alone silently restores the bug:

- **`replayTurn` dropped the gRPC status**, unlike `weave` beside it, so a refusal was indistinguishable from a transient per-entry failure. It now records `status.Code(err)`.
- **The fail-fast only knew `occurred_at`.** `isTimeOverrideRejection` becomes `overrideRejection`, covering every override the run depends on and returning the message naming the flag the operator has to change — the assistant-override message also points at using a different `--mode`, since conversation mode is the only one needing it.

A refused override does not degrade the benchmark, it invalidates it: every remaining entry fails identically. `Runner.Run` already converts the abort into a returned error, so the nonzero exit follows once the rejection is *detected* — detection was the whole gap.

Tests: the rejection table now covers both overrides and asserts the abort message names the right flag; two new tests walk the whole path — the refusal a default server actually returns, through `replayTurn`, into the fail-fast check — plus the success case. I verified the regression test has teeth by reverting the `grpcCode` line (fails) and restoring it (passes).

The branch is also rebased onto current `main` now that #320 has merged. CI green (18 pass).

Ed-Jaramillo (approved):
## Re-review — `15b1e466`

**Reviewed SHA:** `15b1e4663b901492eb50a1a56d508d0d4eec4155` — one commit past `18255ae9`, touching only `cmd/loom-longmemeval/runner.go` and `runner_test.go`.

### MAJOR-001 — resolved

Verified at the exact head:

- `replayTurn` now records `status.Code(err)` into `result.grpcCode` exactly as `weave` does (`runner.go:440`), and the `EntryResult` flows by pointer through `runConversationWith` back into `Runner.Run`'s classification.
- `isTimeOverrideRejection` became `overrideRejection`, covering both overrides. The assistant branch matches `FailedPrecondition` + `replay_assistant_message`, with the bare `allow_assistant_override` substring as fallback; the server's actual rejection text contains both tokens, so both signals fire. The abort message names the flag and suggests using a different `--mode`.
- The abort error is not `context.Canceled`, so `runBenchmark` returns `benchmark run aborted: …` and the command exits non-zero — parity with the existing occurred_at rejection path, as requested.
- Regression coverage: `TestOverrideRejection` (both overrides, primary + fallback signals, negative cases), `TestReplayTurn_RecordsRejectionSoTheRunAborts` (the real server refusal walked through `replayTurn` into the fail-fast check), `TestReplayTurn_SuccessLeavesNoRejection`, and the pre-existing `TestRunAbortsOnTimeOverrideRejection` pinning `Run`'s abort machinery end-to-end.

### Local reproduction (scratchpad snapshot of the exact SHA)

- `go test -tags fts5 ./cmd/loom-longmemeval` — pass.
- `go test -tags fts5 -run 'Replay|Override|OccurredAt' ./pkg/server` — all 16 pass, including whitespace-only rejection, `TimeOverrideDoesNotUnlockReplay`, and multi-agent `StreamWeave` gating. The 8/24 findings remain intact; the head commit touched no server files.
- Mutation check: reverting only the `replayTurn` `grpcCode` line makes `TestReplayTurn_RecordsRejectionSoTheRunAborts` fail with the pinned message — the regression test has teeth.
- Exact-head CI: 18 checks successful (`All CI Checks Passed` included); `gosec` remains neutral/skipped, unchanged from the previously reviewed head.

### Non-blocking notes

- On abort, `runBenchmark` still writes the partial JSONL before exiting non-zero — identical to the pre-existing occurred_at behavior, and the non-zero exit is the guard automation relies on, so this is consistent rather than a defect.
- The new tests compose `Run`-level abort coverage (occurred_at) with classifier coverage (assistant override) rather than adding a second `Run`-level conversation-mode test; since both overrides share `overrideRejection`, the composition covers the requested scenario.

### Merge recommendation

**APPROVE.** MAJOR-001 is fixed with regression coverage at `15b1e466`. The branch is currently `CONFLICTING`/`DIRTY` against `main` and needs a rebase or merge-from-main before it can land; this approval covers the reviewed head.

*Verification performed with Claude Code (Fable 5): exact-SHA claim checks, local test reproduction, and mutation testing of the regression test.*

ilsiepotamus:
Rebased onto current \`main\` at \`6db8d82c\` (previous fix commit unchanged: \`15b1e466\`, the MAJOR-001 fix).

One small conflict in \`cmd/looms/config.go\`: main had added its own new viper default (\`server.allow_time_override\`-adjacent) in the same block as this branch's \`server.allow_assistant_override\` default — both kept, purely additive.

Verified on the pushed head: \`go build -tags fts5 ./...\`, \`go test -tags fts5 -race ./cmd/loom-longmemeval/... ./cmd/looms/... ./pkg/agent/... ./pkg/server/...\` (all pass), \`gofmt -l\` clean, \`golangci-lint run --build-tags fts5 ./...\` (0 issues), \`buf lint\`/\`buf format --diff\`/\`buf breaking --against main\` all clean (no drift after \`buf generate\`), CI green.

@Ed-Jaramillo — ready for another look whenever you have a chance.

cmd/loom-longmemeval/runner.go
```diff
@@ -390,6 +422,33 @@ func (r *Runner) weave(ctx context.Context, sessionID, query, agentID string, oc
...
+		result.grpcCode = status.Code(err)
+		return err
+	}
+
+	if resp.Cost != nil && resp.Cost.LlmCost != nil {
+		result.InputTokens += int(resp.Cost.LlmCost.InputTokens)
+		result.OutputTokens += int(resp.Cost.LlmCost.OutputTokens)
+	}
+
+	return nil
```
Ed-Jaramillo: **Non-blocking.** A successful `Weave` is treated as proof the turn was replayed, but a server that predates field 12 doesn't reject it. Old proto3 binaries ignore unknown fields, so the turn just generates.

Released v1.4.0 is exactly that server. It has `occurred_at`/`allow_time_override` but no `replay_assistant_message`, and its non-strict `viper.Unmarshal` silently drops an `allow_assistant_override: true` line. So `--mode conversation --server <v1.4.0 host>`:
- replaces every dataset assistant turn with a generated one;
- pays for a generation call per turn;
- exits 0 with a result that looks valid.

A server with this PR echoes the scripted text as `resp.Text` with zero output tokens, so the check is cheap:

```go
if resp.GetText() != assistantContent {
	// abort the run like a refused override: "server ignored
	// replay_assistant_message (it predates generation-free replay);
	// upgrade looms or use another --mode"
	return errReplayIgnored
}
```

This also needs:
- an abort branch next to `overrideRejection`;
- the `recordingClient` and `weaveOKClient` stubs echoing `in.ReplayAssistantMessage`. They currently answer replay requests with `"ok"` or an empty response, which this check would correctly reject.

A stub that returns generated text for a replay request would pin the behavior.

Ed-Jaramillo (approved):
## Re-review — `6db8d82c`

**Reviewed SHA:** `6db8d82cf0cef6f2c2710ccc5e20ca5466766c72`. This is the rebase of `15b1e466`, which I approved on 9/11.

**Rebase:** the patches match the approved series. `git range-diff` shows commits 1 and 3 as `=`. Commit 2 differs only in `cmd/looms/config.go` context lines, where main's `skip_embedded_agents`/`patterns_dir` defaults now sit next to the new `allow_assistant_override` default. The runner, `occurred_at.go`, `replay_test.go`, `scripted_response.go`, and the guide are byte-identical to `15b1e466`.

**On the new base:** about 118 main commits landed under this branch, including large changes to the agent loop and server handlers. Checked at the exact head:
- All four `WeaveRequest` entry points still gate before the agent runs. Main's session-ownership check still runs after the gate, so an enabled server still refuses replay into a session the caller can't access.
- The scripted substitution still comes after relief/compression and skips only the generation call. The output-token breaker and the empty-response retry can't fire on it.
- Background sub-agents and broadcast handlers run on `turnIdentityContext`, which carries only the tenant id, so the scripted text can't reach their turns.
- Exact-head CI is green:
  - the `-race` unit-test job passes every replay test and every MAJOR-001 test by name;
  - the proto lint, format, breaking-change, and generation-drift steps pass;
  - `gosec` is neutral, as on earlier heads.

**Non-blocking (inline on `runner.go`):** the harness can't tell when a server silently ignores `replay_assistant_message`, e.g. a v1.4.0 server. In that case the run quietly becomes generative replay and exits 0. I rate this minor: it needs a version mismatch (building looms and the harness from the same tree avoids it), and the fix is small. Fine to do here or as a follow-up.

**Nit:** the description's "How" bullet still says replay is gated by `server.allow_time_override`. Since `7fa638aa` it's `server.allow_assistant_override`. The description also cites 51 single-session-assistant questions, while the guide lists 56.

**Not verified locally:** this machine's Go module cache lacks the head's dependency versions, so my execution evidence is the exact-head CI logs. The inline scenario wasn't reproduced end to end.

**Approve.** The branch is one commit behind `main` (#412, which touches none of this PR's files).

*Verification performed with Claude Code (Opus 5.5): exact-SHA claim checks, a range-diff against the approved series, and a review of the exact-head CI logs.*

PR #332, opened 2026-08-19
ci: cache buf modules and wire optional BUF_TOKEN to stop BSR 403 flakes

ilsiepotamus:
## Summary

Anonymous buf.build (BSR) requests from shared GitHub-runner IPs get rate-limited and intermittently fail `buf generate` with `permission_denied: 403 Forbidden` — most recently [this Security Scan job on PR #327](https://github.com/teradata-labs/loom/actions/runs/32292247000/job/96195491089), where the 8 sibling jobs passed the identical step in the same run. No workflow authenticated to the BSR, and every job re-downloaded modules from scratch.

Two measures, one per traffic class:

- **Module downloads → cached.** A composite action (`.github/actions/buf-cache`) pins `BUF_CACHE_DIR` to a fixed location and caches it keyed on `buf.lock` + `buf.gen.yaml` (per-OS, with a prefix restore key). Wired into all 13 buf jobs across `ci.yml`, `nightly.yml`, `release.yml`, and `windows-quickstart.yml`, always after checkout. Module traffic to the BSR drops to cold-cache-only.
- **Remote plugin execution → authenticated when possible.** `buf.gen.yaml` uses remote plugins, so `buf generate` performs plugin execution on the BSR every run — a cache can't absorb that. A workflow-level `BUF_TOKEN` env (from the `BUF_TOKEN` repository secret) authenticates it. **The secret does not exist yet**: buf ignores an empty token, so behavior is unchanged until someone creates a buf.build token and adds it as a repo secret — at which point all remaining BSR traffic stops being anonymous. This PR is safe to merge before (or without) that step; the cache alone removes most of the flake surface.

## Verification

- `BUF_CACHE_DIR` honored by buf (cache populates `v3/{commits,modulelocks,plugins}`)
- `BUF_TOKEN=""` verified ignored (cold-cache `buf build` succeeds)
- `buf generate` with empty token + cold cache reproduces the committed `gen/` byte-for-byte
- All four workflows + the composite action parse as valid YAML; every inserted cache step verified to sit after its job's `actions/checkout`
- `actions/cache` pinned by SHA (`55cc8345…` = v6.1.0), matching the repo's action-pinning convention


ilsiepotamus:
Fresh full review: **no findings**. Verified: `buf.lock`/`buf.gen.yaml` exist at root so the cache key hashes real inputs; the composite is SHA-pinned; PR-branch caches can't poison main's (GitHub cache scoping); an absent `BUF_TOKEN` secret yields an empty env var buf ignores (fork-safe, stays anonymous as documented); the windows job genuinely runs buf and the `bash` composite shell works there. Merge-ready.

.github/workflows/ci.yml
```diff
@@ -12,6 +12,12 @@ permissions:
   checks: write
   security-events: write  # Required for SARIF upload to code scanning
 
+env:
+  # Authenticates buf's BSR traffic (module downloads and remote plugin
+  # execution) when the BUF_TOKEN repository secret exists; empty is ignored
+  # by buf and requests stay anonymous (rate-limited).
+  BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
```
Ed-Jaramillo, line 19: `BUF_TOKEN` at workflow scope reaches every step of every job in this workflow, not just the buf steps — that includes `go test -race ./...` over PR-authored code, `golangci-lint`, and the `gosec` container step. Only 13 of the jobs here use buf at all; `mcp-interop` and `coverage-check` receive it without needing it.

It's harmless while the secret is empty, but once it exists, anything running in those jobs can read `$BUF_TOKEN`. Same-repo PRs do receive secrets, so a PR that adds a test can reach it. (Fork PRs don't get secrets at all — worth knowing, since it also means fork contributors get no benefit from the token.)

Since it's cheap to settle now and awkward to retrofit once the secret is live, could we scope it down before creating it? Either per-step:

```yaml
      - name: Generate Proto Files
        run: buf generate
        env:
          BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
```

or at `jobs.<id>.env` on just the buf jobs if per-step is too repetitive. (Folding it into the composite action isn't an option — a composite action can't set job-scoped env for a secret it isn't passed, so it'd need an explicit `inputs:` parameter.)

Related question that sets how much this matters: what scope is the token expected to have? If it's read-only this is mostly hygiene; if it carries write/publish rights on the BSR it's worth doing before the secret lands, since our generated Go code derives from BSR content.

Same applies to the equivalent blocks in `nightly.yml`, `release.yml`, and `windows-quickstart.yml`.

.github/actions/buf-cache/action.yml
```diff
@@ -0,0 +1,26 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Pin buf cache location
+      shell: bash
+      run: echo "BUF_CACHE_DIR=${{ runner.temp }}/buf-module-cache" >> "$GITHUB_ENV"
+
+    - name: Restore buf module cache
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```
Ed-Jaramillo, line 21: The pin itself is correct — I checked it against the API rather than the comment: `v6.1.0` → `55cc8345863c7cc4c66a329aec7e433d2d1c52a9`, and v6.1.0 is the current latest release.

Heads-up though that this is the repo's first external action pinned *inside* a composite action — the pre-existing `.github/actions/setup-gpg/action.yml` pins none, so there's no counter-example in history. Dependabot's `github-actions` ecosystem doesn't update references in `.github/actions/*/action.yml`; the docs describe `directory: "/"` as checking *"for workflow files in `.github/workflows`"*, and the gap is a long-standing open request ([dependabot-core#6704](https://github.com/dependabot/dependabot-core/issues/6704), [#7495](https://github.com/dependabot/dependabot-core/issues/7495), [#6936](https://github.com/dependabot/dependabot-core/issues/6936)).

So `actions/cache@55cc8345` won't get the automatic bumps every other pin in this repo receives — and that's easy to miss precisely *because* everything else is maintained for us. If an advisory lands against it, no Dependabot PR appears.

Either a comment noting the pin is manually maintained, or moving the `actions/cache` step up into the workflows and leaving this action to just set `BUF_CACHE_DIR`, would keep it from quietly aging.

.github/actions/buf-cache/action.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        path: ${{ runner.temp }}/buf-module-cache
+        key: buf-modules-${{ runner.os }}-${{ hashFiles('buf.lock', 'buf.gen.yaml') }}
```
Ed-Jaramillo, line 24: The key doesn't include the buf CLI version. buf namespaces its cache by an internal directory (`v3/` today) — if a future release moves to `v4/`, the restored payload becomes dead weight, buf repopulates the new directory, and `actions/cache` then **skips the save** because the primary key still matches exactly.

The Security Scan job already exercises that code path today:

```
Cache hit occurred on the primary key buf-modules-Linux-9127767a…, not saving cache.
```

Which is correct behaviour now, but after a `buf-setup-action` bump it's the mechanism that would keep the new cache from ever being written. The failure mode is silent — CI stays green while the cache quietly stops working, and the module traffic this PR set out to remove comes back. Recovery would need an unrelated edit to `buf.lock` or `buf.gen.yaml`.

Worth folding the buf version into the key — either by hashing the workflow files that pin it, or by parameterising this action with the version already declared at each `bufbuild/buf-setup-action` call site.

Ed-Jaramillo (commented):
Thanks for chasing this down — the caching itself is clean, and I verified it works end to end: cache saved on all three OSes (`buf-modules-Linux-…`, `-Windows-…`, `-macOS-…`), hits within the same run, `BUF_CACHE_DIR` honored including the mixed-separator `D:\a\_temp/buf-module-cache` path on Windows, and `buf generate` reproducing the committed `gen/` byte-for-byte with an empty token.

I also tested the failure mode I was most worried about — introducing a cache into the code-generation input path. Appending a byte to a cached `google/api/annotations.proto` and rebuilding produced:

```
Failure: *** Digest verification failed for "buf.build/googleapis/googleapis:72c8614f3bd0466ea67931ef2c43d608" ***
	Expected digest (from buf.lock): "b5:13efeea24e…"
	Actual digest: "b5:9366f4a37e…"
	This may be the result of a hand-edited or corrupted buf.lock file, a corrupted local cache, and/or an attack.
```

So the integrity guarantee is real and enforced by buf. No objection to merging.

---

One thing I'd like to reconcile before we treat the flake as closed.

Buf's [rate-limits page](https://buf.build/docs/bsr/rate-limits/) documents limit exceedance as *"HTTP 429 with `X-RateLimit-Remaining: 0`, the Connect error code `resource_exhausted`, and a `Retry-After` header."* The failure we saw was `permission_denied: 403 Forbidden` — different status, different Connect code. So rate limiting, as buf implements it, doesn't produce that message, and the root cause looks unconfirmed.

The same table also splits the buckets in a way that matters here:

| Bucket | Sustained rate | Burst |
|---|---|---|
| General API | 30 req/sec | 60 |
| Code generation, unauthenticated | **10 req/hour** | 10 |
| Code generation, authenticated | 960 req/hour | 120 |
| FileDescriptorSetService | 1 req/sec | 2 |

Mapping our traffic onto it, from `buf generate --debug`:

- **Module resolution is General API.** On a cold cache the first BSR call is `buf.registry.module.v1.GraphService/GetGraph`, right after the module store miss. That bucket is 30 req/sec **and identical whether or not you authenticate** — so `BUF_TOKEN` buys nothing here, and 13 jobs are not plausibly exhausting it.
- **Remote plugin execution is code generation.** I measured exactly **one** `CodeGenerationService/GenerateCode` per `buf generate` (all four remote plugins batched into a single request). Every one of the 13 cached jobs runs `buf generate` — including the Windows one, via `quickstart.ps1:314` — so each CI run issues ~13 requests against a **10 req/hour anonymous** budget.

And caching can't touch that second bucket: after a *successful* `buf generate`, `v3/plugins` stays empty and the cache size is unchanged, and a warm-cache run with the network blocked still fails at plugin execution (while `buf build`, `buf lint`, and `buf breaking` all succeed).

So as I read it: the cache removes traffic from the roomy bucket, and the tight bucket is reachable only via `BUF_TOKEN` — which is inert until the secret exists.

I'm still happy to see this merge; determinism, speed, and fewer network dependencies are worth having on their own. Two asks:

1. Could we adjust the PR description **and** the `description:` block in the composite action (lines 3–11, which repeat the same rationale) so neither reads as "flake fixed"?
2. Could we track creating the `BUF_TOKEN` secret as the actual remediation, rather than closing this out on the merge?

---

<details>
<summary>Minor, non-blocking: recurring <code>Failed to save</code> warnings are expected here</summary>

Jobs in a run start together, all miss the same key, and all but one fail to reserve it — e.g. `Failed to save: Unable to reserve cache with key buf-modules-Linux-9127767a…, another job may be creating this cache.` in Build ubuntu and Go Lint. The Windows quickstart job also collides cross-workflow with `Build windows-latest` on the shared `buf-modules-Windows-…` key.

All of those jobs pass and one save per OS does succeed, so this is benign — flagging it only so the recurring warning isn't later mistaken for a defect. A `-${{ github.job }}` suffix would trade the warnings for redundant per-job cache entries, which is probably not worth it.

</details>

*Review notes: findings verified against the PR head `9d87f74`. Posting as comments rather than an approval — leaving the approve/merge call to you.*

Ed-Jaramillo (approved):
Approving. The change is additive-only, touches no application code, and I verified it works end to end rather than taking CI's word for it — cache saved on all three OSes, hits within the same run, `BUF_CACHE_DIR` honored including the mixed-separator Windows path, and `buf generate` reproducing the committed `gen/` byte-for-byte with an empty token. Cache content can't silently corrupt code generation: buf enforces the `buf.lock` digest and hard-fails on mismatch, which I tested directly by tampering with a cached file. Clean single-commit revert if it misbehaves.

The findings in my [earlier review](https://github.com/teradata-labs/loom/pull/332#pullrequestreview-5012813703) are all non-blocking and none need to hold up the merge.

One thing I'd ask us not to lose, though: per Buf's [rate-limits table](https://buf.build/docs/bsr/rate-limits/), module traffic sits in the General API bucket (30 req/sec, identical authenticated or not), while the constrained bucket is code generation at 10 req/hour unauthenticated — one request per `buf generate`, across 13 jobs, and not cacheable. So this merge is unlikely to stop the 403s by itself; `BUF_TOKEN` is the lever that moves that limit to 960/hour, and it stays inert until the repository secret exists.

Merging is the right call — the determinism and reduced network dependence stand on their own. I'd just suggest keeping the flake ticket open until the secret is created.

PR #333, opened 2026-08-19
feat(mcp): loom-mcp-probe — real-server wire testing for the 2026-07-28 client

ilsiepotamus:
## Summary

First deliverable of the real-wire testing plan for the merged MCP client (#327): a probe CLI that connects Loom's client to **real** servers and reports what actually negotiated — revision, era, server identity, tools, an optional tool call, an optional **MRTR elicitation exchange**, and an optional **subscriptions/listen watch**. It exercises the exact negotiation, fallback, MRTR, and transport paths the manager uses in production. `just mcp-probe` runs it.

Includes one client fix the probing itself surfaced (see below).

## Verified live (2026-08-19), all official servers

- **Go SDK conformance everything-server** (the only official 2026-07-28 runtime today):
  - stateless streamable HTTP → negotiated `2026-07-28`, 28 tools, tool call, live subscription delivered a real `notifications/resources/updated`, clean client-side cancel
  - stdio → `2026-07-28` including the stdio `notifications/cancelled` unsubscribe path
  - sessionful mode (`-stateless=false`) → fallback negotiated `2025-11-25` (the newest-legacy request from the #327 review round 1, live)
  - **MRTR**: `test_input_required_result_elicitation` → elicited, answered, `"Hello, Loom!"`; `test_input_required_result_request_state` → the server itself confirmed the opaque `requestState` echoed correctly across the retry
  - **pin matrix**: exact pins honored in both eras; `-pin 2026-07-28` vs a legacy-only endpoint fails loudly with `-32022` listing the server's versions (no silent downgrade)
- **npm `@modelcontextprotocol/server-everything`** (TypeScript SDK):
  - stdio → fallback → `2025-11-25`, `echo` round-trip; client-side schema validation rejected a missing required arg before the wire
  - **sessionful streamable HTTP** (`streamableHttp` mode, :3001/mcp) → the server rejects the pre-initialize discover probe with **HTTP 400 + non-modern `-32000` body (id null)** — the exact case the #327 round-2 fallback rule handles — then the handshake runs, the minted `Mcp-Session-Id` is adopted and echoed on every later request (this server enforces sessions strictly, so the succeeding `echo` call is the proof)
- **Dogfood: Loom's own dual-era endpoint** (`loom-mcp --transport=http` bridging `looms` from the #328 branch): one endpoint verified serving **both eras** — auto-negotiation → `2026-07-28` stateless with 55 bridge tools, a real gRPC-backed `loom_list_agents` round-trip, and `subscriptions/listen` acknowledged by our own server; `-pin legacy` → the `2024-11-05` initialize handshake, matching `handleDiscover`'s advertised `[2026-07-28, 2024-11-05]`

Field notes recorded in the guide: the TS SDK tops out at 2025-11-25 (legacy-ladder peer, not a 2026 target); the SDK server's legacy `test_elicitation` tool is rejected by the SDK itself on 2026 sessions (use `test_input_required_result_*`); the SDK server is dual-era even in stateless mode; it exits 1 after a cancelled stdio listen (bare EOF exits 0 — Loom unaffected); the TS server's pre-initialize 400 semantics are why the fallback rule inspects 400 bodies.

## Client fix surfaced by the probing (`0be8e9e6`)

`dispatchAndWait` logged transport send failures at **Error with a stack trace** — which means every healthy legacy connect against a strict session server (the fallback-400 path above) was reported as an incident. The error is returned to the caller, who knows its actual severity, so the log is demoted to Debug. Without real-wire probing this double-reporting was invisible: no fake ever exercised the fallback through a transport error.

## Tests

Wire-level tests (httptest, real HTTP through the shipped client): stateless negotiation, legacy fallback, and a full MRTR round asserting the retry carries `inputResponses` **plus the exact `requestState`**; handler guards (sampling refused, bad `-answer` JSON); helper tables. `go test -tags fts5 -race ./cmd/loom-mcp-probe` clean; `just check` green.

## Docs

`docs/guides/mcp-probe.md` (flags, what each mode exercises, known-good official servers, real transcripts — including the sessionful-HTTP session-capture and the Loom dual-era dogfood — limitations, field notes), README MCP Integration section + docs table, homepage mcp feature row updated to state negotiated-revision support and the probe — all within implemented-and-verified claims.

### DIFF
--- a/cmd/loom-mcp-probe/main.go
+++ b/cmd/loom-mcp-probe/main.go
@@ -16,15 +16,16 @@
 // HTTP or stdio — and reports what actually negotiated: protocol revision,
 // era (stateless 2026-07-28 core vs legacy initialize handshake), server
 // identity, the tool list, an optional tool call, and an optional
-// subscriptions/listen watch. No fakes anywhere: it exercises the same
-// negotiation, fallback, and transport paths the manager uses in production,
-// against whatever server it is pointed at.
+// subscriptions/listen watch. The probe runs the shipped client and
+// transports unmodified — the same negotiation, fallback, and transport
+// paths the manager uses in production — against whatever real server it is
+// pointed at (its automated tests, by contrast, use scripted HTTP fixtures).
 //
 // Examples:
 //
 //	loom-mcp-probe -url http://localhost:8971 -call test_simple_text -watch 5
-//	loom-mcp-probe -cmd "npx -y @modelcontextprotocol/server-everything stdio" \
-//	    -call echo -args '{"message":"hello"}'
+//	loom-mcp-probe -cmd npx -arg -y -arg @modelcontextprotocol/server-everything \
+//	    -arg stdio -call echo -args '{"message":"hello"}'
 //	loom-mcp-probe -url http://localhost:9090/mcp -pin legacy
 //	loom-mcp-probe -url http://localhost:8971 -call test_elicitation \
 //	    -args '{"message":"pick a username"}' -answer '{"username":"loom"}'
@@ -56,15 +57,17 @@ import (
 
 // options selects the server to probe and what to exercise against it.
 type options struct {
-	URL      string        // streamable HTTP endpoint; mutually exclusive with Cmd
-	Cmd      string        // stdio server command line; mutually exclusive with URL
-	Pin      string        // protocol_version pin: "" or "auto", "legacy", or an exact revision
-	Call     string        // tool to invoke, if any
-	Args     string        // JSON arguments for Call
-	Answer   string        // JSON object accepted for every elicitation (enables the MRTR driver)
-	WatchSec int           // seconds to hold a subscriptions/listen stream (stateless only)
-	Timeout  time.Duration // negotiation-probe/request timeout
-	Verbose  bool
+	URL        string        // streamable HTTP endpoint; mutually exclusive with Cmd
+	Cmd        string        // stdio server executable (verbatim; args ride CmdArgs); mutually exclusive with URL
+	CmdArgs    []string      // arguments for Cmd, one per -arg flag
+	HeadersEnv string        // env var holding a JSON object of HTTP headers (auth without argv exposure)
+	Pin        string        // protocol_version pin: "" or "auto", "legacy", or an exact revision
+	Call       string        // tool to invoke, if any
+	Args       string        // JSON arguments for Call
+	Answer     string        // JSON object accepted for every elicitation (enables the MRTR driver)
+	WatchSec   int           // seconds to hold a subscriptions/listen stream (stateless only)
+	Timeout    time.Duration // per-operation timeout: connect, tools/list, tools/call (incl. MRTR rounds), subscription ack
+	Verbose    bool
 }
 
 // report is what the probe observed on the wire.
@@ -78,8 +81,7 @@ type report struct {
 	CallOutput      string   // first content text of the tool call, if Call was set
 	Elicited        []string // elicitation messages answered by the MRTR driver
 	Notifications   []string // methods received while watching, if WatchSec > 0
-	WatchEndErr     error    // nil = graceful or client-cancelled
-	WatchSkipped    bool     // legacy connection cannot subscribe
+	WatchEndErr     error    // nil = held for the full window (client-cancelled at the deadline)
 }
 
 const toolNamesShown = 6
@@ -96,13 +98,18 @@ func main() {
 	opts := options{}
 	var timeoutMs int
 	flag.StringVar(&opts.URL, "url", "", "streamable HTTP endpoint (e.g. http://localhost:8971)")
-	flag.StringVar(&opts.Cmd, "cmd", "", "stdio server command line (space-separated)")
+	flag.StringVar(&opts.Cmd, "cmd", "", "stdio server executable (path taken verbatim; pass arguments via repeated -arg)")
+	flag.Func("arg", "argument for the -cmd executable (repeatable, order-preserving)", func(v string) error {
+		opts.CmdArgs = append(opts.CmdArgs, v)
+		return nil
+	})
+	flag.StringVar(&opts.HeadersEnv, "headers-env", "", "name of an env var holding a JSON object of HTTP headers (e.g. auth tokens; never passed on the command line)")
 	flag.StringVar(&opts.Pin, "pin", "", `protocol_version pin: "auto" (default), "legacy", or an exact revision`)
 	flag.StringVar(&opts.Call, "call", "", "tool to invoke")
 	flag.StringVar(&opts.Args, "args", "{}", "JSON arguments for -call")
 	flag.StringVar(&opts.Answer, "answer", "", "JSON object accepted for every elicitation; enables the MRTR driver (default: fail fast on input_required)")
 	flag.IntVar(&opts.WatchSec, "watch", 0, "seconds to hold a subscriptions/listen stream (stateless connections only)")
-	flag.IntVar(&timeoutMs, "timeout", 15000, "negotiation-probe/request timeout in ms")
+	flag.IntVar(&timeoutMs, "timeout", 15000, "per-operation timeout in ms: connect (incl. negotiation probe), tools/list, tools/call with its MRTR rounds, and the subscription acknowledgment")
 	flag.BoolVar(&opts.Verbose, "v", false, "debug logging")
 	flag.Parse()
 	opts.Timeout = time.Duration(timeoutMs) * time.Millisecond
@@ -113,10 +120,7 @@ func main() {
 		os.Exit(2)
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
-	defer cancel()
-
-	rep, err := run(ctx, opts, os.Stdout)
+	rep, err := run(context.Background(), opts, os.Stdout)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "PROBE FAILED: %v\n", err)
 		os.Exit(1)
@@ -133,12 +137,28 @@ func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
 		return nil, fmt.Errorf("logger: %w", err)
 	}
 
+	rep := &report{}
+
+	// Every flag that can fail validation is checked before the transport is
+	// built: a stdio transport starts a subprocess, and a validation error
+	// after that point would leak the child with no cleanup registered.
+	var callArgs map[string]interface{}
+	if opts.Call != "" {
+		if err := json.Unmarshal([]byte(opts.Args), &callArgs); err != nil {
+			return nil, fmt.Errorf("-args is not a JSON object: %w", err)
+		}
+	}
+	var mrtr client.InputHandler
+	if opts.Answer != "" {
+		if mrtr, err = buildMRTRHandler(opts.Answer, rep, pr); err != nil {
+			return nil, err
+		}
+	}
+
 	tr, err := buildTransport(opts, logger)
 	if err != nil {
 		return nil, fmt.Errorf("transport: %w", err)
 	}
-
-	rep := &report{}
 
 	cfg := client.Config{
 		Transport:       tr,
@@ -146,11 +166,7 @@ func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
 		ProtocolVersion: opts.Pin,
 		RequestTimeout:  opts.Timeout,
 	}
-	if opts.Answer != "" {
-		mrtr, err := buildMRTRHandler(opts.Answer, rep, pr)
-		if err != nil {
-			return nil, err
-		}
+	if mrtr != nil {
 		cfg.MRTR = client.MRTRConfig{Handler: mrtr}
 	}
 
@@ -158,7 +174,10 @@ func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
 	defer func() { _ = c.Close() }()
 
 	start := time.Now()
-	if err := c.Connect(ctx, protocol.Implementation{Name: "loom-mcp-probe", Version: version.Version}); err != nil {
+	connectCtx, cancelConnect := context.WithTimeout(ctx, opts.Timeout)
+	err = c.Connect(connectCtx, protocol.Implementation{Name: "loom-mcp-probe", Version: version.Version})
+	cancelConnect()
+	if err != nil {
 		return nil, fmt.Errorf("connect: %w", err)
 	}
 	rep.ConnectDuration = time.Since(start)
@@ -171,7 +190,9 @@ func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
 	pr.f("  era        : %s\n", era(rep.Stateless))
 	pr.f("  serverInfo : %s %s\n", rep.ServerInfo.Name, rep.ServerInfo.Version)
 
-	tools, err := c.ListTools(ctx)
+	listCtx, cancelList := context.WithTimeout(ctx, opts.Timeout)
+	tools, err := c.ListTools(listCtx)
+	cancelList()
 	if err != nil {
 		return rep, fmt.Errorf("tools/list: %w", err)
 	}
@@ -186,7 +207,10 @@ func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
 		rep.ToolCount, strings.Join(rep.ToolNames, ", "), more(rep.ToolCount, toolNamesShown))
 
 	if opts.Call != "" {
-		if err := runCall(ctx, c, opts, rep, pr); err != nil {
+		callCtx, cancelCall := context.WithTimeout(ctx, opts.Timeout)
+		err := runCall(callCtx, c, opts, callArgs, rep, pr)
+		cancelCall()
+		if err != nil {
 			return rep, err
 		}
 	}
@@ -200,11 +224,7 @@ func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
 	return rep, nil
 }
 
-func runCall(ctx context.Context, c *client.Client, opts options, rep *report, pr printer) error {
-	var args map[string]interface{}
-	if err := json.Unmarshal([]byte(opts.Args), &args); err != nil {
-		return fmt.Errorf("-args is not a JSON object: %w", err)
-	}
+func runCall(ctx context.Context, c *client.Client, opts options, args map[string]interface{}, rep *report, pr printer) error {
 	res, err := c.CallTool(ctx, opts.Call, args)
 	if err != nil {
 		return fmt.Errorf("tools/call %s: %w", opts.Call, err)
@@ -221,10 +241,11 @@ func runCall(ctx context.Context, c *client.Client, opts options, rep *report, p
 }
 
 func runWatch(ctx context.Context, c *client.Client, opts options, rep *report, pr printer) error {
+	// A requested watch is a gate: every path that cannot deliver a healthy,
+	// acknowledged subscription held for the full window is a probe failure,
+	// never a silent exit 0.
 	if !c.IsStateless() {
-		rep.WatchSkipped = true
-		pr.f("  watch      : skipped (legacy connection; subscriptions/listen is 2026-07-28)\n")
-		return nil
+		return fmt.Errorf("-watch requested, but the connection is legacy (%s): subscriptions/listen is 2026-07-28 only", c.NegotiatedVersion())
 	}
 	subCtx, subCancel := context.WithCancel(ctx)
 	defer subCancel()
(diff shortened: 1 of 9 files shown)


cmd/loom-mcp-probe/main.go
```diff
@@ -0,0 +1,345 @@
...
+
+	deadline := time.After(time.Duration(opts.WatchSec) * time.Second)
+	for {
+		select {
+		case n, ok := <-sub.Notifications:
+			if !ok {
+				rep.WatchEndErr = sub.Err()
+				pr.f("  watch done : %d notifications, end state: %v\n",
+					len(rep.Notifications), endState(rep.WatchEndErr))
+				return nil
```
Ed-Jaramillo: **BLOCKER:** `-watch` exits successfully when the subscription closes with `sub.Err()`, when no acknowledgment arrives before the deadline, or when a requested watch is skipped on a legacy connection. I reproduced MethodNotFound, synthesized stream loss, no-ack timeout, and legacy skip at this head. Please require `notifications/subscriptions/acknowledged`, propagate terminal or early-close errors, reject unsupported legacy watches, and add regression tests for all four paths.

cmd/loom-mcp-probe/main.go
```diff
@@ -0,0 +1,345 @@
...
+	opts := options{}
+	var timeoutMs int
+	flag.StringVar(&opts.URL, "url", "", "streamable HTTP endpoint (e.g. http://localhost:8971)")
+	flag.StringVar(&opts.Cmd, "cmd", "", "stdio server command line (space-separated)")
+	flag.StringVar(&opts.Pin, "pin", "", `protocol_version pin: "auto" (default), "legacy", or an exact revision`)
+	flag.StringVar(&opts.Call, "call", "", "tool to invoke")
+	flag.StringVar(&opts.Args, "args", "{}", "JSON arguments for -call")
+	flag.StringVar(&opts.Answer, "answer", "", "JSON object accepted for every elicitation; enables the MRTR driver (default: fail fast on input_required)")
+	flag.IntVar(&opts.WatchSec, "watch", 0, "seconds to hold a subscriptions/listen stream (stateless connections only)")
+	flag.IntVar(&timeoutMs, "timeout", 15000, "negotiation-probe/request timeout in ms")
```
Ed-Jaramillo: **MAJOR:** This is documented as a request timeout, but `client.Config.RequestTimeout` intentionally applies only to `server/discover`; initialization, `tools/list`, calls, and subscription opening use the fixed 120-second parent context. A stalled `tools/list` ignored the configured probe timeout in reproduction. Please rename/narrow this flag or add a distinct overall operation timeout, with a separate watch-duration context.

cmd/loom-mcp-probe/main.go
```diff
@@ -0,0 +1,345 @@
...
+
+func buildTransport(opts options, logger *zap.Logger) (transport.Transport, error) {
+	if opts.URL != "" {
+		return transport.NewStreamableHTTPTransport(transport.StreamableHTTPConfig{
+			Endpoint:       opts.URL,
+			EnableSessions: true, // legacy servers may mint a session; 2026 servers never do
+			Logger:         logger,
+		})
+	}
+	parts := strings.Fields(opts.Cmd)
```
Ed-Jaramillo: **MAJOR:** `strings.Fields` cannot preserve quoted stdio arguments or Windows executable paths containing spaces, and whitespace-only `-cmd` panics at `parts[0]`. Please accept the executable separately from repeatable arguments and validate it before indexing.

docs/guides/mcp-probe.md
```diff
@@ -0,0 +1,141 @@
...
+$ loom-mcp-probe -url http://localhost:3001/mcp -call echo -args '{"message":"session test"}'
+CONNECTED in 6ms
+  negotiated : 2025-11-25
+  era        : legacy (initialize handshake)
+  serverInfo : mcp-servers/everything 2.0.0
+  tools      : 13 (echo, get-annotated-message, get-env, …)
+  call echo → "Echo: session test"
+```
+
+**Dogfood: Loom's own dual-era endpoint** (`loom-mcp --transport=http` bridging a running `looms`). One endpoint serves both eras:
```
Ed-Jaramillo: **MAJOR:** This presents the #328 dual-era server and, below, dual-revision/official-SDK conformance CI as current. At this head, `loom-mcp --transport=http` is still the legacy endpoint and `pkg/mcp/conformance` is one legacy test file with no `mcp-interop` workflow. #333 is not stacked on #328. Please defer or clearly mark these claims until #328 lands.

cmd/loom-mcp-probe/main.go
```diff
@@ -0,0 +1,345 @@
...
+			pr.f("  elicited   : %q → accepting with -answer\n", params.Message)
+			responses[id] = accept
+		}
+		return responses, nil
+	}, nil
+}
+
+func buildTransport(opts options, logger *zap.Logger) (transport.Transport, error) {
+	if opts.URL != "" {
+		return transport.NewStreamableHTTPTransport(transport.StreamableHTTPConfig{
```
Ed-Jaramillo, line 389: **MAJOR:** The manager passes configured HTTP headers to this transport, but the probe has no equivalent, so authenticated endpoints that work in agent configuration fail here with 401. Please add a safe header/config source that does not expose secrets in process arguments, or document the unauthenticated-only limitation.

cmd/loom-mcp-probe/main.go
```diff
@@ -0,0 +1,345 @@
...
+// run performs the probe and streams human-readable progress to out. The
+// returned report carries the same observations for programmatic use.
+func run(ctx context.Context, opts options, out io.Writer) (*report, error) {
+	pr := printer{w: out}
+	logger, err := buildLogger(opts.Verbose)
+	if err != nil {
+		return nil, fmt.Errorf("logger: %w", err)
+	}
+
+	tr, err := buildTransport(opts, logger)
```
Ed-Jaramillo, line 176: **MINOR:** A stdio subprocess is started here before `-answer` JSON is validated. That validation error returns before any transport/client cleanup is registered; I confirmed the child remains alive after `run` returns. Please validate `-answer` before constructing the transport.

cmd/loom-mcp-probe/main_test.go
```diff
@@ -0,0 +1,232 @@
...
+	statelessOK bool
+	elicitFirst bool
+
+	mu         sync.Mutex
+	callParams []json.RawMessage // raw params of each tools/call attempt
+}
+
+func (s *scriptedHTTPServer) handler(w http.ResponseWriter, r *http.Request) {
+	body, err := io.ReadAll(r.Body)
+	require.NoError(s.t, err)
```
Ed-Jaramillo: **MINOR:** This handler runs in an HTTP server goroutine, where `require.*` may call `FailNow`; Go requires `FailNow` to run in the test goroutine. Please return HTTP errors or record failures for assertion by the test goroutine.

docs/guides/mcp-probe.md
```diff
@@ -0,0 +1,141 @@
+# MCP Probe: real-server testing for Loom's MCP client
+
+**Status**: ✅ Implemented (`cmd/loom-mcp-probe`, wire-level tests included)
+
+`loom-mcp-probe` connects Loom's MCP client to a **real** MCP server — streamable HTTP or stdio — and reports what actually negotiated: protocol revision, era, server identity, tools, an optional tool call, an optional MRTR elicitation exchange, and an optional `subscriptions/listen` watch. It exercises the exact negotiation, fallback, MRTR, and transport code paths the manager uses in production. No fakes anywhere.
```
Ed-Jaramillo: **MINOR:** "No fakes anywhere" is broader than the adjacent evidence: automated tests use `scriptedHTTPServer`. The runtime probe does use the shipped client/transports and the manual sessions use real servers; please distinguish those from the HTTP test fixtures.

Ed-Jaramillo (changes requested):
Requesting changes because `-watch` can exit 0 when no healthy subscription was established, contradicting the documented guarantee that this probe can gate scripts. I independently reproduced MethodNotFound, early SSE stream loss, no acknowledgment, and legacy skip at this exact head under the race detector.

PR #333 is not technically blocked by #328: the probe exercises the client merged in #327 and can land independently after its own defects are fixed. However, the guide currently presents #328-only dual-era server behavior and planned conformance CI as functionality in this repository. Those claims need to be removed, marked as pending, or introduced by explicitly stacking/rebasing after #328.

The official Go SDK negotiation, fallback, MRTR, subscription acknowledgment, and cancellation claims were independently reproducible. The remaining comments cover timeout semantics, stdio parsing, authenticated endpoints, cleanup, test safety, and documentation precision.

ilsiepotamus:
All eight verified and fixed in `95eac806`. The blocker was exactly right — three separate exit-0 escapes plus a skip, each individually enough to break the gating guarantee the guide advertises.

**Blocker — `-watch` gate:** every path that cannot deliver a healthy, acknowledged subscription held for the full window is now a probe failure. A legacy connection with `-watch` requested is an error (was: silent skip); the acknowledgment must arrive as the first stream item within the operation timeout; and any subscription end before the requested window — terminal error *or* graceful server-side closure — fails the probe. The only healthy end is client-side cancellation at the deadline. Four regressions cover the four paths you reproduced (legacy skip, MethodNotFound, no-ack timeout, ack-then-early-close), driven over scripted wire-level SSE fixtures.

**Timeout semantics:** `-timeout` is now what its help says — a per-operation bound applied to connect (including the negotiation probe), `tools/list`, `tools/call` with its MRTR rounds, and the subscription acknowledgment — and the undocumented fixed 120-second envelope is gone. The `-watch` hold runs on its own clock.

**stdio argv:** `-cmd` takes the executable verbatim (Windows paths with spaces survive), arguments ride a repeatable `-arg` flag in order, and a whitespace-only `-cmd` is a validation error instead of an index panic. Doc examples updated to the new form.

**Docs vs. #328:** the guide no longer presents the dual-era `loom-mcp` endpoint or the §10 conformance/interop CI as current functionality — both are marked 🚧 pending PR #328, with the dogfood transcript explicitly labeled as captured against that branch. Agreed #333 stands on the merged #327 client and doesn't need to stack.

**Auth headers:** `-headers-env <VAR>` reads a JSON object of HTTP headers from the named environment variable and feeds the same `Headers` field the manager's transport config uses — tokens never appear in argv or shell history. Documented in the flag table and limitations.

**Minors:** `-args`/`-answer` (and every other fallible flag) validate before the transport is built, so a stdio subprocess is never started and then orphaned by a flag error; the scripted test server's handler reports internal failures as HTTP 500s surfaced by client-side assertions instead of calling `FailNow` off the test goroutine; and "no fakes anywhere" now says precisely what's true — the runtime probe is the shipped client/transports against real servers, the automated tests use scripted HTTP fixtures.

Package suite green under `-race`, lint clean.

ilsiepotamus:
Base note: #328 merged (`bc13fa0b`), so the guide's 🚧 pending-#328 markers from the round-1 fixes just became stale — `db8bce6a` restores the dual-era endpoint and §10 conformance claims as current, attributed to the merged PR.

Ed-Jaramillo (changes requested):
Re-review at `95eac806`. The six findings from the previous round are all resolved — watch failure paths, argument parsing, auth headers, validation ordering, test safety, and the doc claims all look good now. Two issues remain before this is ready to merge, both verified against the code:

### 1. Major — `-timeout` is not enforced during subscription establishment (`cmd/loom-mcp-probe/main.go:250-262`)

`Subscribe` can block indefinitely waiting on HTTP response headers before the ack timer ever starts:

- `Subscribe` → `transport.Send` → `t.client.Do(req)`, and the transport uses a zero-value `&http.Client{}` (`streamable_http.go:96`), so `Do` waits on headers with no deadline.
- `RequestTimeout` doesn't cover this path — it only bounds the `server/discover` negotiation probe (`connect.go:311`).
- Connect, list, and call each get `context.WithTimeout(ctx, opts.Timeout)` (`main.go:177`, `:193`, `:210`), but `runWatch` receives the bare outer context (`main.go:219`) and derives a cancel-only `subCtx`. The `ackTimer` at `main.go:262` starts only after `Subscribe` returns — after headers have already arrived. A server that hangs before sending headers isn't caught by `-timeout` at all, which contradicts the flag's documented contract ("bounds … the subscription acknowledgment").

**Fix:** derive the subscription context with `context.WithTimeout(ctx, opts.Timeout)` so the header-wait and the ack-wait share one deadline. The watch hold itself should keep running on its own clock as documented. Please add a test with a fixture that stalls before sending headers to lock this in.

### 2. Major — stdio examples still use the old `-cmd` form (`docs/guides/mcp-probe.md:14`, `:87`)

The examples pass the full command as one quoted string:

```bash
-cmd "npx -y @modelcontextprotocol/server-everything stdio"
```

Since the argument-parsing change, `-cmd` is taken verbatim as the executable path (`main.go:101-104`, spawned with `Args: opts.CmdArgs`, no re-tokenization at `main.go:351-357`), so this string would be treated as the binary name and fail to exec. The flag table in the same doc and the header comment at `main.go:27` already show the correct semantics — it's just these two examples that lag behind.

**Fix:** update both to:

```bash
-cmd npx -arg -y -arg @modelcontextprotocol/server-everything -arg stdio
```

---

Targeted probe tests pass locally under `go test -tags fts5 -race ./cmd/loom-mcp-probe`. Both fixes are small — happy to re-review as soon as they land.

ilsiepotamus:
Both fixed in `8385596e`.

1. **Subscription establishment now rides `-timeout`.** A pre-ack watchdog cancels the subscription context at the deadline, so the transport's header wait inside `Subscribe` — which you're right has no deadline of its own (`http.Client{}` zero value) — and the ack wait share one bound, while the `-watch` hold keeps its own clock as documented. I went with a watchdog + cancel rather than `context.WithTimeout` on the subscription context because that context governs the whole stream: a hard deadline there would kill any hold longer than `-timeout` (e.g. `-watch 60` with the 15s default). Every pre-ack failure path reports the timeout distinctly when the watchdog was the cause. Regression: a fixture that stalls before sending any response headers, asserting the probe fails with the no-acknowledgment error within the bound rather than waiting the server out.

2. **Both stdio examples updated to `-cmd npx -arg …`.** Embarrassing root cause worth owning: the round-1 doc edit for exactly these lines silently no-opped on a whitespace mismatch in a scripted replacement. The fix asserts its anchors and I've grep-verified the old form is gone from the guide.

Package suite green under `-race` (`-count=2`), lint clean.

ilsiepotamus:
Fresh full re-review with independent eyes at `8385596e` — the core gating guarantees all verified on the wire paths (`isError:true` fails, pins never fall back, `input_required` without `-answer` fails fast, every watch escape errors, subprocess lifecycle bounded). Seven minors surfaced, all closed in `dfa5474b`: a negative `-watch` silently skipped the watch and exited 0; orphan flags (`-args`/`-answer` without `-call`, `-arg` with `-url`, `-headers-env` with `-cmd`) were silently ignored — a script could believe auth or MRTR was exercised when nothing was; `-answer` after a legacy fallback exited 0 with the MRTR driver provably unreachable (now an error mirroring the `-watch` legacy rule); `just mcp-probe` re-tokenized quoted JSON (`{{ARGS}}` → positional `\"$@\"`); the flagship doc example drove `test_elicitation`, the one tool the guide's own field notes say the SDK rejects on 2026 sessions; probe-level fixtures now pin the gate contract where it's advertised (isError, bad `-args` JSON, unanswered `input_required`, legacy `-answer`); and the pre-existing stdio stalled-child teardown edge is documented with an outer-timeout recommendation. Package suite green under `-race`, lint clean.

cmd/loom-mcp-probe/main.go
```diff
@@ -0,0 +1,448 @@
...
+		rep.ToolCount, strings.Join(rep.ToolNames, ", "), more(rep.ToolCount, toolNamesShown))
+
+	if opts.Call != "" {
+		callCtx, cancelCall := context.WithTimeout(ctx, opts.Timeout)
+		err := runCall(callCtx, c, opts, callArgs, rep, pr)
+		cancelCall()
+		if err != nil {
+			return rep, err
+		}
+	}
```
Ed-Jaramillo, line 247: **MAJOR:** `-answer` can still exit 0 without exercising MRTR on a stateless connection. The legacy guard above only proves MRTR is reachable; if the selected tool returns a normal `complete` result without `input_required`, `runCall` succeeds, `rep.Elicited` remains empty, and the probe reports success even though the supplied answer was never used. I reproduced this with the scripted stateless server: `run` returned `nil` with an empty elicitation list. Please fail after the call when `-answer` was supplied but no elicitation occurred, and add a regression using a stateless tool that completes immediately.

Ed-Jaramillo (changes requested):
Re-review at `dfa5474b`, requested 2026-08-24. No changes have landed since my 2026-08-21 comment, and I re-reproduced the open MAJOR at this head: with `-answer` on a stateless connection, a tool that completes without `input_required` lets the probe exit 0 with `rep.Elicited` empty — the supplied answer is never exercised, so the gate guarantees nothing. Fix: after `runCall`, fail when `opts.Answer != ""` and no elicitation occurred, plus a regression with a stateless tool that completes immediately. Everything else from rounds 1–2 remains resolved; package suite green under `-race`; CI green. This is the only remaining blocker.

*This review was performed with the assistance of Claude Fable 5.*

cmd/loom-mcp-probe/main.go
ilsiepotamus: Fixed in 95eac806 + dfa5474b: -watch now requires the acknowledgment before success, fails on sub.Err()/no-ack-deadline, and a watch skipped on a legacy connection is a probe failure, not a silent skip (flags that verify nothing are rejected).

cmd/loom-mcp-probe/main.go
ilsiepotamus: Fixed in 95eac806 + 8385596e: -timeout now bounds the whole probe (initialization, tools/list, calls, and subscription establishment), not just server/discover; flag help reworded to match.

cmd/loom-mcp-probe/main.go
ilsiepotamus: Fixed in 95eac806: executable and arguments are now separate (-cmd plus repeatable -arg), no strings.Fields parsing, and an empty -cmd is validated before use. Docs examples updated to the -cmd/-arg form in 8385596e.

docs/guides/mcp-probe.md
ilsiepotamus: Addressed in db8bce6a — and by events: #328 merged to main on 2026-08-20 (bc13fa0b), so the dual-era endpoint and §10 conformance the guide describes are now current rather than pending.

cmd/loom-mcp-probe/main.go
ilsiepotamus: Fixed in 95eac806: HTTP headers now come from a config-file source (same shape the manager uses), so authenticated endpoints work without exposing secrets in process arguments.

cmd/loom-mcp-probe/main.go
ilsiepotamus: Fixed in 95eac806: -answer JSON is validated via buildMRTRHandler (main.go:171) before buildTransport (main.go:176), so no stdio subprocess starts on invalid input.

cmd/loom-mcp-probe/main_test.go
ilsiepotamus: Fixed in 95eac806: server handlers return http.Error instead of require.*; assertions run in the test goroutine.

docs/guides/mcp-probe.md
ilsiepotamus: Fixed in 95eac806: the overbroad claim is gone — the guide now distinguishes the runtime probe (shipped client/transports, real servers in the manual sessions) from the scripted HTTP fixtures used by the automated tests.

cmd/loom-mcp-probe/main.go
ilsiepotamus: Fixed in dfa5474b: -answer now fails the probe if the run completes without any elicitation being exercised (rep.Elicited empty = error), with probe-level gating fixtures covering the complete-without-input_required path.

cmd/loom-mcp-probe/main.go
ilsiepotamus: Correction on the mechanism: it is -headers-env — the flag names an env var holding a JSON object of headers (headersFromEnv, main.go:402), so tokens ride the environment and never appear on the command line. Not a config file as I wrote above.

ilsiepotamus:
The remaining MAJOR is fixed in `6cc740ac`.

After `runCall`, the probe now fails when `-answer` was supplied but the call completed without ever eliciting (`rep.Elicited` empty): `-answer requested, but tools/call <tool> completed without input_required: the answer was never exercised`. Same rule family and wording as the round-1 `-answer`-on-legacy gate, so a gate that verifies nothing can no longer exit 0.

Regression added first (failing at the old head, exactly your repro — exit 0 with `Elicited` empty), then the fix: `TestProbeAnswerWithoutElicitationIsError` uses a stateless scripted server whose tool completes immediately. The legitimate paths stay green: `-answer` with a tool that does elicit (`TestProbeMRTRElicitation`) and stateless without `-answer` (`TestProbeStatelessServer`).

Validation: `gofmt` clean, `go build -tags fts5 ./...`, full `go test -tags fts5 -race ./cmd/loom-mcp-probe/...` (18 tests) green.

Ed-Jaramillo (approved):
Re-review at `6cc740ac`. **Approving** — the last remaining blocker is genuinely fixed, and I verified it independently rather than taking the commit message for it.

### Blocker verified fixed: `-answer` with a call that never elicits

Red-then-green, checked separately: I applied **only** the new test file to the previous head `dfa5474b` and it fails there with exactly the symptom I reported (`An error is expected but got nil` — the probe exited 0 with `rep.Elicited` empty). At `6cc740ac` the full package is green:

```
go test -tags fts5 -race -count=1 ./cmd/loom-mcp-probe/   →  ok (18 tests)
```

So `TestProbeAnswerWithoutElicitationIsError` is a real regression test, not a test that passes for unrelated reasons.

The fix itself (`cmd/loom-mcp-probe/main.go:244`) is placed correctly — inside the `-call` block immediately after `runCall`, in the same rule family and wording as the legacy-connection gate at `:214`. I checked it can't produce a false failure: `rep.Elicited` is appended only by the MRTR handler (`:375`), and the paths that would leave it empty despite MRTR activity (refused sampling/roots input requests) already return an error out of `runCall`, so `len(rep.Elicited) == 0` after a successful call really does mean the answer was never exercised.

### No collateral damage

- 17 additive lines, no behavior change to any other path.
- `go build -tags fts5 ./...` clean, `gofmt -l cmd/loom-mcp-probe` clean.
- All CI checks green at this head, including MCP SDK Interop and Unit Tests.
- Spot-checked that rounds 1–3 remain resolved: the `-watch` gate, the pre-ack watchdog sharing `-timeout` (`:290`), the `-cmd`/`-arg` argv form, and the up-front orphan-flag validation. Confirmed the last one still fires before any transport is built — `-url … -answer '{"x":1}'` with no `-call` exits 2 with `-answer requires -call`.

### Non-blocking nit — the runtime `-answer` gates aren't in the guide

`docs/guides/mcp-probe.md:137` documents only the flags rejected **up front**. Neither runtime `-answer` failure — legacy connection (`:214`) or a call that never elicits (`:244`) — appears anywhere in the guide. Someone writing a CI gate around the probe can't learn from the docs that `-answer` against a non-eliciting tool is a probe failure by design; today they'd discover it from the error message. One bullet in Limitations, alongside the existing "a silently ignored flag would be a gate that verified nothing" line, would close the loop. Fine to fold into a later docs pass — not worth another round on its own.

Nice work on the four rounds; the gating story is coherent now.

*This review was performed with the assistance of Claude Opus 5.*

PR #337, opened 2026-08-20
fix(mcp): surface MCP server instructions to agents (#336)

### LINKED ISSUE #336: MCP client drops server instructions; agents never receive server guidance
## Summary

MCP servers publish usage guidance in `InitializeResult.instructions` — the spec's mechanism for a server to tell the model how to use its tools. Loom's MCP **client** throws this away: `pkg/mcp/client/client.go` retains `result.ServerInfo` and capabilities from initialize but never stores `result.Instructions`, and nothing in the manager, adapter, or agent integration consumes them. Agents that register a server's tools never see the server's guidance.

Asymmetrically, loom's own MCP **server** implementation does send instructions (`pkg/mcp/server.WithInstructions`, served at initialize) — we emit them, we just don't listen to them.

## Observed (2026-08-20, main @ 6702edc4)

teradata-mcp-v2 publishes server instructions directing agents to load its syntax library (`get_syntax_help`) before writing native-operator SQL. With the server registered in the MCP manager and an agent given `tools: mcp: [{server: teradata, tools: ["*"]}]`, all 12 tools register and work — but the agent starts writing SQL cold, because the guidance never reached it.

The only workaround is duplicating each server's guidance into every agent's system prompt by hand. That defeats the design: the server owns its usage contract, and hand-copied guidance goes stale the moment the server updates (e.g. new syntax topics, changed session-handle semantics).

## Root cause

`Client.initializeWithVersion` (pkg/mcp/client/client.go:253–276) decodes `protocol.InitializeResult` — which has the `Instructions` field (pkg/mcp/protocol/types.go:38) — and keeps only `ServerInfo` and capabilities. There is no accessor, so downstream consumers (`pkg/mcp/manager`, `pkg/mcp/adapter`, `pkg/agent/mcp_integration.go`) could not surface instructions even if they wanted to.

## Proposed fix

1. **Client**: store `result.Instructions` at initialize; add `Client.Instructions() string` (mirroring `ServerInfo()`).
2. **Agent integration**: when an agent registers tools from an MCP server (`RegisterMCPTool`, the `RegisterMCPServer` wildcard path, and config-based `RegisterMCPTools`), attach that server's instructions once (dedup per server name) to the agent's system context — the same auto-append pattern graph-memory and workflow-communication instructions already use — under a header naming the server, e.g. `## Instructions from MCP server "teradata"`.

Agents then follow server guidance with zero prompt engineering, and guidance updates ship with the server.

🤖 Found while wiring teradata-mcp-v2 into loom agents via the MCP manager.

ilsiepotamus:
Fixes #336.

MCP servers publish usage guidance at handshake time — the spec's channel for a server to tell the model how to use its tools. Loom's client parsed the handshake and dropped that field on the floor, so agents registering a server's tools never saw its guidance; the only workaround was hand-copying server guidance into every agent's system prompt, which goes stale the moment the server updates.

## Changes

**Client** (`pkg/mcp/client`): store instructions from both handshake shapes — legacy `InitializeResult.instructions` (initialize) and stateless 2026-07-28 `DiscoverResult.instructions` (server/discover) — and expose them via a new `Client.Instructions()` accessor, mirroring `ServerInfo()`.

**Agent** (`pkg/agent`): all three MCP registration paths (`RegisterMCPTools`, `RegisterMCPServer`, `RegisterMCPTool`) now record the server's guidance when registering its tools. It renders into the system prompt as a `## Instructions from MCP server "<name>"` block via the same supplement pattern as the task-board/skills/workflow blocks. Set-once per server (handshake data is fixed for a connection's life), servers sorted for byte-stable ROM, empty/whitespace guidance is a no-op, thread-safe.

## Verification

- New table-driven tests (all `-race`): attach/render matrix (single server, sorted multi-server, empty-guidance no-op, empty-name no-op, set-once semantics), byte-stability across calls, concurrent attach/render.
- Extended both connect-path tests so the fakes now send instructions and assert `Instructions()` round-trips — legacy initialize and stateless discover.
- **Live validation** against teradata-mcp-v2 (which publishes "always load syntax before writing native SQL" guidance): with all first-action choreography removed from the agent's YAML, the agent quoted the server's guidance section from its context and executed the server-mandated first step (`get_syntax_help(topic='sql-basics')`) before touching SQL — through the real manager-registered MCP tool path.
- `just check` green.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -1220,6 +1220,14 @@ func (a *Agent) getSystemPrompt(ctx context.Context) string {
 	// other agent — nothing outside weave attaches a workflow context.
 	basePrompt += a.workflowCommPromptSupplement()
 
+	// Append usage guidance from MCP servers whose tools this agent
+	// registered (InitializeResult.instructions). The server owns its usage
+	// contract; rendering it here means the guidance ships with the server,
+	// not with per-agent prompt engineering. Session-stable: servers are
+	// registered at agent construction, and instructions are fixed for the
+	// life of a connection.
+	basePrompt += a.mcpInstructionsPromptSupplement()
+
 	return basePrompt
 }
 
--- a/pkg/agent/mcp_instructions.go
+++ b/pkg/agent/mcp_instructions.go
@@ -15,6 +15,7 @@ package agent
 
 import (
 	"fmt"
+	"go.uber.org/zap"
 	"sort"
 	"strings"
 )
@@ -33,6 +34,23 @@ func (a *Agent) attachMCPServerInstructions(serverName, instructions string) {
 		return
 	}
 
+	// The string is server-controlled and lands in the ROM prompt slot,
+	// which pressure relief never sheds: unbounded input would let one
+	// hostile or buggy server starve every session's context permanently.
+	// The cap is generous — real-world instructions are well under 2 KiB.
+	trimmed := strings.TrimSpace(instructions)
+	if len(trimmed) > maxMCPInstructionsBytes {
+		zap.L().Warn("MCP server instructions truncated",
+			zap.String("server", serverName),
+			zap.Int("bytes", len(trimmed)),
+			zap.Int("cap", maxMCPInstructionsBytes))
+		trimmed = trimmed[:maxMCPInstructionsBytes] + "\n[instructions truncated by loom: size cap]"
+	}
+	// Escape line-start markdown headers so one server cannot fabricate a
+	// section attributed to another server (or to loom's own supplements);
+	// the rendered header below is the only unescaped one per server.
+	trimmed = escapeLineStartHeaders(trimmed)
+
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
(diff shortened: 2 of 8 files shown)


ilsiepotamus:
Fresh full review (independent reviewer + my verification): the fix itself is correct, era-complete, and deterministic — legacy `initialize.instructions` and 2026 `server/discover.instructions` both captured, the discover→legacy-negotiated path takes the operative handshake's value (verified against the vendored official schema, which explicitly blesses system-prompt placement), rendering is per-session ROM with no per-turn cache-busting, and multi-server ordering is sorted/attributed. Four findings, fixed in `7c60cefd`: (1) the server-controlled string was **unbounded on its way into the unsheddable ROM slot** — a hostile or buggy server returning a multi-hundred-KB `instructions` string would permanently starve every new session's context (pressure relief never sheds ROM); now capped at 8 KiB per server with truncation marker + warn log. (2) The body was un-fenced markdown, so one server could fabricate a section attributed to another server (`## Instructions from MCP server \"github\"`) — line-start headers are now backslash-escaped; the loom-rendered header is the only live one per server. (3) `RegisterMCPTools` attached instructions even with zero registered tools while `RegisterMCPServer` gated — now consistent. (4) The actual wiring line was deletable without any test failing — new tests pin instructions through `getSystemPrompt`, the size cap, and the spoofing escape. Known limitation left as documented behavior: post-session registration yields tools without instructions (frozen-ROM semantics; production paths all register at construction).

PR #335, opened 2026-08-20
fix(tools): reconcile, filter, and evict stale MCP tool index entries (#334)

ilsiepotamus:
Fixes #334.

The persistent tool index (`$LOOM_DATA_DIR/loom.db.tools`) kept MCP tool rows forever after their servers were removed from configuration. `tool_search` served them to agents with no liveness check, dynamic registration failed every call with `server not found`, and an agent could burn its entire client timeout cycling through dead tools. Observed live: 81 stale rows across 3 servers removed months earlier, while `tool_sources` correctly recorded 0 live MCP tools.

## The three seams, each closed

**1. Reconcile at index time** (`pkg/tools/registry`)
`IndexAll` was upsert-only. Indexers can now implement the optional `ReconcilingIndexer` interface, reporting per run:
- `KnownScopes` — every scope (MCP server) that still exists; rows outside these are orphans of removed servers and are deleted.
- `CompleteScopes` — scopes this run enumerated fully; rows in them not re-reported are removed tools and are deleted. A configured-but-unreachable server is *known* but not *complete*, so a transient failure never wipes its rows.

`BuiltinIndexer` and `MCPIndexer` implement it; plain `Indexer`s keep upsert-only behavior, so custom tools registered out-of-band via `RegisterTool` survive (covered by a test). The existing re-index calls after `AddMCPServer`/`DeleteMCPServer` in `pkg/server/mcp_management.go` now actually do what their comments always claimed. `IndexToolsResponse` gains `pruned_count` (field 7, additive), and `looms serve` logs `pruned_stale_tools` at startup.

**2. Filter at search time** (`Config.LiveMCPServers`)
An optional callback restricts MCP-sourced results in `Search` and `GetToolsByCapability` to servers that currently exist, covering the window between reconciliation runs. `cmd_serve` wires it to `manager.ListServers()`; with no MCP manager at all, every MCP row is hidden (nothing could execute them). `nil` callback = no filtering (library back-compat).

**3. Evict on failed dynamic registration** (`pkg/shuttle`)
`manager.GetClient` now wraps a `manager.ErrServerNotFound` sentinel (message text unchanged). The `mcpManagerAdapter`s translate it to `shuttle.ErrMCPServerNotFound` **only when the server is also absent from configuration** — a configured-but-disconnected server stays transient. On that sentinel, the executor evicts the index row via the new `ToolEvictor` optional interface, so a dead entry fails at most once instead of once per agent turn.

## Verification

[...]

Issue #334, opened 2026-08-20
Tool index serves stale MCP tools from removed servers; agents burn timeouts on 'server not found'


ilsiepotamus:
## Summary

The persistent tool index (`$LOOM_DATA_DIR/loom.db.tools`) keeps MCP tool entries forever, even after their MCP servers are removed from the configuration. `tool_search` serves these dead tools to the LLM with no liveness filtering, the agent calls them, and every call fails with `dynamic registration failed: … server not found`. An agent can burn its entire client timeout cycling through stale tools before it finds a working one.

Observed on v1.4.0, macOS, `looms serve` with a single live MCP backend (stdio).

## Impact (real session, 2026-08-20)

A freshly created agent backed by a live stdio MCP backend was asked to survey databases. `tool_search` returned stale tool names from servers removed months ago; the agent spent ~5 minutes on failing calls and the client hit its deadline (`DeadlineExceeded`). Log excerpts:

```
{"caller":"fabric/circuit_breaker.go:238","msg":"circuit_breaker_failure","error":"tool not found: base_databaseList (dynamic registration failed: failed to get MCP client for server teradata_mcp_streamable: server not found: teradata_mcp_streamable)"}
{"caller":"fabric/circuit_breaker.go:238","msg":"circuit_breaker_failure","error":"tool not found: teradata_execute_sql (dynamic registration failed: failed to get MCP client for server vantage-mcp: server not found: vantage-mcp)"}
{"caller":"fabric/circuit_breaker.go:238","msg":"circuit_breaker_failure","error":"tool not found: dba_databaseSpace (dynamic registration failed: failed to get MCP client for server teradata_mcp_streamable: server not found: teradata_mcp_streamable)"}
```

Meanwhile `loom mcp list` reports `No MCP servers configured.`

## Evidence from the index DB

```
$ sqlite3 ~/.loom/loom.db.tools "SELECT mcp_server, COUNT(*) FROM tools WHERE mcp_server != '' GROUP BY mcp_server;"
rowan-mcp|41
teradata_mcp_streamable|38
vantage-mcp|2          -- 81 tools across 3 servers that no longer exist

$ sqlite3 ~/.loom/loom.db.tools "SELECT name, indexed_at FROM tools WHERE mcp_server='teradata_mcp_streamable' LIMIT 3;"
dba_databaseSpace|2026-03-17T09:55:16-07:00   -- five months stale

[...]

PR #340, opened 2026-08-20
fix(mcp): restore server's original parameter names instead of blind camelCasing (#339)

### LINKED ISSUE #339: MCPToolAdapter camelCases tool arguments, corrupting calls to snake_case MCP servers
## Summary

`MCPToolAdapter` corrupts tool arguments for any MCP server whose input schemas use snake_case property names — which is most Go/Python servers. The adapter presents schema property names to the LLM in snake_case (`toSnakeCase`), then in `Execute` **unconditionally camelCases every parameter key** (`normalizeParametersToCamelCase`, pkg/mcp/adapter/shuttle.go:199) before calling the server, on the hard-coded assumption that "MCP tools expect camelCase".

For a server whose schema already is snake_case, the round trip mangles the keys: schema says `table_name` → LLM (correctly) sends `table_name` → adapter rewrites it to `tableName` → server rejects it. Single-word parameters (`sql`, `database`, `topic`) survive because both conversions are identity on them — which hides the bug until a multi-word parameter is used.

## Observed (2026-08-20, main, live run against teradata-mcp-v2)

Four concurrent agents, real workload. Every tool whose schema has only single-word params worked; **every `describe_table` call failed — 29/29** — because the server validates strictly (`additionalProperties: false`):

```
MCP_CALL_FAILED: invalid arguments for tool describe_table: invalid arguments:
  [(root): table_name is required
   (root): Additional property sessionHandle is not allowed
   (root): Additional property tableName is not allowed ...]
```

while the recorded agent input was correct:

```json
{"database":"data_scientist","session_handle":"tdsh_…","table_name":"titanic"}
```

The same mangling silently degrades servers that validate leniently: unknown camelCase keys are dropped, optional parameters (like `session_handle`, which carries Teradata session affinity) silently vanish, and behavior changes without any error.

## Root cause

Two blind global case conversions built on a false premise:
- `InputSchema()` (shuttle.go:153–170): renames all schema properties (and `required`) to snake_case for the LLM, discarding the mapping to the server's original names.
- `Execute()` (shuttle.go:197–199): renames all incoming params to camelCase — not "back to the original names", just camelCase, regardless of what the server's schema actually says.

## Proposed fix

Replace the blind conversion with an exact reverse map derived from the tool's own schema:

1. At adapter construction, build `paramNameMap: llm-visible (snake_case) name → server's original property name` from `tool.InputSchema.properties`.
2. In `Execute`, rename each incoming key via the map; keys not in the map pass through unchanged. Never invent a casing the schema doesn't contain.
3. Keep the LLM-facing snake_case presentation in `InputSchema()` (that part is fine) — it just must no longer be lossy.

This is correct for camelCase servers (map returns the original camelCase names — current behavior preserved), snake_case servers (map is identity), and mixed schemas.

## Relation to recent issues

Third in the series from wiring teradata-mcp-v2 into loom agents end-to-end: #334 (stale tool index), #336 (dropped server instructions), and this — the first bug on the actual argument path.

ilsiepotamus:
Fixes #339.

`MCPToolAdapter` presented schema properties to the LLM in snake_case, then **unconditionally camelCased every argument** before calling the server — corrupting calls to any MCP server whose input schemas are already snake_case (most Go/Python servers). `table_name` became `tableName`: strict servers (`additionalProperties: false`) reject the call outright; lenient servers silently drop the unknown key and behavior changes with no error. Only multi-word parameter names break, which hid the bug behind single-word tools (`sql`, `database`, `topic`).

Live trigger: a four-agent run against teradata-mcp-v2 where **29/29 `describe_table` calls failed** with `Additional property tableName is not allowed` while every single-word-param tool worked.

## Change

Replace the blind conversion with an exact reverse map derived from the tool's own schema:

- `buildParamNameMap` (at adapter construction): maps each LLM-visible snake_case name → the server's **original** property name, recording only names that actually differ. camelCase schema → exact restore map; snake_case schema → nil map (identity).
- `Execute` renames incoming keys via the map only; unmapped keys pass through unchanged. The adapter can no longer invent a casing the server's schema doesn't contain.
- The LLM-facing snake_case presentation in `InputSchema()` is unchanged — it just stops being lossy.
- Dead helpers `toCamelCase` / `normalizeParametersToCamelCase` removed; schema-cache keys now derive from the restored params.

camelCase servers keep today's behavior exactly; snake_case servers stop being corrupted.

## Verification

- New table-driven tests (all `-race`): `buildParamNameMap` matrix (nil schema, no properties, snake identity, camel mapping), `restoreParameterNames` (camel restore, snake pass-through, unknown-key pass-through, nil), and a camelCase round-trip test replacing the old one that asserted the buggy behavior.
- Existing schema-cache tests updated to the restored-params key derivation.
- **Live validation**: with this fix in the running server, `describe_table` against teradata-mcp-v2 returns clean column metadata through an agent — the exact call shape that failed 29/29 before.
- `just check` green (`✅ All checks passed! (matches GitHub CI)`).

Third in the series from wiring teradata-mcp-v2 into loom agents end-to-end: #334 (stale tool index → #335), #336 (dropped server instructions → #337), and this — the first bug on the argument path itself.

### DIFF
--- a/pkg/mcp/adapter/shuttle.go
+++ b/pkg/mcp/adapter/shuttle.go
@@ -128,10 +128,25 @@ func buildParamNameMap(inputSchema map[string]interface{}) map[string]string {
 		return nil
 	}
 	m := make(map[string]string)
+	ambiguous := map[string]bool{}
 	for original := range props {
-		if visible := toSnakeCase(original); visible != original {
-			m[visible] = original
+		visible := toSnakeCase(original)
+		if visible == original {
+			// An identity property claims its own name: a differently-cased
+			// sibling that collapses to it must not shadow it.
+			ambiguous[visible] = true
+			delete(m, visible)
+			continue
 		}
+		if _, taken := m[visible]; taken || ambiguous[visible] {
+			// Two original properties collapse to one visible name
+			// (e.g. tableName + table_name). Renaming would be a
+			// nondeterministic guess; ambiguous names pass through as-is.
+			ambiguous[visible] = true
+			delete(m, visible)
+			continue
+		}
+		m[visible] = original
 	}
 	if len(m) == 0 {
 		return nil
--- a/pkg/mcp/adapter/shuttle.go
+++ b/pkg/mcp/adapter/shuttle.go
@@ -91,15 +91,23 @@ type MCPToolAdapter struct {
 	serverName    string      // Used as backend identifier
 	uiResourceURI string      // From tool._meta.ui.resourceUri (MCP Apps)
 	logger        *zap.Logger // Structured logger (defaults to no-op)
+
+	// paramNameMap maps the LLM-visible property name (snake_case, as
+	// presented by InputSchema) back to the server's ORIGINAL property name
+	// from the tool's own schema. Execute renames via this map only — a key
+	// with no entry passes through unchanged, so the adapter never invents a
+	// casing the server's schema doesn't contain (issue #339).
+	paramNameMap map[string]string
 }
 
 // NewMCPToolAdapter creates a new adapter that wraps an MCP tool
 func NewMCPToolAdapter(client *client.Client, tool protocol.Tool, serverName string) *MCPToolAdapter {
 	adapter := &MCPToolAdapter{
-		client:     client,
-		tool:       tool,
-		serverName: serverName,
-		logger:     zap.NewNop(), // No-op by default; use SetLogger to enable
+		client:       client,
+		tool:         tool,
+		serverName:   serverName,
+		logger:       zap.NewNop(), // No-op by default; use SetLogger to enable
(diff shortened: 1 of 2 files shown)


ilsiepotamus:
Fresh full review at `955f4fc8`. The schema-derived restore map is the right fix and `InputSchema()` provably presents via the same `toSnakeCase`, so map and presentation stay consistent. One real edge found and fixed in `e0f1c77d`: two original schema properties that collapse to one snake_case name (`tableName` + `table_name`) restored nondeterministically (map iteration order picked the winner, so the sent parameter name could differ across process restarts). Ambiguous names are now excluded from the map — they pass through unchanged, deterministically, and an identity property can never be shadowed by a differently-cased sibling. Regression added; `-race` suite and lint green.

PR #342, opened 2026-08-20
fix(server): wire registry subsystems on single-provider servers

ilsiepotamus:
## Problem

Graph-memory store, task manager, and tool-surface policy were injected into the agent registry **only inside the `providerPool != nil` branch** of the serve wiring (`cmd/looms/cmd_serve.go`). On a **single-provider server** (no provider pool configured), that branch never runs — so every registry-built agent got a **nil graph-memory store**:

- gRPC-created agents (`CreateAgentFromConfig`)
- config-hot-reloaded agents
- benchmark `--isolate` temp agents

With a nil store, graph-memory **extraction silently no-ops** and **cross-session recall returns nothing**. Statically-loaded agents were unaffected (they get `WithGraphMemoryStore` at load time), which is why this hid for so long.

## How it surfaced

In the LongMemEval **multi-session** benchmark, `--isolate` creates a fresh registry-built agent per question. Every one answered "I don't know" → **6.2%** (the no-memory blind floor). A clean two-session repro confirmed: extraction wrote nothing for isolate/registry agents, while a *statically-loaded* agent with the same config extracted and recalled correctly.

## Fix

Move the three registry-subsystem injections out of the pool-gated branch into `wireRegistrySubsystems`, called **unconditionally** whenever a registry exists. `SetProviderPool` and the ABTest judge/eval wiring stay pool-gated (they genuinely need the pool).

The helper takes an interface seam (`registrySubsystemSink`) so a spy test verifies all three setters fire with **no provider pool involved** — the regression guard. The helper signature has no pool parameter at all, which is the point.

## Verification

- **Unit:** `TestWireRegistrySubsystems` passes with `-tags fts5 -race`; full `cmd/looms` package green with `-race`.
- **Live, end-to-end:** on a single-provider (Azure gpt-4o) server, a hot-reloaded registry agent populated its graph (entities + memories persisted) and a **fresh session recalled the stored facts**. The LongMemEval gpt-4o multi-session subset went **6.2% → 66.7%** (back in the published full-history band) with only this change.

## Notes / follow-ups (not in this PR)

- `database.path` is silently ignored by the storage backend when `storage.sqlite.path` carries its default — data lands in `~/.loom/loom.db` regardless. Separate config bug, worth a follow-up.

### DIFF
--- a/cmd/looms/cmd_serve.go
+++ b/cmd/looms/cmd_serve.go
@@ -373,6 +373,53 @@ type judgeServerSetter interface {
 	SetJudgeServer(js *server.JudgeServer)
 }
 
+// registrySubsystemSink is the subset of *agent.Registry that receives the
+// server-level subsystems. Extracted so wireRegistrySubsystems is unit-testable
+// with a spy.
+type registrySubsystemSink interface {
+	SetTaskManager(manager *task.Manager, decomposer *task.Decomposer)
+	SetGraphMemoryStore(store memory.GraphMemoryStore, embedder memory.Embedder)
+	SetSuppressedBuiltinTools(names []string)
+}
+
+// wireRegistrySubsystems injects the server-level subsystems (task manager,
+// graph-memory store, tool-surface policy) into the agent registry so that
+// registry-built agents — gRPC-created, config-hot-reloaded, or benchmark
+// --isolate temp agents — receive them.
+//
+// This MUST run independently of whether a provider pool is configured. These
+// injections previously lived inside the `providerPool != nil` branch of the
+// serve wiring, so a single-provider server silently gave every registry-built
+// agent a nil graph-memory store (no extraction, no cross-session recall), a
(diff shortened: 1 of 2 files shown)


ilsiepotamus:
Fresh full review at the current head: **no findings**. Verified the old pool-gated injections are fully removed with exactly one new call site, `wireRegistrySubsystems` runs unconditionally when a registry exists (the regression this fixes), and the spy test genuinely covers both the wired and skipped paths. CI green. Merge-ready from my side.

ilsiepotamus:
Closing as a duplicate of #321, which fixes the same bug (registry task-manager/graph-memory/suppressed-tools wiring trapped inside `if providerPool != nil`) and is more complete — it also hoists the judge-fallback and eval-store wiring out of the pool gate. #321 predates this and is already CI-clean. The single-provider graph-memory fix was verified live end-to-end during LongMemEval (gpt-4o multi-session 6.2% → 66.7%); that verification applies equally to #321's version.

PR #344, opened 2026-08-20
feat(mcp): park failed tool calls on a linked resource, wake on resources/updated (#343)

### LINKED ISSUE #343: MCP adapter: park failed tool calls on a linked resource and wake on resources/updated (no client retry loops)
## Summary

When an MCP tool call fails on a transient resource-exhaustion condition (e.g. a session-handle budget), agents fall into client-side retry loops: a live 64-agent run against teradata-mcp-v2 produced **951 connect calls with 935 budget-full rejections** — one wasted round-trip *and one wasted LLM turn* per retry.

The 2026-07-28 machinery to eliminate this exists on both sides but is not connected: servers can mark a resource subscribable and push `notifications/resources/updated` when the condition clears (teradata-mcp-v2 now does, via its `teradata://session-handles` resource), and loom's client already implements `subscriptions/listen` with per-URI `ResourceSubscriptions` filters (PR #327). Nothing in loom *listens* and turns the notification into agent-visible progress.

## Proposal: park-and-wake in the MCP tool adapter, via a spec-native convention

**Convention (works for any spec-following server, zero server-specific code):** a failed `CallToolResult` whose content includes a `resource_link` is declaring "this resource's next update may clear the failure." On seeing that, `MCPToolAdapter.Execute`:

1. opens a `subscriptions/listen` stream filtered to that URI (stateless 2026-07-28 connections; legacy connections keep the plain error),
2. parks the in-flight tool call until `notifications/resources/updated` arrives (bounded by a wait budget),
3. retries the call; repeats until success or budget exhausted, then returns the last error.

The agent sees one tool call that succeeds late — no retry loop in the prompt, no polling, no extra turns. Servers opt in by attaching a resource link to their retryable errors, which is plain MCP.

Also needed: `protocol.Content` currently drops `resource_link` content (no top-level `uri` field), so the link a server attaches is invisible to loom today.

## Scope

- `pkg/mcp/protocol`: parse `resource_link` content (add `URI`/`Name` fields).
- `pkg/mcp/adapter`: park-and-wake in `Execute` with a bounded wait budget; per-adapter default, disabled for legacy connections.
- Tests: URI-detection unit tests + an end-to-end test with loom's own MCP server (which already supports server-side subscriptions, PR #328): failing tool links a resource → update notification → parked call retries and succeeds.

🤖 Motivated by the teradata-mcp-v2 slot-availability feature and a 64-agent live A/B.

ilsiepotamus:
Fixes #343.

A tool failure on a transient exhaustion condition previously forced agents into client-side retry loops — a 64-agent live run against teradata-mcp-v2 produced 951 connect calls with 935 `session_handle_budget_full` rejections: one wasted round-trip and one wasted LLM turn per retry.

## The convention (any spec-following server)

A failed `CallToolResult` whose content includes a **`resource_link`** declares *"this resource's next update may clear the failure."* The MCP tool adapter now parks such a call on a `subscriptions/listen` stream filtered to that URI (2026-07-28 stateless connections) and retries when `notifications/resources/updated` arrives — bounded by a 60s budget capped to the caller's deadline. The agent sees one tool call that succeeds late: no retry loop, no polling, no extra turns. Anything that prevents waiting (no link, legacy connection, subscribe failure) degrades to the pre-existing error unchanged.

Servers opt in with plain MCP — teradata-mcp-v2 already attaches the link on budget-full errors (`teradata://session-handles`).

## Seam fixes required en route

- `protocol.Content` now parses `resource_link` (its `uri`/`name` were silently dropped).
- `Client.CallTool` returns a typed `ToolResultError` preserving the full error result (content was flattened to a string, destroying any link). `Error()` renders the historical message byte-for-byte, so string-matching callers and analytics see no change.

## Verification

- `-race` tests: park→notify→retry→success end-to-end over a scripted stateless transport (asserts exactly one retry and one listen stream), no-link fast-fail with zero subscriptions opened, park-timeout surfacing the original error, and the `ToolResultError` matrix (message compat + link extraction).
- `just check` green (matches CI).

### DIFF
--- a/docs/guides/mcp-tool-call-lifecycle.md
+++ b/docs/guides/mcp-tool-call-lifecycle.md
@@ -1,0 +1,198 @@
+# MCP Tool-Call Lifecycle: Park-and-Wake and Session-Handle Auto-Release
+
+**Version**: v1.4.0
+**Status**: ✅ Implemented
+
+Two runtime behaviors of Loom's MCP tool adapter that manage the lifecycle of
+a tool call beyond the single request/response exchange:
+
+- **Park-and-wake** (issue #343): a failed tool call that links a resource
+  waits for that resource to update, then retries — one agent-visible call
+  instead of a retry loop.
+- **Session-handle auto-release** (issue #345): session handles minted by MCP
+  tools during a conversation are released by the runtime when the
+  conversation ends, instead of relying on the agent to clean up.
+
+Both behaviors are conventions the **server opts into**. A server that does
+not follow either convention sees exactly the pre-existing behavior.
+
+## Table of Contents
+
+- [Park-and-Wake](#park-and-wake)
+  - [The Convention](#the-convention)
+  - [Requirements and Budget](#requirements-and-budget)
+  - [How a Wait Ends](#how-a-wait-ends)
+  - [What Does NOT Trigger a Park](#what-does-not-trigger-a-park)
+- [Session-Handle Auto-Release](#session-handle-auto-release)
+  - [The Convention (Schema-Gated)](#the-convention-schema-gated)
+  - [Release Semantics](#release-semantics)
+  - [Scope: Per Chat, Not Per Session](#scope-per-chat-not-per-session)
+- [Observability](#observability)
+- [Limitations](#limitations)
+
+## Park-and-Wake
+
+A tool failure on a transient exhaustion condition (for example, a
+session-handle budget that is temporarily full) previously forced agents into
+client-side retry loops: one wasted round-trip and one wasted LLM turn per
+retry.
+
+### The Convention
+
+A server declares "this resource's next update may clear the failure" by
+including a **`resource_link`** content item in a failed `CallToolResult`
+(`isError: true`):
+
+```json
+{
+  "isError": true,
+  "content": [
+    { "type": "text", "text": "session-handle budget full" },
+    { "type": "resource_link", "uri": "example://session-handles", "name": "availability" }
+  ]
+}
+```
+
+When Loom's MCP tool adapter sees this, it parks the call: it opens a
+`subscriptions/listen` stream filtered to that URI and retries the original
+call when `notifications/resources/updated` arrives for it. The agent sees
+one tool call that succeeds late — no retry loop, no polling, no extra turns.
+
+The trigger is **`resource_link` content only**. An embedded plain `resource`
+content item in an error result is treated as payload (for example,
+diagnostic data), not as a retry condition, and never triggers a park.
+
+### Requirements and Budget
+
+- Requires a **2026-07-28 (stateless) connection** — the wait rides on
+  `subscriptions/listen`. Legacy connections surface the error unchanged.
+- The wait is bounded by a **60-second budget**, further capped by the
+  caller's context deadline (minus a 1-second margin). When the budget is
+  exhausted the original error surfaces.
+- Anything that prevents waiting — no linked resource, a legacy connection,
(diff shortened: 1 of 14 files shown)


ilsiepotamus:
Fresh full review (independent reviewer + my verification, key regressions re-run): the park-and-wake half held up under adversarial reading — demux registered before send, no send-on-closed panics, budget math correct, byte-for-byte error-message compat — but the **session-handle half had two independent ways to silently no-op, with no observability to reveal either**, and both activation conventions were too broad. All fixed in `7da959b4`:

- **Release could never work against half the servers** (F1/F2): `ReleaseAll` bypassed the adapter's param naming (its wire spelling differed from agent-issued calls — the PR's own tests encoded the divergence), and client-side schema validation rejected the bare one-arg call for any mint tool with required params — every auto-release failing silently, leaking exactly the handles #345 exists to release. The release path now derives the property name from the mint tool's own schema (`releaseHandle` or `release_handle`, whichever the server declared — no dependency on #340's helper either way) and only fires when required ⊆ {release property}, else WARN-skip to pre-PR leak-to-TTL.
- **Tracking is now schema-gated** (F6): a permissive server whose results merely contain a `session_handle` string no longer gets convention calls it never declared (which could mint fresh handles — the auto-release *creating* leaks). ⚠️ One architecture decision deliberately left open: release scope is still **per-`chat()`**, so a handle reused from conversation history in a later message will have been released — session-long lifetimes need a session-end hook that doesn't exist yet. Documented on the collector and in the new guide; not changed.
- **Production observability** (F3): real loggers wired at all five adapter constructor sites and the agent's ReleaseAll — park entry/exit, every retry outcome, every release attempt/failure/skip now logs with server/tool/handle fields (previously a 60s stall produced zero log lines).
- **Missed-wake window closed** (F4): one optimistic retry on subscription ack. **Park trigger narrowed to `resource_link`** (F5). **Ack honored-subset respected** (F7): a declining server fails fast instead of a full-budget stall. Multi-content mint results tracked (F8), `resource_link` uri/name preserved on the success path (F9), release budget 3s **total** and concurrent (F10, was 10s serial on the chat return path), an e2e test drives the real `Agent.Chat` wiring, and `docs/guides/mcp-tool-call-lifecycle.md` documents both conventions with limitations (F11).

Gates: build, `-race` across pkg/mcp + pkg/agent + pkg/shuttle (adapter `-count=3`), gofmt, golangci-lint — all green.

ilsiepotamus:
Fresh independent review of the rebased stack completed (all findings personally verified; -race -count=3 on the MCP packages green).

**Verdict: solid.** The park-and-wake protocol handling held up under adversarial reading — registration-before-send closes the subscribe/notify race (with the post-ack optimistic retry covering the residual window, tested), honored-subset fail-fast, clean degradation to pre-#343 behavior on every obstruction, no goroutine leaks, and re-invokes happen only on server-declared tool errors (never ambiguous transport failures), so no double-execution hazard. The rebase kept `restoredParams` consistent at every call site including inside both wait wrappers.

One finding, fixed in `00a3aa8f`: **HandleCollector keyed dedup on the bare handle string**, so identical handle strings minted by two different servers collided — the second mint was never tracked (TTL leak at conversation end) and a release through server B forgot server A's handle. Keys now carry the server identity on both `add` and `forget`; `TestHandleCollectorScopedPerServer` pins it.

Also noted for a separate issue (pre-existing path, not this PR's): MCP tools resolved via tool_search dynamic registration execute through `mcpToolWrapper`, which bypasses the adapter entirely — no resource park, no backpressure freeze, no handle tracking, no param-name restore.

PR #347, opened 2026-08-21
fix(llm): keyed shared rate limiters — honor per-agent rate_limit, scope to real quota boundaries

### LINKED ISSUE #346: LLM rate limiter: global singleton ignores per-agent rate_limit after first client; Bedrock-tuned defaults throttle every provider
## Symptom

Running 512 concurrent agents against Azure OpenAI (gpt-4o GlobalStandard, 1,500K TPM / 9,000 RPM quota), 274 of 512 died with:

```
rate limiter error: rate limiter queue timeout after 5m0s
```

Azure never throttled — the loom client did.

## Two compounding causes

**1. Defaults are Bedrock-tuned but apply to every provider.** `DefaultRateLimiterConfig()` in `pkg/llm/rate_limiter.go:64` is documented as "conservative defaults for AWS Bedrock" (2.0 RPS, 40,000 TPM) but is the zero-value backfill for openai, azure-openai, mistral, and gemini clients too. An Azure deployment with a 9,000 RPM quota gets throttled to 120 RPM client-side unless every agent YAML overrides it.

**2. The limiter is a process-global `sync.Once` singleton — first config wins, silently.** `getOrCreateGlobalRateLimiter` (e.g. `pkg/llm/azureopenai/client.go`) creates the limiter once per process with whichever agent's config happens to construct the first client. Every later agent's `spec.llm.rate_limit` is accepted by validation and then ignored. There is no log line saying so. Changing `rate_limit` in an agent YAML has no effect until the server restarts, and even then only if that agent's client is constructed first.

## Why it matters

- At high concurrency the queue drains at 2 req/s while `QueueTimeout` is 5m, so any backlog beyond ~600 queued requests is guaranteed death — the failure scales with fleet size.
- Per-agent `rate_limit` config is a documented proto field (`LLMRateLimitConfig`, agent_config.proto) that does not do what it says for the 2nd..Nth agent.

## Suggested direction

- Key the shared limiter per (provider, effective-config) or make it per-endpoint with the max of requested configs, instead of first-wins.
- Provider-appropriate defaults (or at least a startup log stating the effective limiter numbers and which agent's config produced them).
- When a later client requests a different config than the singleton holds, log a warning naming both.

## Repro

64+ concurrent conversations on any non-Bedrock provider with default `rate_limit`; watch `queue timeout after 5m0s` appear once the arrival rate exceeds 2 req/s for ~5 minutes.

ilsiepotamus:
Fixes #346.

## The bug

Every LLM provider package kept its rate limiter in a process-global `sync.Once` singleton: the **first client's config silently applied to every later client** for the process lifetime. Per-agent `spec.llm.rate_limit` was accepted by validation and then ignored. The defaults backfilling zero fields are Bedrock-tuned (2 rps / 40K TPM) but applied to every provider.

Observed at 512-way concurrency against an Azure gpt-4o deployment with a 9,000 RPM quota: the shared limiter ran at the 2 rps default, the queue drained at 120 requests/minute with a 5-minute `QueueTimeout`, and **274 of 512 agents died** with `rate limiter queue timeout after 5m0s`. Azure never throttled once.

## The fix

- **`llm.SharedRateLimiter(scope, config)`** (`pkg/llm/shared_limiter.go`): a keyed registry replacing the six per-package singletons (anthropic, azureopenai, bedrock×2 call sites, gemini, ollama, openai). Clients that draw on the same upstream quota share one token bucket; a different scope or different effective config gets its own limiter.
- **Scopes name real quota boundaries**: `azure-openai|endpoint|deployment`, `bedrock|region|model`, `openai|endpoint|model`, `anthropic|model`, `gemini|model`, `ollama|endpoint`. Two Azure deployments no longer fight over one bucket.
- **Normalization before keying**: an explicit value equal to the default shares the default's limiter (no accidental quota splitting).
- **Operator visibility**: limiter creation logs the effective numbers (scope, rps, tpm, burst, queue timeout); a second distinct config on the same scope logs a warning that the combined rate can exceed the quota.
- **Latent bedrock bug fixed on the way**: `getOrCreateGlobalRateLimiter` replaced the entire caller config with defaults whenever `Logger == nil`, discarding every override.
- anthropic keeps its provider-default merge (extracted to `mergedRateLimiterConfig`, now covered by the keyed registry).

## Not changed (deliberately)

The default numbers themselves (2 rps / 40K TPM) stay: raising them per provider is guesswork that could 429-storm existing deployments. With this PR the per-agent config actually works and the effective numbers are visible in logs, which is the safe fix. A per-provider default matrix can follow with data.

## Docs / proto

- `LLMRateLimitConfig.requests_per_second` proto comment now states the sharing semantics (swagger regenerated — comment-only diff).
- Five stale "global singleton / first client wins" claims corrected in `docs/reference/llm-{anthropic,gemini,ollama}.md`.

## Tests

`pkg/llm/shared_limiter_test.go`: same scope+config share one instance; explicit-default equals default; differing config split (the #346 repro); scope isolation; 32-goroutine race on one key; normalization backfill. Full `just check` green (lint, `-race` tests, build).

## Validation in the field

After patching all 512 gauntlet agents with `rate_limit: {requests_per_second: 100, tokens_per_minute: 1200000}` and restarting: 512 concurrent agents, **zero** queue-timeout deaths, zero Azure 429s (vs 274 deaths before).

### DIFF
--- a/docs/reference/llm-anthropic.md
+++ b/docs/reference/llm-anthropic.md
@@ -221,7 +221,9 @@ repeated system prompts or long conversations are detected.
 ## Rate Limiter
 
 ✅ The Anthropic client includes a built-in rate limiter (`pkg/llm.RateLimiter`) that is shared
-as a global singleton across all Anthropic client instances. The rate limiter:
+by every Anthropic client on the same model with the same rate-limit values (clients that set
+different values each get their own limiter, with a logged warning that the combined rate can
+exceed the quota). The rate limiter:
 
 - Automatically retries HTTP 429 (rate limit) responses with exponential backoff
 - Respects Anthropic's per-tier ITPM (input-tokens-per-minute) limits
--- a/docs/reference/llm-gemini.md
+++ b/docs/reference/llm-gemini.md
@@ -65,7 +65,7 @@ The Gemini provider connects Loom to Google's Gemini models through the Google A
 - Token-by-token streaming via Server-Sent Events (SSE)
 - Implicit caching (automatic since May 2025; cached tokens tracked in usage metadata)
 - Cost calculation per model
-- Client-side rate limiting (shared singleton across all Gemini client instances)
+- Client-side rate limiting (shared per model + rate-limit config)
 
 **Source code**: `pkg/llm/gemini/client.go`, `pkg/llm/gemini/types.go`
 
@@ -470,7 +470,7 @@ The `gemini-3.1-pro-preview` and `gemini-3.1-flash-lite-preview` catalog IDs are
 
 ## Rate Limiting
 
-The Gemini provider uses a **global singleton rate limiter** shared across all Gemini client instances in the process. This is initialized once via `sync.Once`.
+The Gemini provider uses a **shared rate limiter**: clients on the same model with the same rate-limit values share one limiter; a different model or different values gets its own (a warning is logged when two configs share one quota boundary).
 
 Rate limiting is optional and controlled by the `RateLimiterConfig` passed in the `Config` struct. When enabled via YAML:
 
@@ -564,7 +564,7 @@ error reading stream: unexpected EOF
 
 7. **No call IDs**: Gemini does not provide function call IDs in its response. The reversed tool name is used as the call ID.
 
-8. **Global rate limiter**: The rate limiter is a process-wide singleton. All Gemini client instances share the same rate limiter, initialized from the first client that enables it.
+8. **Shared rate limiter**: Clients on the same model with the same rate-limit values share one limiter. Differing values on the same model are honored separately (their combined rate can exceed the quota; a warning is logged).
 
 9. **Cached tokens and rate limits**: Gemini's implicit caching provides cost savings, but cached tokens still count against rate limits.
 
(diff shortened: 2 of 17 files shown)


ilsiepotamus:
Fresh full review: the keyed-registry design is right and honestly framed — config normalization before keying (explicit-default shares the default's limiter), the Logger correctly excluded from the key, config-splitting on one scope surfaced with an operator warning rather than silently first-wins, no leftover singletons anywhere in pkg/llm, and the 32-goroutine race test plus the #346 repro cover the core. One finding, fixed in `67700b55`: **the per-key-quota providers were missing the credential dimension** — Anthropic/OpenAI/Gemini quotas attach to the API key, so two agents with different keys (independent upstream quotas) shared one bucket keyed only by model/endpoint and would over-throttle each other (safe direction, wrong sharing). Their scopes now include `CredentialScope`, a short SHA-256 fingerprint — never the key itself — with the empty ambient credential fingerprinting identically so single-key deployments keep exactly the sharing this PR designed. Azure (per-deployment), Bedrock (region+model), and Ollama (endpoint) scopes were already correct. `-race` green across all of pkg/llm, lint clean.

PR #350, opened 2026-08-21
fix(llm,looms): attach rate limiter on the custom-LLM path; make dispatch concurrent

### LINKED ISSUE #348: looms custom-LLM path builds provider clients with no rate limiter (and azure branch has no env-var fallback)
## Symptom

512 agents launched against Azure OpenAI from a looms instance whose per-agent clients came from `createLLMProviderFromProtoConfig` (cmd/looms/cmd_serve.go:749). Result: an unthrottled thundering herd — Azure metered **~1.7M TPM in the first five minutes** (over the deployment's 1.5M quota), a 429 storm, and **503 of 512 agents dead within six minutes** (419 in a single minute). 11.06M tokens burned, mostly on the doomed burst.

The same fleet on a different machine, where this path failed and clients fell through to `pkg/agent/registry.go`'s path instead, was rate-limited normally (zero 429s).

## Two defects

1. **`createLLMProviderFromProtoConfig` passes no `RateLimiterConfig` for any provider** (anthropic, bedrock, ollama, openai, azure-openai, mistral, gemini). The zero value means `Enabled: false` → nil limiter → no pacing and no 429 retry (retry lives inside the limiter). Any agent whose client is built by this path bypasses client-side rate limiting entirely — silently.

2. **The azure-openai branch reads only `serverConfig.LLM.AzureOpenAIAPIKey` with no env fallback** (unlike the anthropic/openai branches, which fall back to their env vars). Combined with the viper gap where `llm.azure_openai_api_key` is never registered for env binding, an env-only key makes this path fail client creation — which then *accidentally* lands agents on the registry path that does have a limiter. Putting the key in looms.yaml (the documented fix for the env gap) is precisely what switches a fleet onto the unthrottled path. The safe behavior today is reached only through a failure.

## Why this is nasty

Which of the two client-construction paths serves an agent decides whether rate limiting exists at all, and the selection depends on where the API key happens to live. Nothing logs the difference.

## Suggested direction

- Thread the agent's `LLMRateLimitConfig` through `createLLMProviderFromProtoConfig` exactly as `registry.buildRateLimiterConfig` does (or better: converge the two construction paths — this is duplicated provider-switch logic that has already drifted).
- Add the env fallback to the azure branch for parity with anthropic/openai.
- Related: #346 / PR #347 fixed the limiter's first-wins singleton; this issue is about clients that never get a limiter at all. The serial dispatch of `processQueue` (one goroutine, MinDelay + full call latency per request) is a third, separate problem worth its own issue.

## Evidence

- 429 body: "Your requests to gpt-4o for gpt-4o in eastus2 have exceeded rate limit."
- Azure TokenTransaction: 2.79M + 5.85M tokens in the first two 5-min buckets, then ~450K/5min from the 9 survivors.
- Death timeline: 5 at 02:53, 419 at 02:54, 79 at 02:55 (launch 02:49).

### LINKED ISSUE #349: RateLimiter executes LLM calls serially — one goroutine runs each call to completion, capping fleet throughput at ~1 call/s
## Symptom

512 concurrent agents against Azure gpt-4o (1.5M TPM quota, client limiter set to 100 rps / 1.2M TPM): Azure metered only **88K TPM actual** — 6% of budget — while 88 agents died on `queue timeout after 5m0s` and 253 more hit their 40-minute deadlines. The configured rps/TPM were never the constraint.

## Cause

`RateLimiter.processQueue` (pkg/llm/rate_limiter.go:264) is a single goroutine that calls `processRequest` **synchronously**, and `processRequest` runs `executeWithRetry` — the full LLM HTTP round-trip, seconds long — before the next queued request is even dequeued. Effective fleet-wide throughput is:

```
1 request per (MinDelay + full LLM call latency) ≈ 0.2–1 call/s
```

regardless of `RequestsPerSecond`, `TokensPerMinute`, or `BurstCapacity`. Two aggravators:

- `MinDelay` defaults to 300ms and is backfilled even when the caller sets a high rps, and an explicit `min_delay: 0` cannot be expressed (zero means "use default").
- Because execution is serialized behind the queue, a big fleet converts the limiter into a global mutex around the provider.

## Fix direction

Dispatch should pace **admission**, not serialize **execution**: the queue loop waits for a request-bucket token (+ MinDelay as inter-admission spacing), then hands the call to a goroutine. Token-per-minute accounting stays as is. rps/TPM then behave as documented.

Related: #346 / PR #347 (first-wins singleton), #348 (custom-LLM path attaches no limiter at all). Evidence and measured constants are from the same 512-agent runs.

ilsiepotamus:
Fixes #348. Fixes #349.

## #348 — custom-LLM path builds unthrottled, retry-less clients

`createLLMProviderFromProtoConfig` (cmd/looms/cmd_serve.go) built every provider client with **no `RateLimiterConfig`**: nil limiter means no pacing and no 429 retry (retry lives inside the limiter). Its azure branch also read only `serverConfig` with no env fallback, so an env-only key failed this path and agents *accidentally* fell through to the registry path that does attach a limiter — where the API key lived silently decided whether a fleet was throttled.

Field evidence: 512 agents from a looms whose azure key was in YAML (→ this path) died 503/512 on real Azure 429s within six minutes (419 in a single minute), bursting ~1.7M TPM over a 1.5M quota. The identical binary and agent YAMLs with an env-only key (→ registry path) had zero 429s.

**Fix:**
- `pkg/agent.BuildRateLimiterConfig` exported; `Registry.buildRateLimiterConfig` delegates to it, so both construction paths share one semantics.
- `rlCfg` threaded through all nine provider branches (anthropic, bedrock, ollama, openai, azure-openai, mistral, gemini, huggingface, litellm).
- Azure branch gains `AZURE_OPENAI_{API_KEY,ENDPOINT,ENTRA_TOKEN}` env fallbacks for parity with the anthropic/openai branches and the registry path.

## #349 — serial dispatch capped fleets at ~1 call/s

`RateLimiter.processQueue` was one goroutine running each call **to completion** (MinDelay sleep + the full LLM round-trip) before dequeuing the next. Fleet-wide throughput ≈ 1 request per (MinDelay + call latency), regardless of `requests_per_second`/`tokens_per_minute`. Field evidence: 512 agents configured at 100 rps pushed only 88K TPM to Azure (Azure's own meter) while 88 died on `queue timeout after 5m0s` waiting to enter the tiny queue channel (capacity burst×2) and 253 more hit their 40-minute deadlines.

**Fix:** the dispatch loop now paces **admission** — request-token bucket, then MinDelay as inter-admission spacing — and hands execution to a goroutine. Documented semantics preserved: `RequestsPerSecond`/`BurstCapacity`/`MinDelay` gate request starts exactly as before; `QueueTimeout` unchanged; `Close()` still drains (the queue goroutine holds a waitgroup count while adding, so Add cannot race Wait at zero).

**Noted, deliberately not changed here:** `TokensPerMinute` was and remains metrics-only — the sliding window is reported in debug logs but never enforced. Making it a real gate is a behavior change deserving its own PR.

## Tests

- `TestRateLimiter_ConcurrentDispatch`: 8 × 200ms calls complete in <half the serial time (measured 0.21s vs 1.6s serial).
- `TestRateLimiter_AdmissionStillPaced`: 1 rps / burst 1 spaces the second start ~1s after the first.
- `TestCreateLLMProviderFromProtoConfigAzureEnvFallback` / `...YAMLKeyStillWorks`: key placement no longer breaks the path.
- `TestBuildRateLimiterConfigDefaults`: enabled-by-default, explicit disable, pass-through numbers.
- Full `just check` green (`-race` suites for pkg/llm, pkg/agent, cmd/looms; one unrelated npx-download conformance flake reproduced and cleared by cache warmup).

Relates to #346 / #347 (keyed shared limiters): independent branches; whichever merges second rebases trivially — #347 touches `NewRateLimiter`/client singletons, this touches dispatch and the cmd_serve construction path.

## Two more links in the same chain (added after field testing the branch)

**`spec.llm.rate_limit` was silently dropped by the YAML loader.** `LLMConfigYAML` had no field for it, so the proto `RateLimit` was never populated from agent YAML — every rate-limit block anyone ever wrote was discarded at load, and both construction paths fell back to provider defaults regardless of configuration. Now mapped in both YAML formats (legacy `agent:` and k8s-style `spec:`), with a round-trip test.

**HTTP 429 never reached the retry machinery.** The azureopenai client wrapped only the raw `httpClient.Do` in the rate limiter — which returns *nil error* for any HTTP status — so `executeWithRetry` never saw throttling and 429s went straight to the caller un-retried (field evidence: 503/512 agents dead on first-attempt 429s, zero "throttled, retrying" log lines). `callAPI` and `ChatStream` now build a fresh request per attempt and surface 429 as an error from inside the rate-limited call. Regression test: two 429s then success = three server calls, no error to the caller. The same wrap-the-raw-send pattern exists in other provider clients and should get the same treatment in a follow-up.

Field validation of the full branch: 512 concurrent agents, YAML-configured limits now provably reach the keyed limiter (creation log shows the configured 15 rps instance alongside the weaver's default), no thundering herd.

### DIFF
--- a/cmd/looms/cmd_serve.go
+++ b/cmd/looms/cmd_serve.go
@@ -823,6 +823,11 @@ func createLLMProviderFromProtoConfig(protoConfig *loomv1.LLMConfig, serverConfi
 
 	timeout := time.Duration(serverConfig.LLM.Timeout) * time.Second
 
+	// Rate limiting must be attached on THIS path too: a client built without
+	// it is unthrottled and retry-less (429 retry lives inside the limiter),
+	// which produced the 512-agent thundering-herd outage of issue #348.
+	rlCfg := agent.BuildRateLimiterConfig(protoConfig.RateLimit, logger)
+
 	switch protoConfig.Provider {
 	case "anthropic":
 		model := protoConfig.Model
@@ -835,11 +840,12 @@ func createLLMProviderFromProtoConfig(protoConfig *loomv1.LLMConfig, serverConfi
 			apiKey = os.Getenv("ANTHROPIC_API_KEY")
 		}
 		return anthropic.NewClient(anthropic.Config{
-			APIKey:      apiKey,
-			Model:       model,
-			MaxTokens:   maxTokens,
-			Temperature: temperature,
-			Timeout:     timeout,
+			APIKey:            apiKey,
+			Model:             model,
+			MaxTokens:         maxTokens,
+			Temperature:       temperature,
+			Timeout:           timeout,
+			RateLimiterConfig: rlCfg,
 		}), nil
 
 	case "bedrock":
@@ -851,15 +857,16 @@ func createLLMProviderFromProtoConfig(protoConfig *loomv1.LLMConfig, serverConfi
 		// caching SDK client and others to the Converse client (single source of
 		// truth for Bedrock client selection — see bedrock.NewClientForModel).
 		return bedrock.NewClientForModel(bedrock.Config{
-			Region:          serverConfig.LLM.BedrockRegion,
-			AccessKeyID:     serverConfig.LLM.BedrockAccessKeyID,
-			SecretAccessKey: serverConfig.LLM.BedrockSecretAccessKey,
-			SessionToken:    serverConfig.LLM.BedrockSessionToken,
-			BearerToken:     serverConfig.LLM.BedrockBearerToken,
-			Profile:         serverConfig.LLM.BedrockProfile,
-			ModelID:         modelID,
-			MaxTokens:       maxTokens,
-			Temperature:     temperature,
+			Region:            serverConfig.LLM.BedrockRegion,
+			AccessKeyID:       serverConfig.LLM.BedrockAccessKeyID,
+			SecretAccessKey:   serverConfig.LLM.BedrockSecretAccessKey,
+			SessionToken:      serverConfig.LLM.BedrockSessionToken,
+			BearerToken:       serverConfig.LLM.BedrockBearerToken,
+			Profile:           serverConfig.LLM.BedrockProfile,
+			ModelID:           modelID,
+			MaxTokens:         maxTokens,
+			Temperature:       temperature,
+			RateLimiterConfig: rlCfg,
 		})
 
 	case "ollama":
@@ -868,11 +875,12 @@ func createLLMProviderFromProtoConfig(protoConfig *loomv1.LLMConfig, serverConfi
 			model = serverConfig.LLM.OllamaModel
 		}
 		return ollama.NewClient(ollama.Config{
-			Endpoint:    serverConfig.LLM.OllamaEndpoint,
-			Model:       model,
-			MaxTokens:   maxTokens,
-			Temperature: temperature,
-			Timeout:     timeout,
+			Endpoint:          serverConfig.LLM.OllamaEndpoint,
+			Model:             model,
+			MaxTokens:         maxTokens,
+			Temperature:       temperature,
+			Timeout:           timeout,
+			RateLimiterConfig: rlCfg,
 		}), nil
 
 	case "openai":
@@ -886,11 +894,12 @@ func createLLMProviderFromProtoConfig(protoConfig *loomv1.LLMConfig, serverConfi
 			apiKey = os.Getenv("OPENAI_API_KEY")
 		}
 		return openai.NewClient(openai.Config{
-			APIKey:      apiKey,
-			Model:       model,
-			MaxTokens:   maxTokens,
-			Temperature: temperature,
-			Timeout:     timeout,
+			APIKey:            apiKey,
+			Model:             model,
+			MaxTokens:         maxTokens,
+			Temperature:       temperature,
+			Timeout:           timeout,
+			RateLimiterConfig: rlCfg,
 		}), nil
 
 	case "azure-openai", "azureopenai":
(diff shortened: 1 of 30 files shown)


ilsiepotamus:
Independent review + fix round completed (fresh-eyes review agent, findings verified and fixed).

**Review found, now fixed:**
- **Blocker**: the branch was cut from the commit immediately before #347 merged — on that base, per-agent `rate_limit` was dead code behind the old first-wins singletons, and merging as-is would have thrown custom-LLM fleets into the 2-rps #346 outage mode. Rebased onto main (clean); the composed keyed-limiter semantics re-validated with the full suite.
- **Major**: retries bypassed admission — under throttling, aggregate outbound = admitted RPS + synchronized no-jitter retry waves, violating the rate exactly when the upstream is throttling. Fixed in `31962fe1`: a throttled attempt now sleeps its backoff and **re-enters the admission queue** (paced like fresh work, exactly-once result delivery, ctx/stop escapes), with ±50% jitter and a 5-minute cap. `TestRateLimiter_RetriesReenterAdmission` pins inter-attempt spacing at the wire.
- **Major**: Retry-After was discarded — all retries could burn inside one throttle window. New `llm.ThrottleError{RetryAfter}` + `RetryAfterFromHeaders` (delta-seconds, HTTP-date, retry-after-ms, x-ratelimit-reset forms); the retry wait is `max(Retry-After, jittered backoff)`. Wired in azureopenai and anthropic; wire-level tests assert server-observed gaps.
- **Minors**: openai/gemini/ollama returned 429s as nil-error responses (limiter retry never fired) and built the request body once (a retry would resend a consumed body) — each now has a per-attempt `sendOnce` surfacing 429 as `ThrottleError` with the body rebuilt per attempt (mistral/litellm/huggingface inherit); byte-identical-second-attempt regressions per client. `waitForAdmission` now skips abandoned requests without consuming admission. `tokens_per_minute` is documented as observational-only everywhere it appears (it was never enforced; enforcement belongs to the slot scheduler).

Validation: `gofmt`/`buf lint` clean; `go build -tags fts5 ./...`; `-race -count=2` across pkg/llm and all provider packages + cmd/looms; golangci-lint 0 issues. The "429 retry lives inside the limiter" rationale is now actually true for every HTTP-shaped provider.

PR #351, opened 2026-08-21
docs(architecture): LLM slot scheduler — ending LLM starvation (design)

ilsiepotamus:
Design proposal for the deep redesign of LLM capacity management, written from the 2026-08 512-agent load campaign (R7a/R7b/az512b/az512c against Azure gpt-4o + teradata-mcp-v2).

**Thesis: waiting becomes a state, not an error.** Today an agent blocks in a FIFO queue with a 5-minute watchdog that kills the whole conversation while it holds Teradata handles and MCP slots. The design replaces that with:

- **SlotScheduler** per provider quota scope (the PR #347 scope keys): priority classes with a run-to-completion bias (`RESOURCE_HOLDER` > `IN_FLIGHT` > `NEW`), priority inheritance from held external resources, starvation aging, weighted fair share.
- **Park-and-wake** through the existing session-persistence suspend path (same shape as HITL suspends), with runtime-owned resource disposition — the direct generalization of the MCP handle park-and-wake, including the R3 lesson that parking without completion-driven churn starves.
- **Admission control at the door**: derived `max_active_conversations` per scope, visible door queueing, `RESOURCE_EXHAUSTED` + retry-after beyond that. Starve at the door for free instead of mid-task at high cost.
- **Layered capacity calibration**: provider ratelimit headers (Azure's full set verified empirically in-doc, incl. renewal periods), error-body hints (Gemini `retryDelay`), and an AIMD fallback for signal-free providers (Bedrock, Ollama). This also finally makes `tokens_per_minute` enforced — today it is metrics-only — with reservation-aware accounting (prompt + max_tokens, the thing that actually 429-stormed az512b).

Includes a `llm_scheduler.proto` sketch (proto is law — names final only after buf lint in the implementing PR), failure invariants (no death by waiting, bounded starvation, bounded inversion, quota-collapse behavior), observability spec, a 4-phase shipping plan on top of PRs #347/#350, open questions, and the measured constants appendix.

Doc only — no code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #352, opened 2026-08-21
feat(llm): LLM slot scheduler core — priority bands, park-and-wake, capacity calibration

ilsiepotamus:
Phase 2 core of the design in #351 (docs/architecture/llm-slot-scheduler.md). **Stacked on #347 and #350** — the diff includes their commits until they merge; review the top four commits here.

Motivated by the az512c field result: with every client-side rate-limit bug fixed (#346/#348/#349), 512 uniformly-admitted agents still produced **zero completed tasks** — 148 throttle-retry exhaustions plus fleet-wide handle contention. Nothing prioritized finishing, and completion is what releases resources.

## What this adds

**Proto (law first)** — `proto/loom/v1/llm_scheduler.proto`: `SlotPriorityClass` (RESOURCE_HOLDER > IN_FLIGHT > NEW), `SlotOrigin` bands (INTERACTIVE > BATCH), `LLMSchedulerConfig`, `SlotState`/`SlotWaiter`, and `LLMSchedulerService` (observability/admin only — slot grant itself is in-process, never RPC).

**`pkg/llm/scheduler`** — the scheduler core:
- `Acquire` **never fails for lack of capacity**: waiters park and are woken by completion churn, calibration, or aging; the only error it returns is the caller's `ctx.Err()`. One completion wakes bounded work — no herd re-forms.
- Priority dispatch with run-to-completion bias; new arrivals cannot overtake parked higher classes; starvation aging promotes one class per interval **within its band only** (liveness bound: top of band in ≤ 2× starvation_age).
- **Interactive band**: a human waiting on a response outranks the entire batch band. Edge-triggered per turn — the caller stamps INTERACTIVE only on the single conversation-loop turn answering a human action (chat message, HITL response), so human-started autonomous work never squats in the human lane. Batch admission is capped at `budget × (1 − interactive_headroom)` (default 15%), so an arriving human never waits behind a fully batch-reserved window.
- **Reservation-aware accounting** mirroring provider admission math: prompt + max_tokens debited at grant, trued-up to actuals at release (the exact gap that 429-stormed az512b).
- **Layered capacity calibration**: provider header telemetry outranks everything (`UpdateFromHeaders`); `Retry-After`/reset scheduled wakes; **provider-agnostic AIMD fallback** (`ObserveThrottle` halves, `ObserveSuccess` grows additively) for providers that state nothing — Bedrock, Ollama, header-stripping proxies.
- Registry (one scheduler per quota scope) + gRPC `Service` implementation.

**Capacity feed** — new `llm.CapacityObserver` seam; the azureopenai client harvests `x-ratelimit-limit/remaining/reset-tokens` on every response (header set verified live against our deployment), `Retry-After` on 429s, and drives AIMD when a gateway strips headers.

**Zero-value expressiveness fixed by construction** (the #346-lineage trap): `InteractiveHeadroom` 0 = default, negative = disabled.

## Tests (all `-race`)

Park/wake contract; ctx-only cancellation; cross-class dispatch order; no-overtake; aging promotion within band, never across; interactive NEW beats batch RESOURCE_HOLDER regardless of arrival order; batch parks at the headroom boundary while an identical interactive request passes; reservation true-up; header calibration precedence; AIMD; 64-goroutine race; Close-wakes-parked; end-to-end header harvest through a fake Azure. Full `just check` green.

## Deliberately not here (next PRs per #351 phasing)

`runConversationLoop` slot acquisition with real class/origin stamping, door admission control in looms, service registration, and durable park/persist/resume.


ilsiepotamus:
Wiring commit pushed: the scheduler is now live end to end behind `llm.scheduler_enabled` (default off).

- **loom CLI** stamps `loom-slot-origin: interactive` gRPC metadata only when both stdin and stdout are TTYs — humans ride the interactive band, scripted/piped fleets are batch. Edge-triggered per turn.
- **StreamWeave** installs the turn's `SlotInfo` from that metadata (per request, never conversation-lifetime).
- **`chatWithRetry`** — the single funnel every provider call passes through — acquires a slot per call and releases it with the response's actual usage (reservation true-up). Class: NEW → IN_FLIGHT → RESOURCE_HOLDER via the `MarkResourceHolder` seam.
- **Both azure construction paths** attach the scope's scheduler as `CapacityObserver`; the client's new `SchedulerScope()` keeps telemetry and arbitration on the same scope key.
- **`LLMSchedulerService`** registered on the looms gRPC server.

Tests `-race` green across pkg/agent, pkg/server, pkg/llm/scheduler, pkg/llm/azureopenai, cmd/looms; full `just check` green. Still deferred per #351 phasing: door admission, durable park/resume, MarkResourceHolder call sites in the MCP adapter (integration-branch follow-up).

ilsiepotamus:
Independent review + fix round completed (fresh-eyes review agent, all findings verified against the code, then fixed).

**Verdict on the core**: the park/wake machinery is sound — no lost wakeups, correct band/class ordering, leak-free grant accounting, race-clean at `-count=3`. All defects were in the calibration layer and the wiring; the core is untouched by the fixes.

**Blocking, fixed in `0129a576`:**
- **AIMD wedge**: `ObserveThrottle` halved per 429 *attempt* (client retry loops report each attempt; concurrent callers multiply) with a floor of 1 — three halvings put `effectiveTPM` below the smallest admissible reservation, and a scope that admits nothing observes nothing and can never repair itself. Now: one multiplicative decrease per congestion event (hold-off = max(Retry-After, 15s cooldown)), a floor that keeps the scope's smallest served reservation admissible, and a work-conserving escape valve — an idle scope (nothing reserved, empty window) always admits its head waiter, making the wedge structurally impossible. `TestWedgeImpossibleIdleScopeAlwaysAdmits` pins it with the ceiling forced to 1.
- **Reset-header freeze**: `UpdateFromHeaders` armed `nextWake` from the informational reset header on *every successful response*, hard-gating all admission (interactive included) for the reset duration — the PR's own Azure fixture (745K remaining, reset 30s) froze a healthy scope 30s per success, and `TestUpdateFromHeadersCalibrates` ended `_ = err`, asserting nothing. Now: reset gates only when remaining is actually exhausted (below the smallest reservation), with the local window realigned to the provider's renewal; the test asserts prompt admission on the healthy fixture, plus an inverse exhausted-gates-until-reset test.

**Majors, fixed in `0129a576` + `44e71cf4`:**
- Cancelled grants charged their full reservation as phantom usage → new `Grant.Cancel()` returns the reservation uncharged; mass-cancel leaves the window empty (test).
- Small lower-class waiters backfilled past a large starved top-class head indefinitely → once a head is past its aging point at top class, backfill below it halts until it fits (test: large RESOURCE_HOLDER vs a stream of small NEW).
- Only StreamWeave was scheduled; unary Weave bypassed *and poisoned* the shared ceiling → `installTurnSlotInfo` now runs at all four turn entry points, with `priorCalls` genuinely seeded from session existence (resumed conversations classify IN_FLIGHT).
- The TUI — the primary human surface — was banded BATCH → it now stamps `loom-slot-origin` per turn; the client-asserted trust model is documented on the proto.
- The "provider-agnostic AIMD" was wired to azureopenai only → observation moved to the agent's LLM funnel, driven by the typed `llm.ThrottleError`/clean-completion outcome, so **every provider's scope gets throttle/success observations with zero per-provider code**; Azure keeps only header calibration (2xx-only now), which outranks AIMD.

**Minors**: `SetSchedulerConfig` maps `interactive_headroom`; `SlotState.next_wake`/`SlotWaiter.waiting_since` populated and door-owned proto fields annotated honestly; scheduler logger installed before provider construction; `Close()` fails Acquire fast and keeps accounting consistent; stale comments corrected (no MCP MarkResourceHolder call site yet; no batch budget share — interactive may starve batch by design).

Validation: scheduler `-race -count=3`, all 14 pkg/llm packages + agent/server/cmd/looms/tui `-race`, `buf lint`/`generate`, gofmt, vet — all green.

PR #353, opened 2026-08-21
feat(scheduler): door admission control — queue batch turns at the front door

ilsiepotamus:
**Stacked on #352** (base branch is `feat/llm-slot-scheduler`); Phase-2b of the #351 design.

Starving a conversation at the door is free; starving it mid-task wastes held resources and partial work — az512c demonstrated the mid-task version (0/512 completions from uniform admission). This adds the front door:

- **`DoorGate`**: caps concurrently active batch conversation turns. Excess queues FIFO, and a released slot is handed *directly* to the next waiter — the active count never dips, so no release/enter race can overshoot the ceiling. Waiting is bounded only by the caller's context. The one capacity error, `ErrDoorFull`, fires *before* any waiting, when the door queue itself is at cap.
- **StreamWeave wiring**: batch turns pass the gate with a deferred release at turn end; a full door surfaces as `RESOURCE_EXHAUSTED` with retry guidance. **Interactive turns bypass entirely** — a human must never queue behind a fleet (the slot scheduler's interactive headroom protects their capacity; validated live: a pure-LLM human probe answered in 26s through 261-agent saturation).
- **Config**: `llm.max_active_conversations` (0 = gate off, the default) and `llm.max_door_queue` (0 = unbounded queueing).

Tests (`-race`): disabled transparency; ceiling + deterministic FIFO hand-off (a single release drives the whole admission cascade); queue-cap rejection; ctx-cancel leaves no queue leak including the admit/cancel race; idempotent release (a double release must not mint a phantom slot); 64-goroutine churn. Full `just check` green.

### DIFF
--- a/cmd/looms/cmd_serve.go
+++ b/cmd/looms/cmd_serve.go
@@ -1051,6 +1051,7 @@ func runServe(cmd *cobra.Command, args []string) {
 	// construction: clients attach their CapacityObserver at build time,
 	// and agents are loaded well before gRPC service registration.
 	llmscheduler.SetEnabled(config.LLM.SchedulerEnabled)
+	llmscheduler.SetDoorLimits(config.LLM.MaxActiveConversations, config.LLM.MaxDoorQueue)
 
 	// Export config values to environment variables for tools
 	exportConfigToEnv(config)
--- a/cmd/looms/config.go
+++ b/cmd/looms/config.go
@@ -1675,6 +1675,17 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("unsupported LLM provider: %s (must be anthropic, bedrock, ollama, openai, azure-openai, mistral, gemini, huggingface, or litellm)", c.LLM.Provider)
 	}
 
+	// Door admission knobs: a negative value is a configuration error, not a
+	// silent disable (max_active_conversations) or an unbounded queue
+	// (max_door_queue). Zero keeps its documented meaning: gate off /
+	// unbounded queueing.
+	if c.LLM.MaxActiveConversations < 0 {
+		return fmt.Errorf("llm.max_active_conversations must be >= 0, got %d (0 disables the door gate)", c.LLM.MaxActiveConversations)
+	}
+	if c.LLM.MaxDoorQueue < 0 {
+		return fmt.Errorf("llm.max_door_queue must be >= 0, got %d (0 means unbounded queueing)", c.LLM.MaxDoorQueue)
+	}
+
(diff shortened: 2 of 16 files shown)


ilsiepotamus:
Independent review + fix round completed (fresh-eyes review agent, findings verified, then fixed in `16de7d69`).

**Verdict on the primitive**: DoorGate is verifiably leak-free and FIFO-sound — mutex-serialized handoff keeps `active` exact through release/enter races, release is idempotent, ctx-cancel while parked frees the position, and it is fully generic (ctx-only, no backend/tool assumptions). All findings were at the wiring seam.

- **Blocker — unary `Weave` bypassed the door entirely**, so `max_active_conversations` was not a ceiling: the MCP bridge, the TUI's unary path, and grpc-gateway `/v1/weave` all drive unary Weave. Admission is now one helper (`enterTurnDoor`) called at **all four** turn entry points right after slot-info installation, interactive bypass inside the helper. `TestUnaryWeaveRespectsDoorCeiling` proves two concurrent unary Weaves against maxActive=1 serialize (fault-injection checked: removing the wiring fails the test).
- **Major — a door rejection was cached as a durable outcome** in the idempotency dedupe for 10 minutes, so clients following the prescribed same-key retry got the stale RESOURCE_EXHAUSTED instantly after the queue drained. `ResourceExhausted` is now transient in `isTransientOutcome`; `TestDoorRejectionNotCachedByDedupe` pins reject → drain → same-key retry succeeds.
- **Major — HTTP/SSE clients could never be INTERACTIVE** (no gRPC metadata on `r.Context()`), turning web humans into parkable batch. New `X-Loom-Slot-Origin` header maps into the stream context (same client-asserted trust model as the gRPC metadata, documented in docs/reference/streaming.md with the SSE docs).
- **Minors**: `GetSlotState` now populates `active_conversations`/`door_queue_depth` from the live DoorGate; rejections logged at Warn with counters, parks at Debug with `door_wait`; negative door knobs are startup errors in `Config.Validate()`; `SetDoorLimits`'s swap-wholesale constraint documented; test hygiene (no `require` in goroutines, readiness polling instead of sleeps, corrected accounting comment).

Validation: scheduler `-race -count=3`, server/cmd/looms/mcp suites `-race`, `buf lint`/`generate`, gofmt — all green.

Note for merge sequencing: this branch stacks on #352 (base = feat/llm-slot-scheduler), which stacks on #350 — merge #350 → #352 → #353, and retarget this PR before the base branch is deleted.

PR #355, opened 2026-08-21
feat(mcp): freeze tool calls on capacity backpressure — flow control never reaches the model

### LINKED ISSUE #354: Freeze agent tool calls on MCP capacity backpressure instead of surfacing retries to the model
az512e (512 gpt-5.2 agents vs one teradata-mcp server) measured the failure mode: 139/512 agents (27%) never obtained a session handle. Their connects met a full 256-handle pool, the budget-full error was surfaced to the model, and the models retried politely twice and gave up. Agents that did discover wait_s asked for 3600s and were silently clamped to 300s — far below the ~20-min median slot turnover.

Flow control must not be the model's job. The runtime owns the loop: when an MCP tool call fails with capacity backpressure, freeze the call below the LLM (the conversation just sees a slow tool call) and re-invoke until capacity frees or the agent's deadline expires.

Server side (teradata-mcp-v2, shipped): budget_full/server_busy/admission_timeout errors now carry a machine-readable contract — retryable, retry_after_s, wait_param, max_wait_s — and connect's wait_s cap is operator-configurable (--connect-max-wait-s).

Loom side (this issue, extends #343/#344): parse the contract into client.ToolResultError; in the MCP adapter, dispatch retryable failures to a freeze loop bounded by the ctx deadline — re-invoking with wait_param set to the remaining budget (clamped to max_wait_s) so each retry parks server-side in the FIFO wait queue and holds one queue position. Errors without the contract keep the #343 resource-link park; everything else surfaces unchanged.

ilsiepotamus:
Closes #354. Stacked on #344 (base: `feat/mcp-resource-wait-343`) — review only the top commit.

## Measured problem (az512e: 512 gpt-5.2 agents, one teradata-mcp server)

**139/512 agents (27%) never obtained a session handle.** Their connects met the full 256-handle pool, the budget-full error was surfaced to the model, and the models retried politely ~2.4 times and gave up. Agents that discovered `wait_s` from the error prose asked for 3600s and were silently clamped to the server's 300s cap — far below the ~20-min median slot dwell.

## Change

A tool failure carrying the machine-readable backpressure contract freezes in the MCP adapter — below the LLM — and re-invokes until it succeeds, fails differently, or the conversation's ctx deadline expires. The model sees a tool call that took longer, never the flow control.

```json
{"code": "session_handle_budget_full", "message": "…",
 "retryable": true, "retry_after_s": 412,
 "wait_param": "wait_s", "max_wait_s": 300}
```

- **`client.BackpressureHint`** — contract parsing on `ToolResultError`. Task-level failures (SQL errors, timeouts, deadlocks) never carry it and reach the model unchanged.
- **`adapter.awaitBackpressure`** — with a `wait_param`, each re-invoke passes the remaining ctx budget clamped to `max_wait_s`, parking the retry **server-side** in the FIFO wait queue: one parked call holds one queue position for the whole wait. Without one, polls with `retry_after_s` clamped to [1s, 30s].
- **Degraded-park guard** — a server that advertises parking but refuses instantly gets polling pace, never a hot spin.
- Contract-less failures keep #343's resource-link park; the circuit breaker is untouched (tool-level errors never reached it).

## Server half (shipped separately)

teradata-mcp-v2 `5a6b357`: `fail()` emits the contract on `budget_full`/`server_busy`/`admission_timeout`; new `--connect-max-wait-s` makes the wait ceiling operator-configurable (raise to agent scale for fleets).

## Tests

Scripted-transport tests (`-race`): server-park with cap clamping, ctx-deadline clamping, poll path timing, condition-change surfacing, budget exhaustion before ctx death, contract parsing table.

### DIFF
--- a/pkg/agent/mcp_integration.go
+++ b/pkg/agent/mcp_integration.go
@@ -38,6 +38,10 @@ type MCPServerConfig struct {
 	// AutoClose determines if the client should be closed when agent is done
 	// Default: false (client lifecycle managed externally)
 	AutoClose bool
+
+	// Logger receives registration events and is handed to every tool
+	// adapter (backpressure freezes, resource-wait parks). Nil = no-op.
+	Logger *zap.Logger
 }
 
 // RegisterMCPTools connects to an MCP server and registers all its tools with the agent.
@@ -75,12 +79,12 @@ func (a *Agent) RegisterMCPTools(ctx context.Context, config MCPServerConfig) er
 		return fmt.Errorf("MCP client must be initialized before registering tools (call client.Initialize first)")
 	}
 
-	// Get logger (use agent's tracer logger if available, otherwise create new one)
-	logger := zap.NewNop()
-	if a.tracer != nil {
-		// Tracer might have a logger we can use
-		// For now, just use nop logger
-		logger = zap.NewNop()
+	logger := config.Logger
+	if logger == nil {
+		// Fall back to the process-global logger, not a nop: adapter
+		// lifecycle events (park entry/exit, session-handle release
+		// attempts and skips) must stay visible in production logs.
+		logger = zap.L()
 	}
 
 	logger.Info("registering MCP tools",
@@ -98,9 +102,7 @@ func (a *Agent) RegisterMCPTools(ctx context.Context, config MCPServerConfig) er
 	tools := make([]shuttle.Tool, len(mcpTools))
 	for i, mcpTool := range mcpTools {
 		mcpAdapter := adapter.NewMCPToolAdapter(config.Client, mcpTool, config.Name)
-		// Adapter lifecycle events (park entry/exit, session-handle release
-		// attempts and skips) must be visible in production logs.
-		mcpAdapter.SetLogger(zap.L())
+		mcpAdapter.SetLogger(logger)
 		tools[i] = mcpAdapter
 	}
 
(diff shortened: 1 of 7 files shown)


ilsiepotamus:
Fresh independent review completed (findings personally verified; -race -count=3 green). The freeze mechanism itself checked out — budget rides the ctx deadline with sound margin arithmetic, hint refresh is deadline-capped, the anti-spin floor works, no leaks, and the circuit-breaker-untouched claim is verified true.

**One must-fix, fixed in `a2eed904`** (now `4f6306ed` after rebasing onto #344's hardening): `awaitBackpressure` injected `wait_s` into the caller's params map **in place**. When no schema case-restore happens — snake_case servers, i.e. teradata-mcp itself — that map IS the model's `toolCall.Input`, so the injected parameter leaked into the model's replayed tool history (imitation bait) and the persisted tool-execution record misattributed it as model-sent — the exact flow-control leak this PR exists to prevent. Re-invokes now use a private copy; `TestBackpressureCallerParamsNotMutated` pins the wire-carries-it/caller-never-sees-it invariant.

Also fixed from the review: a hint refresh that changes or drops `wait_param` now deletes the previously injected key (no stale server-side park stacked on client-side pacing); the degraded-park path honors the server's `retry_after_s` when it exceeds the 1s anti-spin floor (fleet-scale, fixed-1s polling against an already-refusing server was needless load); the branch-once mechanism handoff is documented as deliberate recursion bounding; and the `Logger()`/`NewManager` doc comments are untangled.

Acknowledged, tracked separately: the tool_search dynamic-registration wrapper bypasses the adapter (freeze, park, handle tracking, param restore) — pre-existing path, filed as its own issue.

PR #356, opened 2026-08-21
fix(memory): graph-memory recall survives fleet load

ilsiepotamus:
A 512-agent fleet run (az512h, 989 accumulated memories) measured recall **silently dead for every agent**: `accessed_at` never updated, zero context injections. Diagnosis and fixes:

1. **Keyword fallback for the query distillation.** `extractSearchQuery` is an LLM side-call with a 10 s timeout; under a saturated token pipe it starves and returned `""`, disabling recall for the conversation. It now falls back to the user message's content words (barewords, stopwords dropped, cap 12).
2. **FTS5 sanitization.** `toFTS5OrQuery` fed LLM/user text into `MATCH` raw — one apostrophe/comma/colon/'?' is a parse error both recall callers swallow as zero results. Free text is now tokenized to barewords; explicit operator queries pass through only with balanced quotes.
3. **Observability.** The path had no span and no log — a fleet ran with recall dead and nothing showed it. `injectGraphMemoryContext` now emits a `graph_memory.recall` span with `recall.outcome` at every gate, plus query/candidate/injected attributes.
4. Drive-by: `rerankMemories` wrote the user message **twice** into the rerank prompt (truncation branch appended instead of replacing).

Known-but-untouched (follow-ups): `min_salience_threshold`/`max_recall_candidates` are parsed but unused (limits hardcoded 50/30/5); `VectorRecall` has no callers (embeddings are write-only).

Tests: keyword-query table test + cap, FTS5 sanitization table test (punctuation, unbalanced quotes, operator pass-through); `-race` green on both packages.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -3162,6 +3162,13 @@ func (a *Agent) injectGraphMemoryContext(ctx context.Context, session *types.Ses
 	if a.graphMemoryStore == nil || a.graphMemoryConfig == nil || !a.graphMemoryConfig.Enabled {
 		return
 	}
+	// The recall path had zero observability: every gate below returned
+	// silently, which is how a fleet ran with recall dead and nobody knew
+	// (az512h). One span, outcome always set.
+	ctx, span := a.tracer.StartSpan(ctx, "graph_memory.recall")
+	defer a.tracer.EndSpan(span)
+	outcome := "unknown"
+	defer func() { span.SetAttribute("recall.outcome", outcome) }()
 
 	// Wait for any in-flight extractions to finish before querying.
 	// This ensures recently ingested content is available for recall.
@@ -3169,6 +3176,7 @@ func (a *Agent) injectGraphMemoryContext(ctx context.Context, session *types.Ses
 
 	budget := a.graphMemoryTokenBudget()
 	if budget <= 0 {
+		outcome = "no_budget"
 		return
 	}
 
@@ -3182,6 +3190,7 @@ func (a *Agent) injectGraphMemoryContext(ctx context.Context, session *types.Ses
 		}
 	}
 	if userMessage == "" {
+		outcome = "no_user_message"
 		return
 	}
 
@@ -3190,8 +3199,10 @@ func (a *Agent) injectGraphMemoryContext(ctx context.Context, session *types.Ses
 	// becomes something like "first issue with new car after first service".
 	searchQuery := a.extractSearchQuery(ctx, userMessage)
 	if searchQuery == "" {
+		outcome = "no_query"
 		return
 	}
+	span.SetAttribute("recall.query", searchQuery)
 
 	// Gather candidate memories from multiple sources.
 	seen := make(map[string]bool)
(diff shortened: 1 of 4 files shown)


ilsiepotamus:
Independent review found this PR **did not close the bug it claims to close, and introduced a regression**. Both verified by executing queries against a live migrated `porter unicode61` table rather than by reading. Fixed in `cf3ac859`.

**Blocking 1 — the bug survived.** The "already contains FTS5 operators, pass through" branch returned the string verbatim whenever it contained `" AND "` / `" OR "` / `" NOT "` / `" NEAR "` with balanced quotes. The input is always free text from an LLM or a raw user message, so ordinary prose took the raw path and every apostrophe, comma, `?` and paren became FTS5 syntax. Four reproduced errors:

```
Coach's AND player's stats             -> fts5: syntax error near "'"
first issue NOT resolved yet?          -> fts5: syntax error near "?"
car service AND GPS malfunction, March -> fts5: syntax error near ","
volatile table OR overflow (DECIMAL)   -> fts5: syntax error near "overflow"
```

That is the az512h signature — recall silently dead — still live on this branch. The pass-through is now **deleted outright**: natural language in, always-safe barewords out, with no escape hatch and the reproduced errors recorded in the doc comment so nobody reintroduces it. Verified there is no caller that hands it an FTS5 expression (two call sites, all natural-language or a bare entity name).

**Blocking 2 — non-ASCII recall regressed from working to erroring.** `fts5Barewords` accepted only `[a-z0-9_]`, but FTS5 barewords include every codepoint above 127 and this table is `porter unicode61`. A Cyrillic/Greek/CJK query sanitized to nothing, and `MATCH ''` is itself a hard error. A/B on identical data: `"Привет мир Москва"` returned a row **before** this PR and errored **after** it. Now: codepoints > 127 are barewords, terms are counted in runes, a lone CJK ideograph is kept as a whole word, and the empty-terms case never reaches `MATCH` — `Recall` falls through to its existing non-FTS branch and `SearchEntities` returns no matches without querying.

**Majors also fixed:** store errors now `RecordError` with a distinct `recall.outcome=store_error` (they were swallowed, so both failure modes above would have shown up in the next fleet run as `no_candidates`, indistinguishable from an honest miss — the observability this PR added didn't buy the diagnosability it was added for); `recall.query_source=llm|keyword` records whether the LLM side-call starved, which is the question the PR exists to answer; `recall.query` is truncated like other user-derived span text. Stopwords no longer drop `first`/`when`/`what`/`which`/`who`/`why`/`call`/`only` — temporal words are exactly the discriminators for the recall cases this targets, and the doc comment's own exemplar query uses "first". Minimum term length aligned at 2 across both tokenizers via cross-referencing named constants.

One extra fix beyond the review: `keywordSearchQuery` was also ASCII-only, so a non-Latin message plus a starved side-call produced an empty fallback — the same defect class sitting directly on the path being fixed.

**The real defect the review exposed** was that reverting either headline fix left every test green, because the tests asserted on strings instead of executing the query. The new tests **execute** every produced query against a real migrated FTS5 table (apostrophe+operator, comma, `?`, paren, quotes, `NEAR`, colon, Cyrillic, Greek, CJK, accented Latin, mixed-script) on both FTS tables. Isolated reverts produce **18 subtest failures**; all pass with the fixes. `graph_memory_store_fts_test.go` also gained the `//go:build fts5` tag every other file in that package had.

Still open, out of scope for this round: the Postgres backend's AND semantics diverging from sqlite's OR for the same `RecallOpts.Query`, the unbounded injected-context budget, and the parsed-but-dead `max_recall_candidates` / `min_salience_threshold` fields.

PR #357, opened 2026-08-22
feat(memory): lesson-lane extraction — capture assistant discoveries

ilsiepotamus:
The extraction prompt told the model to **ignore the assistant's error messages and process notes** — i.e., to discard exactly the material that carries method knowledge. A 30-wave transfer-learning study measured the consequence: agents recalled 300–900 task-echo memories per wave while never learning the single error→fix lesson (INTEGER overflow → BIGINT) that governed an **8× correctness gap** between trap-bearing and trap-free task shapes. Conclusion the data forced: *memory transfers method only if extraction captures method.*

Extraction now runs two lanes:
1. **User facts** — unchanged.
2. **Lessons** — assistant-side discoveries: error+resolution pairs, schema facts learned by probing, environment quirks, techniques that worked after failures — stated in reusable cause-and-fix form, `memory_type: "lesson"`, salience 0.8–1.0.

The chatter exclusion survives but narrows (progress narration, politeness, uninformative retries); an error paired with its resolution is explicitly exempt. Schema enum gains `lesson`.

Validation plan: ladder pass 3 on the live fleet rig after the current control pass completes — if trap-shape correctness moves off its 7% floor, the causal chain closes.

Related: #356 (recall survives fleet load). Full study data: The Memory Curve artifact, Part II.


ilsiepotamus:
Independent review found two blocking defects in the lane's core premise. Both fixed in `17c22814` (now `186a455a` after rebasing onto main), with revert-based evidence for each.

**Blocking 1 — "Verified lessons" were not verified in the default configuration.** Private-mode lessons shared a partition with the per-turn extractor and the model-callable `graph_memory` tool, and the recall filtered only on type — so the model could write `memory_type: "lesson", salience: 1.0` and get it served back under "each fix was observed to succeed". Fixed structurally rather than by filtering, three ways: mined lessons now live in a **derived partition** (`<agent>__lessons`) a future caller cannot accidentally write to; the per-turn extractor **drops** lesson-typed memories (not relabels — the revert experiment showed relabelling to `fact` keeps the identical untested theory in the store); and the `graph_memory` tool refuses the reserved type with an explanatory error.

**Blocking 2 — correct lessons were being demoted for doing nothing wrong.** Credit charged −0.15 whenever the post-injection event tail was empty, which is the normal ending of an interactive turn, and demotion was absorbing: below the floor a lesson is never recalled, so never injected, so can never earn its way back. Now a loss requires **observed contradiction** — a failure after the injection, or a mechanically-judged no-work "success" of the failing class (so the measured vacuous-fix poison still scores negative, while silence doesn't). Demotion is non-absorbing via bounded **re-trial and reinstatement**: one demoted lesson rides a 1-in-8 slot on the error lane only — the only lane that scores outcomes, so re-trial is never pure cost — presented under an honest "Under re-trial" heading, and a win reinstates it to the floor outright rather than crawling back at +0.02.

**Major 3 — the grounding gate switched itself off exactly when least grounded.** Fragment-less pairs (the transient-retry shape: identical input fails then succeeds) are now dropped before mining, so the gate is unconditional.

**Major 4 — the trust boundary is now written down**, following the lease-contract precedent: a new section in the architecture doc traces server text → `Result.Error.Message` → mining prompt → system role → the shared lane, with matching threat-model rows in `docs/reference/security-model.md` and a SECURITY note on the proto field. Three cheap mitigations are listed as follow-ups rather than built.

Also corrected the false claim that mining "runs once at conversation end over the full transcript" — it runs per user message, which makes a turn-1 error fixed on turn-2 structurally unmineable. That's now stated in both the code comment and the doc.

**Three things worth flagging that the review didn't raise:**
- `GraphMemoryStore.DecayAll` has **no production caller anywhere**, so outcome credit isn't merely the only salience-*raising* path — it's the only live salience signal at all.
- Fix 3 has a real recall cost: a genuine cross-statement fix whose baseline predates the ledger window also yields zero fragments and is now dropped. Documented rather than papered over.
- Private-mode lessons written before this change are stranded under the old partition — not migrated, simply no longer recalled. The measured campaign ran fleet-shared, where nothing moves.

The Status block now carries a warning that the measured 7%-vs-57% numbers predate these fixes. This PR stays out of the door and RESOURCE_HOLDER proof runs deliberately — its extra per-turn mining call and injected system messages would confound exactly those measurements.

Josh-Schoen:
The gap you're describing — extraction explicitly told to ignore the only
material that would teach an agent a reusable error→fix lesson — is real
and reachable: I confirmed the exclusion line in `graph_memory_extractor.go`
predates this PR and is unchanged by it. The approach (a structurally
separate ledger-mined lane, rather than loosening the existing exclusion)
also holds up: it keeps unverified assistant narrative out while still
capturing verified error→fix pairs, and the code enforces "verified" as a
property of *who wrote the row* (reserved memory type + dedicated storage
partition, both checked in the store query) rather than a prompt
instruction other writers are trusted to respect.

**Verdict: approve once F1 is addressed.**

## What this makes possible

For anyone running graph-memory-enabled agents against tool-using
workloads, this adds a genuinely new memory class: lessons that can't be
forged by the model (the `graph_memory` tool's `remember` action now
refuses `memory_type: "lesson"` outright, and the per-turn extractor drops
it silently) and are delivered two ways — at conversation start
(task-relevant) and the moment a matching error recurs (error-triggered,
querying on the error text itself, which the task-relevant lane structurally
can't reach). The outcome-credit machinery (Fix B) means a lesson that
keeps preceding failures gets demoted rather than sitting at its initial
salience forever, and demotion is reversible rather than an absorbing
state — I walked the full demote→retrial→win→reinstate path in
`TestDemotedLessonGetsRetrialAndCanReturn` and it's real, not aspirational.

## Where it could hurt

Everything here is opt-in behind `GraphMemoryConfig.Enabled` (unchanged
default `false`), so most consumers see zero behavior change. Two real
gaps, both non-blocking:

- **Postgres-backed deployments get a quieter feature than SQLite-backed
  ones.** `AdjustSalience` is implemented only on
  `pkg/storage/sqlite/graph_memory_store.go`; `pkg/storage/postgres/graph_memory_store.go`
  has no equivalent, so outcome credit silently no-ops there (soft type
  assertion, no error). Mining, storage, and both recall lanes work fine on
  Postgres — a bad lesson just never gets demoted. Doesn't look like it's
  called out anywhere as SQLite-only.
- The new `security-model.md` section on tool-output-as-system-prompt is
  the right call and genuinely candid about the risk — flagging it here
  only because it's the one thing an operator absolutely needs to read
  before touching `fleet_lesson_sharing`, not as a criticism.

If genuinely no other downside: the two-lane split doesn't touch the
existing per-turn extraction behavior at all (same prompt, same exclusion,
one added comment), so there's no regression risk to the memory class this
PR doesn't touch.

## Verified credit

- `buf lint`, `buf breaking --against main`, `buf generate`, and
  `go run ./cmd/generate-weaver` all ran clean with zero diff against
  committed `gen/`/`embedded/weaver.yaml` — proto and generated code are in
  sync.
- `go test -tags fts5 -race ./pkg/agent/... ./pkg/memory/...
  ./pkg/storage/sqlite/... ./pkg/shuttle/...` — all green, no race reports.
- The new 1016-line test file isn't schema-validation theater —
  `TestExtractLessonsAtEndStorePath` runs the exact INTEGER→BIGINT
  scenario from your PR description end to end (real ledger, real error
  text, a grounding-gate rejection of an ungrounded sibling "DECIMAL"
  theory) and asserts on partition isolation, salience clamping, and
  single-call idempotency.
- No migration needed and none is missing — `memory_type` is a free-text
  column with no `CHECK` constraint in either SQLite or Postgres
  migrations.

## Findings

**F1 (blocking approval, small fix).** `pkg/storage/postgres/graph_memory_store.go`
has no `AdjustSalience` method, so outcome credit is SQLite-only. Either
implement it there (mirrors the SQLite `UPDATE ... SET salience =
MIN(1.0, MAX(0.05, salience + ?))`), or explicitly document the SQLite-only
scope in `docs/architecture/lesson-grounding-and-credit.md` and
`docs/reference/security-model.md` so a Postgres operator doesn't read
"outcome credit demotes bad lessons" and assume it applies to them.

**F2 (non-blocking).** `injectGraphMemoryContext`'s
`span.SetAttribute("recall.injected", len(relevant))` doesn't count the
lessons injected in the same call — undercounts the observability metric
whenever a lesson fires with no echo-memory matches. One-line fix.

**F3 (non-blocking, description-only).** The PR body's "chatter exclusion
survives but narrows... error paired with its resolution is explicitly
exempt" reads like the per-turn prompt's own exclusion list changed. It
didn't — `graph_memory_extractor.go`'s exclusion sentence is untouched.
The actual mechanism (a second, structurally separate lane) is arguably a
*better* story than "narrowed an exclusion" — worth saying it that way so
a future debugger doesn't go looking for a changed exclusion list that
isn't there.

**F4 (non-blocking, pre-existing).** Both extraction prompts are hardcoded
Go strings, not `pkg/prompts.FileRegistry`. Confirmed this predates the PR
— not a regression — but this PR adds ~50 more lines to the pile. Worth a
separate tracked cleanup, not a hold on this PR.

## Things I was wrong about

- Assumed the missing-migration observation in my brief was a live bug
  (constrained column missing an updated CHECK constraint). It's not —
  `memory_type` is free text in both dialects, confirmed by grepping the
  migration SQL directly. No migration was ever needed.
- Assumed "schema enum gains lesson" referred to the proto change. It
  doesn't — the proto diff is entirely the new `fleet_lesson_sharing`
  config field; the lesson type itself is a plain Go string constant with
  no proto surface at all.

## Environment gotchas

- `go run ./cmd/generate-weaver` and `buf generate` both need to run
  before trusting a local `go build` — neither changed anything here, but
  it's easy to skip and get confusing unrelated-looking diffs.
- `buf breaking` needs a real local ref for `--against`; `origin/main` as
  fetched by a plain `git fetch origin main` won't resolve without also
  naming a local tracking ref.

Josh-Schoen (changes requested):
Requesting changes on F1 only — see the comment above for full detail. The core two-lane extraction mechanism is sound and well-tested (buf generate/gen parity confirmed, go test -race clean across all four touched packages, a genuine end-to-end error+resolution test, not schema-acceptance theater). But AdjustSalience (demoting a bad lesson) is implemented only for SQLite — a Postgres-backed deployment silently never demotes a wrong lesson. Small fix, worth holding for.

ilsiepotamus:
F1 fixed in `57237e09`, plus a merge of current `main` in `287d5b7b`.

**F1 — outcome credit was SQLite-only.** `AdjustSalience` now exists on the Postgres store too. The clamp could not be copied from the SQLite statement: `MIN`/`MAX` are *aggregates* in postgres, so `MIN(1.0, MAX(0.05, salience + ?))` would not port — it uses `LEAST`/`GREATEST` inside the store's `execInTx` + tracer ceremony, matching `DecayAll` beside it.

Tested against a real PostgreSQL 16 rather than compiled: both clamp bounds bind, a loss lowers and a win raises, the floor stops at 0.05 so demotion stays reversible (the property `TestDemotedLessonGetsRetrialAndCanReturn` depends on), and an unknown id is a no-op since credit iterates lessons that may have been forgotten mid-conversation. The test also asserts the store satisfies the capability interface `pkg/agent` type-asserts for, so removing the method later fails at compile time instead of silently disabling credit again — which is what made this quiet in the first place.

Separately, the branch needed a merge of `main`, which had extracted the conversation loop's inline per-call body into `dispatchOneCall`/`commitToolRow` (shared with `ResumeChat`'s parked-batch completion) while this branch had woven the lesson hooks into that same body. The hooks moved into `commitToolRow`, the per-batch error buffer moved onto `batchState` alongside `pendingSidecars`, and `maybeParkResourceBatch` runs before the lesson lane so the park sees the batch tail exactly as it did pre-lessons.

One thing that merge exposed, in neither side: `park.go`'s `completeParkedBatch` drained sidecars but had no lesson injection, so after the merge a resumed batch would collect error evidence through the shared body and silently drop it. It now mirrors the loop's drain.

CI green (18 pass); `-race` clean across `pkg/agent`, `pkg/memory`, `pkg/storage/...`, `pkg/shuttle/...`.

ilsiepotamus:
Merged current \`main\` at \`83e7cbd8\` (previous fix commit: \`57237e09\`, Postgres \`AdjustSalience\` parity).

That merge needed one real conflict resolved by hand in \`pkg/agent/agent.go\`: main had independently added an implicit-task-completion \`defer\` at the same call site (right after \`runConversationLoop\`) since this branch's last sync. Both now run — the async ledger-grounded lesson-mining pass and the task-completion defer are sequential, non-conflicting additions at that point in \`Chat()\`. Also regenerated \`gen/go/loom/v1/agent_config.pb.go\` via \`buf generate\` rather than hand-merging the generated diff.

Verified on the pushed head: \`go build -tags fts5 ./...\`, \`go test -tags fts5 -race ./pkg/agent/... ./pkg/memory/... ./pkg/shuttle/... ./pkg/storage/postgres/... ./pkg/storage/sqlite/...\` (all pass, including the Postgres AdjustSalience path), \`gofmt -l\` clean, \`golangci-lint run --build-tags fts5 ./...\` (0 issues), \`buf lint\`/\`buf format --diff\`/\`buf breaking --against main\` all clean, CI green.

PR #358, opened 2026-08-22
fix(mcp): map LLM parameter names onto the schema's declared names

ilsiepotamus:
## Problem

`MCPToolAdapter.Execute` has always blindly converted every outgoing parameter key from snake_case to camelCase (`normalizeParametersToCamelCase`), assuming MCP servers expect camelCase. For servers whose input schemas declare snake_case names this corrupts every call that carries a multi-word parameter:

- **Strict servers / strict client validation** (MCP 2026 client): the renamed key is rejected — `invalid arguments for tool execute_statement: Additional property sessionHandle is not allowed`.
- **Lenient decoders**: the renamed key is silently **dropped**. Observed in production-scale testing against a snake_case Teradata MCP server: `session_handle` → `sessionHandle` → ignored, so sessions were never pinned and `release_handle` never released — leaking server-side session handles until the budget filled.

The bug was masked until the MCP 2026-07-28 client migration made schema parsing and argument validation strict; then every snake_case parameter became a hard failure.

## Fix

Replace blind camelization with schema-aware mapping (`normalizeParametersToSchema`): a parameter name is kept as-is when the tool's `inputSchema` declares it, converted to camelCase only when the schema declares the camelCase form (preserving the existing LLM-sees-snake_case / camelCase-server behavior), and passed through untouched otherwise so validation errors name exactly what the model sent.

`InputSchema()` still presents snake_case names to the LLM; this change only fixes the reverse direction.

## Testing

- New regression tests: snake_case server params pass through untouched (the `session_handle` case), camelCase servers still get camelCase, unknown keys untouched, no-schema tools untouched.
- Updated cache-key tests to the schema-aware mapping.
- `go test -tags fts5 -race ./pkg/mcp/...` clean.
- Verified end-to-end against a live snake_case MCP server (Teradata, 12-agent fleet): rejections disappeared, sessions pin, handles release.

### DIFF
--- a/pkg/mcp/adapter/shuttle.go
+++ b/pkg/mcp/adapter/shuttle.go
@@ -194,13 +194,14 @@ func (a *MCPToolAdapter) InputSchema() *shuttle.JSONSchema {
 func (a *MCPToolAdapter) Execute(ctx context.Context, params map[string]interface{}) (*shuttle.Result, error) {
 	startTime := time.Now()
 
-	// Convert parameter names from snake_case back to camelCase for MCP tools
-	// LLMs naturally use snake_case but MCP tools expect camelCase
-	camelCaseParams := normalizeParametersToCamelCase(params)
+	// Map the LLM's parameter names onto the names the tool's schema
+	// declares (snake_case kept for snake_case servers, camelized only
+	// when the schema itself is camelCase).
+	serverParams := a.normalizeParametersToSchema(params)
 
 	// Check schema cache for schema-related tools (#4: Schema Caching)
 	if a.isSchemaLookupTool() {
-		cacheKey := a.buildSchemaCacheKey(camelCaseParams)
+		cacheKey := a.buildSchemaCacheKey(serverParams)
 		if cached, ok := globalSchemaCache.get(cacheKey); ok {
 			return &shuttle.Result{
 				Success:         true,
@@ -216,8 +217,8 @@ func (a *MCPToolAdapter) Execute(ctx context.Context, params map[string]interfac
 		}
 	}
 
-	// Call MCP tool with camelCase parameters
-	mcpResultInterface, err := a.client.CallTool(ctx, a.tool.Name, camelCaseParams)
+	// Call MCP tool with the schema-declared parameter names.
+	mcpResultInterface, err := a.client.CallTool(ctx, a.tool.Name, serverParams)
 	executionTime := time.Since(startTime).Milliseconds()
 
 	if err != nil {
@@ -255,7 +256,7 @@ func (a *MCPToolAdapter) Execute(ctx context.Context, params map[string]interfac
 
 	// Cache schema results (#4: Schema Caching)
 	if a.isSchemaLookupTool() {
-		cacheKey := a.buildSchemaCacheKey(camelCaseParams)
+		cacheKey := a.buildSchemaCacheKey(serverParams)
 		if str, ok := data.(string); ok {
 			globalSchemaCache.set(cacheKey, str)
 		}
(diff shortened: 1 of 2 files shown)


InfinityOverflow (changes requested):
Standards

Hard violations (documented standards)

1. Dead code left behind (CLAUDE.md: "Dont take shortcuts: No hardcodes, and do NOT add tech debt!") — normalizeParametersToCamelCase (shuttle.go:443-455) is now unreachable from any production path. Execute was its only caller, and the diff switches that call site to a.normalizeParametersToSchema. Confirmed by repo-wide grep: the function is referenced only by its own test (shuttle_test.go:752, now repurposed to exercise the new method rather than removed). The PR body itself says "Replace blind camelization with schema-aware mapping" — the old function should have been deleted along with its now-orphaned test.
2. Documentation, function size, error handling, context-first-param: no violations. The new godoc on normalizeParametersToSchema (shuttle.go:457-465) explains the why (the session_handle leak) well, matching CONTRIBUTING.md's documentation standard.

Judgement calls (smell baseline)

- toCamelCase is not dead — it's reused by the new method, so only normalizeParametersToCamelCase itself is the orphan.
- No Speculative Generality, Duplicated Code (beyond the redundancy noted above), Message Chains, Feature Envy, Data Clumps, Middle Man, Refused Bequest, Repeated Switches, Shotgun Surgery, or Divergent Change observed — the diff is narrowly scoped.

Spec

(a) Missing/partial — Same dead-code point as Standards: the spec's word "Replace" implies removal of normalizeParametersToCamelCase, which wasn't done. Not a functional bug, but risks a future reader assuming the old function is still load-bearing.

(b) Scope creep — None. The diff touches only Execute's parameter-mapping call sites, the new normalizeParametersToSchema method, and its tests. InputSchema(), toCamelCase/toSnakeCase, caching, and error handling are untouched, matching the PR's claim of only fixing "the reverse direction."

(c) Implemented but wrong — None of substance. Verified against the spec line-by-line:
- Schema-aware mapping (declared-as-is → camelCase-if-declared → pass-through) matches shuttle.go:466-486 exactly, including params == nil and no-properties-key edge cases (covered by TestNormalizeParametersToSchema_NoSchema).
- Confirmed InputSchema() (shuttle.go:133-191) is untouched and still presents snake_case to the LLM, as claimed.
- Nested/array param values aren't recursively mapped — but this matches pre-existing behavior of the old function too, so it's not a regression the spec asked to fix.
- Tests pass (race detector itself hit a sandbox TSan issue unrelated to this code; confirmed clean without -race).

---

Summary: Standards — 1 finding (1 hard violation: dead code). Spec — 1 finding (same dead-code gap, framed as an incomplete "replace"). Both axes converge on the identical, minor issue: delete normalizeParametersToCamelCase and its now-superseded direct test before merge. No scope creep, no logic bugs, no missing test coverage. This is a small, low-risk cleanup, not a blocker on functional grounds.

ilsiepotamus:
Addressed in `c1b6a2c8` — you were right, "replace" left the replaced thing behind.

`normalizeParametersToCamelCase` is deleted. Rather than dropping its table test outright, I folded its cases into `TestNormalizeParametersToSchema`: the nil-params and nested-value-preservation scenarios it covered are now asserted against the schema-aware method, alongside two cases the old test could not express (a snake_case server keeping its declared names, and a schema with no `properties` key). `TestNormalizeParametersToCamelCase_RoundTrip` — which already exercised the new method — is renamed, and the stale section header is corrected. Repo-wide grep confirms no reference to the old name survives.

Your read of the rest matches mine: `toCamelCase` stays (the new method uses it), `InputSchema()` is untouched, and nested/array values are still only top-level-mapped — pre-existing behaviour, not something this PR set out to change.

`go build` and `go test -tags fts5 -race ./pkg/mcp/adapter` pass; `gofmt` clean.

ilsiepotamus:
Closing this as superseded by #340 (`93da4b4e`, "fix(mcp): restore server's original parameter names instead of blind camelCasing", issue #339), which merged to `main` on 2026-08-24 and fixes the same bug this PR was opened for — including the `session_handle` case, with tests.

I only found this while resolving the merge conflict here, and it explains the CI oddity on this PR too: the branch is 35 commits behind `main` and conflicts in both files it touches, so GitHub could never build the `refs/pull/358/merge` ref and the `pull_request` workflows never started. Only CodeQL ran, because its default setup runs on `refs/pull/358/head` and needs no merge ref. The conflict is in exactly the lines where `main` superseded this work.

`main`'s approach is strictly better than the one here, so merging this would be a regression:

| | this PR (`normalizeParametersToSchema`) | `main` (`buildParamNameMap` + `restoreParameterNames`) |
|---|---|---|
| Recovering the original name | guesses via `toCamelCase` | inverts `toSnakeCase` from the schema, restoring the **exact** property name |
| Originals camelCase cannot reconstruct (`TableName`, `max_Rows`) | falls through unmapped | restored correctly |
| Two properties collapsing to one visible name (`tableName` + `table_name`) | silently picks one | detects the ambiguity and passes through untouched |
| Cost | remaps on every call | map built once at adapter construction |

`main` has also already removed `toCamelCase` and `normalizeParametersToCamelCase`, so @InfinityOverflow's dead-code finding — which was correct, and which I fixed here in `c1b6a2c8` — is resolved there too, by a different route. Thanks for that review; sorry it went into a branch that had already been overtaken.

Nothing here needs to be salvaged into `main`. The branch is left at `c1b6a2c8` if anyone wants the history.

PR #359, opened 2026-08-22
fix(llm): throttling pauses the conversation instead of killing it

ilsiepotamus:
## Problem

A 429 is provider flow control, not failure — but three stacked gaps turned every throttle into a dead conversation (measured on a 12-agent gpt-4o fleet on Azure eastus2: every 429 killed its conversation mid-task):

1. **Azure OpenAI client**: the request closure passed to the rate limiter returned the raw `*http.Response` — a 429 status is a perfectly valid HTTP response, so the limiter's *existing* backoff-and-retry loop never saw throttles at all. (The Anthropic client converts 429→error inside its closure for exactly this reason.)
2. **Agent streaming path**: `chatWithStreaming` bypassed retry entirely ("streaming already handles errors" — it doesn't), so the first throttle error propagated straight up and killed the conversation. This is the path fleet agents actually run (progress callbacks ⇒ streaming).
3. **Rate limiter backoff** was blind to provider-stated windows ("Please retry after 60 seconds" / `Retry-After`), so short blind backoffs burned attempts inside the throttle window.

## Fix

- Azure client: new `doWithRateLimit` converts 429 + transient 5xx (500/502/503/504/529) to errors **inside** the limiter closure, **rebuilds the `http.Request` per attempt** (a reused request's body reader is consumed by the first send — retries silently sent an empty body), and echoes `Retry-After` into the error text.
- Agent: shared `throttlePatience` helper used by the streaming path, the retry loop, and the retry-disabled path. Throttles never consume retry attempts; waits honor the provider-stated retry-after (else escalate 2s→60s); hard stop at a 10-minute patience budget or the conversation's own deadline. Streaming retries reset the token buffer so no tokens from a failed attempt are duplicated.
- Rate limiter: waits at least the parsed retry-after (capped 300s); exported `llm.IsThrottlingError` and `llm.RetryAfterHint`.

## Testing

- `pkg/llm`: hint parsing (Azure wording, header echo, cap), exported classifier, limiter retry loop end-to-end.
- `pkg/llm/azureopenai`: httptest — 429 converts to a classifier-parsable error with the Retry-After surviving; a throttled-then-retried request resends the **full body** (regression for the consumed-body-reader trap).
- `pkg/agent`: throttle-once stubs — retry-disabled path survives, throttles consume no attempts, the streaming path (the one that killed conversations) pauses/re-invokes with no duplicated tokens, patience budget hard-stops.
- `go test -tags fts5 -race ./pkg/agent ./pkg/llm/...` clean.


InfinityOverflow:
Please resolve the merge conflicts everything else is good and well implemented with minor nits.

Spec: ✅ Matches PR description point-for-point — 300s retry-after cap, 10-min patience budget, per-attempt body rebuild, token buffer reset, and throttle-doesn't-consume-attempt are all implemented and covered by passing tests. No scope creep. One cosmetic nit: stale comment at llm_retry.go:40 ("streaming already handles errors") should be updated since this PR disproves it.

Standards:
- 🔴 chatWithRetry and chatWithStreaming (pkg/agent/llm_retry.go) now exceed the repo's <100-line-per-function guideline (116 and 104 lines). Consider extracting the throttle-wait branches.
- 🔴 The new multi-minute throttle pause is only logged (zap.L().Warn), not traced to hawk — breaks "every operation traced to hawk." Worth adding a span/event for the wait duration + retry-after hint.
- 🔴 pkg/llm/throttle_test.go's TestIsThrottlingErrorExported/TestRetryAfterHint are hand-rolled instead of table-driven, unlike the sibling TestIsThrottlingError.
- 🟡 (nits, judgement calls) attempt--; continue in the retry loop is a bit fragile/implicit; throttlePatience.wait duplicates backoff logic already in RateLimiter.executeWithRetry; magic 529 status has no comment; "API error (status %d): %s" formatting is duplicated 3x in azureopenai/client.go.

Nothing here blocks merge — all are pre-existing-pattern-adjacent, not correctness bugs.

Josh-Schoen:
Thanks for chasing this down to a concrete field symptom (12-agent fleet, every 429 killing its
conversation) rather than a hypothetical — that made it very checkable. The problem is real and
the streaming-path piece of it is still real today.

**Verdict: blocked.**

Not on code quality — the diff itself is careful (typed rebuild-per-attempt, well-targeted
tests, no shortcuts). It's blocked because `git merge origin/main` (thrown away, not pushed)
produces real conflicts in `pkg/llm/azureopenai/client.go`, `pkg/llm/rate_limiter.go`, and a
same-filename collision on `pkg/llm/throttle_test.go`. The reason is bigger than a stale branch:
`main` picked up PR #352 ("LLM slot scheduler core," merged two days after this PR opened) which
independently fixed two of your three claimed bugs — the Azure client swallowing a 429 as a
valid response, and the backoff ignoring the provider's wait window — via a typed
`llm.ThrottleError` + `RetryAfterFromHeaders` (reads the real `Retry-After`/`retry-after-ms`/
`x-ratelimit-*` headers, not a text regex) that's a more complete mechanism than
`doWithRateLimit`/`RetryAfterHint` here. I checked `main`'s `pkg/llm/azureopenai/client.go`
directly — its `sendOnce` closure does exactly what your `doWithRateLimit` does: rebuild the
request per attempt, convert 429 to an error inside the closure.

**What's still real and still worth landing**: your bug #2. I wrote a throwaway test against
current `main` (not your code — deleted after, not part of any diff) with a stub `ChatStream`
that fails once with a 429-style message. It propagated straight up and killed the call, exactly
as your PR description says. The scheduler work that landed since observes the throttle for
future capacity calibration, but that doesn't save the conversation that's throttled right now —
`dispatchChat`'s streaming branch is still `resp, err := streamingProvider.ChatStream(...); if
err != nil { return nil, err }`, unchanged. And I confirmed this is the path that matters: every
streaming RPC the multi-agent server exposes wires a progress callback unconditionally
(`pkg/server/multi_agent.go:1306`), which is what routes `dispatchChat` into the streaming
branch — this is the default path for fleet agents, not an edge case.

**Where it hurts**: as-is, resolving the conflicts by keeping both your
`IsThrottlingError`/`RetryAfterHint` and `main`'s `IsThrottle`/`RetryAfter` would leave two
parallel, overlapping throttle-classification APIs in `pkg/llm` for the next person to trip
over. And your `doWithRateLimit`/`RetryAfterHint` changes to the Azure client and rate limiter
are now redundant with what shipped — they'd need to come out, not merge in.

**Verified credit**: `TestDoWithRateLimit_RebuildsBodyPerAttempt` genuinely proves the
consumed-body-reader trap and its fix, against a real `httptest.Server` — good regression test,
worth keeping the concept even though `main`'s `sendOnce` already does the rebuild. All four new
`pkg/agent` throttle tests and both `pkg/llm/azureopenai` tests pass with real `=== RUN` lines
under `-race -tags fts5`. Full `pkg/agent` and `pkg/llm`/`pkg/llm/azureopenai` suites pass clean
on your branch, no races, no regressions. One piece of your Azure-client change is *not* on
`main` and is a genuine improvement worth keeping: treating transient 5xx (500/502/503/504/529)
as retry-worthy alongside 429 — `main`'s current `sendOnce` only handles 429.

**Findings**:
- F1/F2 — real merge conflict, root-caused to `main`'s independent fix (PR #352, `main
  b5e8361`) for two of your three bugs. Affects: whoever rebases this.
- F3 — streaming path still kills the conversation on first throttle. Still real, still
  unfixed on `main`. Affects: operator running fleet agents via streaming RPC. Fix location on
  rebase: `main`'s current `pkg/agent/llm_retry.go` `dispatchChat`/`chatWithStreaming` — port
  your `throttlePatience` pause-and-retry there, using `llm.IsThrottle`/`llm.RetryAfter` (the
  symbols `main` now exports) instead of the ones this PR adds.
- F8 — worth checking before you drop the Azure-body-message regex: does Azure OpenAI ever
  state the retry window only in the JSON error body, not in a `Retry-After` header? If so,
  `main`'s header-only `RetryAfterFromHeaders` misses it and your regex approach is still worth
  folding in (as a body-text fallback), not discarding.
- F10 — the 5xx-as-retryable addition (`retryableUpstreamStatus`) isn't on `main`. Worth
  porting into `main`'s `sendOnce` separately from the rest.

**Things I was wrong about**: I went in assuming all three claimed bugs were independently real
and this PR was blocked only by an accident of timing. Two of the three turned out to already be
fixed on `main`, better, by unrelated work that landed while this PR sat open. Only the
streaming-path fix is still needed.

**Suggested path**: don't try to resolve the conflicts as a merge. Rebuild this PR against
current `main` as a small, focused change: port `throttlePatience` into `main`'s
`chatWithStreaming`, wire it to `llm.IsThrottle`/`llm.RetryAfter`, keep (or adapt) your
streaming test, and separately consider the 5xx-retryable and body-message-regex points above.
That should end up as a much smaller diff than this one.

Josh-Schoen (changes requested):
Requesting changes — see the comment above. Real conflicts against current main (confirmed via a throwaway merge attempt), and two of the three claimed fixes are already superseded by a better version merged separately (PR #352). The third — streaming bypasses retry, which is the default path fleet agents run — is still genuinely broken on main today (confirmed with a throwaway stub-429 test). Needs a rebase that drops the redundant changes and keeps the still-needed streaming fix.

ilsiepotamus:
Your read matches what I found independently, and it is already done — your review landed on `bb98f238`; the head is now `ccb540e2` (merge of current `main`), pushed shortly after.

That merge does exactly what you recommend:

- **Dropped as superseded by #352**: `pkg/llm/azureopenai/client.go` (main's inline `sendOnce` already rebuilds the request per attempt and raises 429 inside the limiter closure) and `pkg/llm/rate_limiter.go` (main's `retryDelay` already floors on the server-stated wait, and adds jitter this branch lacked). The branch's `IsThrottlingError` / `RetryAfterHint` / `parseRetryAfter` are deleted in favour of main's `llm.IsThrottle` / `llm.RetryAfter` / `ThrottleError`, whose `RetryAfterFromHeaders` parses `retry-after-ms`, HTTP-date and `x-ratelimit-reset-*` instead of regexping the error string.
- **Kept**: the streaming-path fix — still broken on main today, exactly as you confirmed with your stub-429 test.

The PR is now 4 files instead of 6; the diff against `main` is just the patience helper, its tests, one reconciled test, and an Azure regression test.

Two things the merge surfaced that were in neither side, both fixed here:

1. **The absorbed throttle was invisible to AIMD.** `observeSchedulerOutcome` runs in a defer that only sees the call's *final* error, so once patience rides a 429 out and the retry succeeds, the seam saw SUCCESS and grew the scope's ceiling — on precisely the signal that must shrink it. `absorbThrottle` now reports at the moment of absorption (the scheduler already collapses repeats within one congestion event, so a burst still halves once). Pinned by `TestChatWithRetry_AbsorbedThrottleStillHalvesCeiling`.
2. **`TestFunnelObservesThrottleAndSuccessProviderAgnostic` hung.** Its stub throttles every call, which patience rode to the 10-minute budget. That is a genuine semantic change, not a flake: a permanently-throttling provider now surfaces at the caller's deadline rather than instantly. The test gets a 200ms deadline; its intent — a surfaced throttle halves the ceiling — is unchanged.

I also ported the branch's body-resend assertion onto main's structure: `TestCallAPI429IsRetriedThroughRateLimiter` never inspects the request body, so it passes even if a retry ships an empty one.

CI on `ccb540e2`: 18 pass, 0 failures. Ready for another look.

ilsiepotamus:
Merged current \`main\` at \`42e46a82\` (previous state: \`ccb540e2\`, the merge that already dropped the changes superseded by #352 and kept the streaming-retry fix, per the earlier reply above).

That merge needed one real conflict resolved by hand in \`pkg/agent/llm_retry.go\`: main had independently grown a \`NotifyStreamActivity\`/tool-input-heartbeat mechanism in \`chatWithStreaming\` since this branch's last sync. Combined it with the throttle-patience retry loop — the stream now runs through \`streamCtx\` (carrying the activity callback) inside the patience retry loop, so a throttled streaming call still gets tool-input liveness pulses across retries.

Verified on the pushed head: \`go build -tags fts5 ./...\`, \`go test -tags fts5 -race ./pkg/agent/... ./pkg/llm/...\` (all pass), \`gofmt -l\` clean, \`golangci-lint run --build-tags fts5 ./...\` (0 issues), \`buf lint\`/\`buf format --diff\`/\`buf breaking --against main\` all clean, CI green.

PR #362, opened 2026-08-24
fix(mcp): free the stdio listen id before the overflow cancelled cue

ilsiepotamus:
## Summary

`notifications/cancelled` is the stdio binding's gap signal: the client is expected to re-subscribe and refetch as soon as it arrives. But the listen id was only released in the pump's deferred cleanup — *after* the cue was already on the wire. A client that re-subscribed promptly could race the defer and get bounced with `duplicate subscription id on this connection` (an error response, no `method` field).

This surfaced as a rare CI flake in `TestStdioSubscriptionOverflowSendsCancelled` on #356 (an unrelated graph-memory PR): the re-subscribe read `"" does not contain "notifications/subscriptions/acknowledged"` — the empty method of that error response. The window is the gap between the pump's `t.Send(cancelled)` and its deferred `conn.remove(listenID)`.

## Fix

Cleanup (subscription-map delete + `conn.remove`) is now single-flight via `sync.Once`:
- the **overflow branch runs it inline, before sending `cancelled`** — so the id is provably reusable by the time the cue can reach the client (cleanup → send → client read is a happens-before chain);
- the deferred call remains for every other exit path and becomes a no-op after the inline run, so the old pump can never remove a successor's registration under the same id.

The stream (HTTP/gRPC) binding is unaffected: it closes the whole stream on overflow and has no per-connection listen-id registry.

## Testing

- Regression tightened to encode the contract: re-subscribe **immediately** on the cue, with no wait on server internals in between; asserts exactly the replacement subscription remains registered.
- `go test -tags fts5 -race -run TestStdioSubscription -count=100 ./pkg/mcp/server/` clean.
- Full `just check` gate green (lint + tests + build, matches GitHub CI).


Josh-Schoen:
Nice catch, and a nice fix. The `notifications/cancelled` cue on stdio really is a promise that
the listen id is reusable, and having cleanup run in the pump's defer *after* the send broke that
promise for a narrow window on the overflow path specifically. Framing it as a happens-before
chain (cleanup → send → client read) is the right way to reason about this, and the `sync.Once`
guard is the right tool for "run this exactly once from either of two call sites without one
clobbering a *later* subscription's registration under the same id."

**Verdict: approve.**

**What this makes possible:** a client can now re-subscribe with the same listen id the instant
it sees the cancellation cue, with zero dependency on server-internal timing. That's exactly the
contract the binding was already documented to offer.

**Where it could hurt:** genuinely, nowhere I can find. This only moves already-happening cleanup
earlier on one branch; every other exit path is unchanged, and the stream binding doesn't share
this code path at all.

**Verified credit:**
- Reproduced the pre-fix race directly rather than trusting the description: pre-fix source +
  your tightened test, `-race -count=1000 -cpu=1,2,4,8` → 52/4000 failures, with the *exact*
  symptom you described (`"" does not contain "notifications/subscriptions/acknowledged"`, plus a
  transient 0-not-1 subscription count). Same settings against this PR's actual code → 4000/4000
  clean.
- Full `pkg/mcp/server` package under `-race -v` — no regressions elsewhere.
- Confirmed the regression test runs in CI's required race-detector job (`./pkg/... ./internal/...
  ./cmd/...`), not an advisory-only job — no skip, no build-tag gate.
- Confirmed the old test genuinely could not have caught this: it waited on
  `len(s.subscriptions) == 0` before re-subscribing, which waits out the exact race window it
  should have been probing. The new test's "re-subscribe immediately, no wait" shape is what
  makes it a discriminating test rather than a happy-path one.

**Findings:**
- None blocking. One thing worth a one-line comment for the next contributor rather than a
  change: the `sync.Once` guard only protects call sites that go through the shared `cleanup()`
  closure — if a future exit path is added to this pump, it needs to call `cleanup()` too, not
  duplicate the delete/`conn.remove` inline. Not asking for anything here, just flagging it since
  the file will likely grow more exit paths over time.

**Things I was wrong about:** I expected a modest-iteration `-race` run against the pre-fix code
to fail readily if the race were real, and it didn't — 200/200 clean at first pass. Only widening
to `-cpu=1,2,4,8` fan-out surfaced it, at ~1.3%. That's consistent with your own "rare CI flake"
framing, not a contradiction of it — worth remembering for the next race-condition PR review: a
clean low-iteration run doesn't refute a claimed rare race, it just means the budget didn't hit
the window.

**Environment gotchas:** none beyond the usual `-tags fts5` requirement, which this diff doesn't
interact with.

Josh-Schoen (approved):
Approve. Reproduced the race fix directly: reverted to pre-fix code in a scratch worktree, ran the PR's own regression test under `-race -count=1000` across `-cpu=1,2,4,8` — 52/4000 failures on old code, 0/4000 on the PR's code, matching the exact described symptom. Full findings in the comment above.

PR #363, opened 2026-08-25
feat(client): add the schedule surface to the client SDK

Josh-Schoen:
`pkg/tui/client` is the client every Loom surface goes through — `cmd/loom`, `cmd/looms`, and `loom-standalone` all sit on it — and it has **no schedule methods at all**. Loom's scheduling subsystem is therefore unreachable from any of them: no surface can list a routine, read its history, or trigger one, regardless of how its UI is drawn.

This adds list, get, history, create, update, delete, trigger, pause, resume, and workflow executions.

## Details worth keeping

- **`ListSchedules` takes `enabledOnly`, and management views should pass `false`.** A paused schedule that is invisible looks deleted, and nothing can resume what it cannot see. It also takes `pageSize`/`pageToken` and returns the server's next page token, matching its sibling `ListWorkflowExecutions` — the server doesn't paginate this RPC yet (`multi_agent.go` still has a `TODO: Implement pagination`), so today every call returns everything and an empty token regardless of `pageSize`, but the plumbing is there so this method doesn't need a breaking signature change once the server catches up.
- **`GetScheduleHistory` maps `limit<=0` onto the server's own default** rather than forwarding `0`, which some servers read as "none".
- **`UpdateSchedule` leaves a nil pattern or schedule untouched**, so a caller changing only the cron does not have to round-trip the pattern and risk writing back a stale copy of it.
- **`CreateSchedule` returns the stored `ScheduledWorkflow`, not just its ID.** The server-computed `next_execution_at` is the only honest confirmation that a cron expression parsed the way its author intended, and callers should show it back to them.

`Wrap()` builds a `Client` over a connection the caller already holds. For a caller whose own generated bindings share that connection, the alternative is a second connection to the same server or a reimplementation of this package. It **borrows** the conn: `Close` is a no-op and closing remains the owner's job. `Wrap(nil)` returns `nil` rather than a `Client` that would panic on first use.

## Testing

The tests assert what the **server received**, not just what came back. These methods mostly translate arguments into request messages, so a test that only checked the return value would pass even with a field dropped on the way out. `GetSchedule` and `GetWorkflowExecution` — the two single-field passthroughs that initially shipped untested — now have request-assertion tests too, plus a dedicated `ListSchedules` paging test.

`gofmt`, `go vet` clean; `pkg/tui/client` passes under `-race` with `-count=1`.

## Scope

**No proto changes.** Purely additive to the SDK — 3 files.

A follow-up adds `CancelWorkflowExecution` (proto + scheduler + server) and the matching `CancelExecution` SDK method; it stacks on this branch because both touch `schedules.go`. This one stands alone and has no contract implications.

## Rebase / review-feedback note (2026-09-10)

Merged current `main` in (clean, zero conflicts — `pkg/tui/client/schedules.go` and `schedules_test.go` are new files, and `client.go`'s changes touch a different region than anything main added since this branch was cut). Then addressed the outstanding `CHANGES_REQUESTED` review:

- Added `TestGetSchedule`/`TestGetWorkflowExecution` (were at 0% coverage).
- Wired `pageSize`/`pageToken`/`nextPageToken` through `ListSchedules` (previously the only method in the file without pagination plumbing).
- Narrowed the file-level doc-comment's FailedPrecondition claim — it doesn't apply to `ListWorkflowExecutions`/`GetWorkflowExecution`, which read from the workflow store directly.
- Documented `Wrap(nil)`'s return-`nil` behavior in its own godoc.

See PR comment for the full response to both reviews, including where I left one minor (error-wrapping style) unchanged and why.


InfinityOverflow (commented):
Verified UpdateSchedule's nil-preserving behavior, GetScheduleHistory's limit<=0 default, CreateSchedule's full-object return, and Wrap()'s no-op Close() — all match the PR description and are tested via a real bufconn server asserting on the request received, not just the response.
 One gap: GetSchedule and GetWorkflowExecution are implemented but have no tests. The PR's own rationale for the request-assertion test style ("a test that only checked the return value would pass even with a field dropped") applies equally to these two — worth adding before/after merge.
Minor nit (non-blocking): Wrap()/borrowed in client.go is a second, unrelated change bundled into a "schedules" PR — fine since it's called out in the description, just flagging for awareness.
No hawk tracing on the new methods, but that's consistent with every other method in this package (untraced), so not a regression introduced here.

ilsiepotamus (changes requested):
# PR #363 Review: `feat(client): add the schedule surface to the client SDK`

Reviewed at PR head (`caf1ad15` + merge commit `863b9569`, merged with `main` @ `5d150b2e` on 2026-09-06). Note: `origin/main` has advanced 3 more commits since that merge (`fb727480`, `2af3bce4`, `b24fce41` — a `pkg/agent` feature and two dependency bumps), none touching `pkg/tui/client/`. Diff still applies cleanly, build/vet/lint/tests all pass at PR head, so the `mergeable: UNKNOWN` GitHub shows is almost certainly just an uncomputed cache state, not a real conflict — worth one final rebase-check immediately before merge since `main` is a live target, but nothing here suggests actual conflict risk.

## 1. DESCRIPTION

"Schedule surface" = the full CRUD + control surface for Loom's cron/workflow scheduler, added to `pkg/tui/client.Client` (the shared SDK every surface — `cmd/loom`, `cmd/looms`, `loom-standalone` — sits on):

- **`pkg/tui/client/schedules.go`** (new, 207 lines): `ListSchedules`, `GetSchedule`, `GetScheduleHistory`, `CreateSchedule`, `UpdateSchedule`, `DeleteSchedule`, `TriggerSchedule`, `PauseSchedule`, `ResumeSchedule`, `ListWorkflowExecutions`, `GetWorkflowExecution` — 11 new methods.
- **`pkg/tui/client/client.go`**: adds `Wrap(conn *grpc.ClientConn) *Client` (client.go:174-184) and a `borrowed bool` field (client.go:41-43) so `Close()` (client.go:188-196) becomes a no-op on a wrapped connection. This is connection-lifecycle plumbing, not schedule-specific, but is the mechanism that lets a caller with its own gRPC connection (e.g. a binary with its own generated bindings against the same server) pick up the schedule methods without a second dial.
- **`pkg/tui/client/schedules_test.go`** (new, 357 lines): bufconn-based mock-server tests for 9 of the 11 methods plus `Wrap`.

**RPC/proto existence verified**: `git diff origin/main...HEAD -- gen/go/ proto/` is empty — confirmed no proto or generated-code changes in this PR. Every RPC method (`ListScheduledWorkflows`, `GetScheduledWorkflow`, `GetScheduleHistory`, `ScheduleWorkflow`, `UpdateScheduledWorkflow`, `DeleteScheduledWorkflow`, `TriggerScheduledWorkflow`, `PauseSchedule`, `ResumeSchedule`, `ListWorkflowExecutions`, `GetWorkflowExecution`) and every message type it touches already exists in `gen/go/loom/v1/loom_grpc.pb.go` (interface at lines 241-258) and is already implemented server-side in `pkg/server/multi_agent.go` (e.g. `ScheduleWorkflow` at 4332, `ListScheduledWorkflows` at 4455, `GetScheduleHistory` at 4597). **This PR does not call anything that doesn't already exist on `main`.** Purely additive SDK work, proto-first-compliant by construction (nothing to generate).

## 2. SHAPE REVIEW

Read `CLAUDE.md` in full. Assessment against it:

**Good:**
- **Thin wrapper, no reimplemented logic.** Every method builds a request struct from its arguments and calls the generated stub directly; the only "logic" is argument-shaping (`GetScheduleHistory`'s `limit<=0` default at schedules.go:72-74; `UpdateSchedule`'s nil-passthrough at schedules.go:115-128), and both are verified faithful to what the server itself does with those fields (server honors nil `Pattern`/`Schedule` as "leave alone" at `pkg/server/multi_agent.go:4411-4416`).
- **Modularity**: new `schedules.go` file rather than growing the already-915-line `client.go` — directly matches CLAUDE.md's "Break large files into logical smaller files. Avoid monolithic files."
- **Tests assert wire-level correctness**, not just return values: `scheduleMockServer` (schedules_test.go:32-107) records every request it receives, and tests assert on the server-observed request fields (e.g. schedules_test.go:279-284 checks `mock.gotUpdate.Pattern == nil` reaches the wire as nil, not a zero-value message). This is a materially better discipline than most of the pre-existing `client_test.go`/`error_test.go` suite.
- **No marketing speak.** Scanned all new comments against the banned-word list (production-ready, comprehensive, seamless, enterprise-grade, robust, etc.) — none present. Every substantive doc-comment claim I spot-checked against the server implementation turned out accurate (see below for the one exception).
- `gofmt -l`, `go vet -tags fts5 ./pkg/tui/client/...`, and `golangci-lint run ./pkg/tui/client/...` are all clean (verified independently, matches the PR's own claims).

**Issues (all minor):**
- **Doc-comment overclaims scope.** schedules.go:23-26 states as a blanket rule: *"The server answers these only when it was built with a scheduler; without one it reports FailedPrecondition."* Verified against `pkg/server/multi_agent.go`: true for 9 of the 11 wrapped RPCs (all gate on `sched == nil` → `FailedPrecondition`, e.g. lines 4352-4354, 4394-4396, 4441-4443, 4461-4463, 4500-4501, 4534-4536, 4561-4562, 4584-4585). **False for `ListWorkflowExecutions`/`GetWorkflowExecution`** (multi_agent.go:4244-4257, 4261-4290ish) — these read straight from `s.workflowStore` and never check `s.scheduler` at all; they work fine on a schedulerless server. A caller following this doc-comment literally would misinterpret an empty/working executions list as a missing-scheduler signal.
- **`Wrap`'s exported godoc never documents its nil behavior.** client.go:165-173 (the doc comment) describes ownership semantics but says nothing about `conn == nil`; the actual behavior (`return nil` at client.go:175-177) is only discoverable by reading the body or the test (`TestWrapNilConn`, schedules_test.go:353-357).
- **Error-wrapping convention is split, and this PR follows the older half.** `schedules.go` never wraps gRPC errors (bare `return nil, err` / direct passthrough throughout, e.g. schedules.go:48-51, 62). This matches the *original* Session/Weave-era style in client.go (`GetSession` client.go:311-317, `ListSessions` client.go:320-332), but every method added more recently in the same file — MCP servers, models, artifacts — wraps with `fmt.Errorf("failed to X: %w", err)` (e.g. client.go:595-601 `ListMCPServers`, client.go:685-691 `ListArtifacts`). Since `client.go` itself is already inconsistent, this isn't a clean violation, but it's a case where the newer/majority-of-recent-additions convention was available and not used.
- Minor scope note: `Wrap()`/`borrowed` (client.go) is genuinely useful but not schedule-specific; bundled into a PR titled specifically about the schedule surface. Well-disclosed in the PR description and well-tested, so not a real problem — just flagging that it's a second, independent change riding along.

## 3. COMPREHENSIVE REVIEW

**Correctness — field mapping (verified against `gen/go/loom/v1/loom.pb.go` struct definitions):**
All request/response field mappings check out exactly, with zero dropped or mismatched fields: `ListSchedules`→`EnabledOnly` (loom.pb.go:6486); `GetScheduleHistory`→`ScheduleId`+`Limit` (6807-6809); `CreateSchedule`→`WorkflowName`/`Pattern`/`Schedule`/`Metadata` (6248-6254, all 4 fields of `ScheduleWorkflowRequest`); `UpdateSchedule`→`ScheduleId`/`Pattern`/`Schedule` (6376-6380, all 3 fields of `UpdateScheduledWorkflowRequest`); `TriggerSchedule`→`ScheduleId`/`SkipIfRunning`/`Variables` (6651-6655); `ListWorkflowExecutions`→`StatusFilter`/`PatternTypeFilter`/`PageSize`/`PageToken` (6011-6017, all 4 fields, and it correctly returns `NextPageToken`, schedules.go:197). This is careful, complete work.

**Correctness — gap:** `ListSchedules` (schedules.go:43-54) exposes only `enabledOnly` and completely ignores `ListScheduledWorkflowsRequest.PageSize`/`.PageToken` (loom.pb.go:6488-6490), and discards `ListScheduledWorkflowsResponse.NextPageToken` (loom.pb.go:6552) — no pagination is reachable through this method at all, unlike its sibling `ListWorkflowExecutions` in the *same new file*, which fully wires `pageSize`/`pageToken`/`nextPageToken` through. Currently harmless in effect, because the server's own `ListScheduledWorkflows` (`pkg/server/multi_agent.go:4481`: `// TODO: Implement pagination if needed (page_size, page_token)`) ignores those fields too today — but this SDK method is additive-only surface that other code will now build against, and Go method signatures can't grow new positional params later without breaking callers. Once server-side pagination lands, this method needs a breaking signature change or a parallel method to expose it, whereas doing it now (even as a no-op pass-through) would have cost nothing. Given CLAUDE.md's "do NOT add tech debt," worth raising as a design gap rather than a bug.

**Correctness — context/cancellation:** consistent with every sibling method in `client.go` — `ctx` flows straight into the generated stub with no local timeout/deadline logic (verified: `Config.Timeout` is stored at client.go:87-88 but never consulted by *any* method, old or new — confirmed by grep across the whole package — so this is a pre-existing, package-wide characteristic, not something the schedule methods regress). Nil-response handling matches sibling idioms exactly: generated stubs guarantee `(nil, err)` on failure, so unwrapping (`resp.Schedule`, `resp.Executions`, etc.) at schedules.go:53, 86, 107, 127 is safe by construction, same as pre-existing `ListSessions`/`ListArtifacts`.

**Consistency:** naming departs from the RPC names deliberately and reasonably — proto RPCs are verbose (`ScheduleWorkflow`, `GetScheduledWorkflow`, `UpdateScheduledWorkflow`, `DeleteScheduledWorkflow`, `TriggerScheduledWorkflow`), SDK renames to a tight, consistent `*Schedule` CRUD vocabulary (`CreateSchedule`/`GetSchedule`/`ListSchedules`/`UpdateSchedule`/`DeleteSchedule`/`TriggerSchedule`). This reads better than a 1:1 mirror and isn't a drift — no complaint. Argument style (positional args decomposed from the request, e.g. `CreateSchedule(ctx, workflowName, pattern, schedule, metadata)`) matches the older Session/Artifact convention (`ListArtifacts` takes 6 positional params) rather than the newer "pass the whole `*loomv1.XxxRequest` struct" convention used by MCP-server/Pattern methods — both conventions already coexist in `client.go`, so this is a defensible pick, not an invented pattern.

**Test coverage:** Ran `go build -tags fts5 ./...`, `go vet -tags fts5 ./pkg/tui/client/...`, `golangci-lint run ./pkg/tui/client/...`, and `go test -tags fts5 -race -count=1 ./pkg/tui/client/...` myself — all clean, all 40 tests pass, matching the PR's own claims. Coverage (`go test -cover`) shows every new method covered **except two**: **`GetSchedule` (schedules.go:57-63) and `GetWorkflowExecution` (schedules.go:201-207) are both at 0.0% coverage** — confirmed both by the coverage report and by source inspection: `scheduleMockServer` (schedules_test.go:32-107) implements no `GetScheduledWorkflow` or `GetWorkflowExecution` handler, and no `TestGetSchedule`/`TestGetWorkflowExecution` function exists anywhere in the file (full function listing checked). This directly undercuts the PR's own stated test rationale ("these methods mostly translate arguments into request messages, so a test that only checked the return value would pass even with a field dropped on the way out") — that argument applies just as much to these two single-field passthroughs, and nothing currently verifies `ScheduleId`/`ExecutionId` actually reaches the wire for either. Low risk given both are one-line passthroughs, but cheap to close and inconsistent with the rest of the file's own diligence. Everything else (`ListSchedules`, `GetScheduleHistory` — table-driven, 3 cases, schedules_test.go:164-194 —, `TriggerSchedule`, `PauseSchedule`/`ResumeSchedule`/`DeleteSchedule`, `CreateSchedule`, `UpdateSchedule`, `ListWorkflowExecutions`, `Wrap`) is well tested, including a genuine behavioral test for `Wrap` (schedules_test.go:311-351, `TestWrapBorrowsConnection`, proves `Close()` on a wrapped client doesn't kill the underlying conn by making a second RPC call afterward — not just a type-check).

**Documentation accuracy:** covered above under Shape Review — one real inaccuracy (file-level FailedPrecondition claim overclaims to `ListWorkflowExecutions`/`GetWorkflowExecution`), one omission (`Wrap`'s nil behavior undocumented in its own godoc). Every other doc-comment claim I checked against the current server implementation (nil-pattern-preserved, `next_execution_at` as confirmation, `enabledOnly` semantics, `skip_if_running` default recommendation) is accurate today — no aspirational or invented claims.

---

**Majors:**
1. `GetSchedule` and `GetWorkflowExecution` (schedules.go:57-63, 201-207) ship with zero test coverage (0.0%), contradicting the PR's own stated testing rationale.
2. `ListSchedules` (schedules.go:43-54) exposes no pagination controls and silently drops `NextPageToken`, unlike its sibling `ListWorkflowExecutions` in the same file — currently harmless (server doesn't paginate either) but bakes in a signature that can't grow pagination later without a breaking change.

**Minors:** file doc-comment overclaims FailedPrecondition-without-scheduler scope to 2 RPCs that don't gate on it (schedules.go:23-26 vs. multi_agent.go:4244-4290ish); `Wrap`'s godoc omits its nil-input contract (client.go:165-184); error-wrapping style diverges from the more-recently-established convention elsewhere in the same file (not a clean violation, since `client.go` is already split on this).

VERDICT: MINOR ISSUES — Majors: 2, Blockers: 0

Josh-Schoen:
Rebased onto current `main` (clean merge, zero conflicts — the only overlapping file was `client.go`, and main's changes since this branch was cut land in a different region) and addressed the outstanding review feedback. Head is now `61e9e13`.

## @ilsiepotamus — CHANGES_REQUESTED

**Major 1 — `GetSchedule`/`GetWorkflowExecution` at 0% coverage.** Fixed. Added `scheduleMockServer.GetScheduledWorkflow`/`GetWorkflowExecution` handlers and `TestGetSchedule`/`TestGetWorkflowExecution`, both asserting the ID reaches the wire (not just that it comes back), consistent with the rest of the file. Both methods are now at 100% coverage (`go tool cover -func`).

**Major 2 — `ListSchedules` has no pagination, unlike its sibling `ListWorkflowExecutions`.** Agreed and fixed — this is exactly the kind of thing that's free before a method ships and a breaking change after. `ListSchedules` now takes `pageSize, pageToken` and returns `(schedules, nextPageToken, error)`, matching `ListWorkflowExecutions`'s shape. Added `TestListSchedulesPaging` asserting both the request fields and the returned token. No existing callers outside this package/its own tests (grepped the tree), so the signature change is free right now.

**Minor — doc-comment overclaims FailedPrecondition scope.** Fixed — narrowed the file doc-comment to name `ListWorkflowExecutions`/`GetWorkflowExecution` as the exceptions that read from `workflowStore` directly, per your `multi_agent.go` line references.

**Minor — `Wrap`'s godoc omits its nil behavior.** Fixed — added a paragraph to `Wrap`'s godoc documenting `Wrap(nil) == nil`.

**Minor — error-wrapping style split.** Left as-is. You already noted this isn't a clean violation since `client.go` itself is split between the older bare-passthrough convention (`GetSession`, `ListSessions`) and the newer `fmt.Errorf("failed to X: %w", err)` convention (MCP servers, models, artifacts). `schedules.go` follows the older half deliberately, matching the file's closest siblings (`Session`/`Weave`-era methods) rather than picking a side in an inconsistency this PR didn't create. Happy to revisit as part of a dedicated cleanup PR across the whole file rather than as a one-off in a schedules-only change.

## @InfinityOverflow

Thanks for verifying the nil-preserving/limit-default/full-object-return/`Wrap` behavior against the server independently — that overlaps with ilsiepotamus's majors above and both are now fixed the same way (request-assertion tests for `GetSchedule`/`GetWorkflowExecution`). Agreed on `Wrap`/`borrowed` being a disclosed-but-unrelated rider; left it as is per your own "fine, just flagging" note. No hawk tracing added — still consistent with every other method in this package, so not introducing a new gap.

## Verification

- `go build -tags fts5 ./...`, `go vet -tags fts5 ./...` — clean
- `gofmt -l` on all 3 changed files — clean
- `go test -tags fts5 -race -count=1 ./pkg/tui/client/...` — all pass, including the new tests
- `go test -tags fts5 -race -timeout=10m ./pkg/... ./internal/... ./cmd/...` and `./examples/...` (the CI commands) — 0 failures, 0 data races
- `buf lint`, `buf breaking --against '.git#branch=origin/main'` — clean (still no proto/gen changes, confirmed via empty `git diff origin/main...HEAD -- gen/go/ proto/`)

Force-pushed with lease pinned to the branch's prior head (`863b9569`).

ilsiepotamus (approved):
Re-reviewed at `61e9e136` (up from `863b9569`). Confirmed this isn't a rebase-only push: the merge commit (`a370127c`) picks up `main`'s drift with zero conflicts and doesn't touch the schedule surface at all, but the second commit (`61e9e136`, "fix(client): address schedule-surface review feedback") is real, scoped work against exactly the three files this review covered.

**Major 1 — `GetSchedule`/`GetWorkflowExecution` at 0% coverage: fixed.** `TestGetSchedule` (schedules_test.go:328) and `TestGetWorkflowExecution` (schedules_test.go:346) now exist, backed by new mock handlers and recorder fields, and assert the ID reaches the wire rather than just checking the return value. Verified independently, not just via the commit message: `go tool cover -func` shows `GetSchedule` (schedules.go:69) and `GetWorkflowExecution` (schedules.go:213) both at 100.0%.

**Major 2 — `ListSchedules` had no pagination: fixed.** It now takes `pageSize int32, pageToken string` and returns `(schedules, nextPageToken, error)` (schedules.go:53), matching `ListWorkflowExecutions`'s shape. `TestListSchedulesPaging` (schedules_test.go:182) asserts both directions round-trip. I grepped for other callers of `Client.ListSchedules` before signing off on the breaking signature change — there are none outside this package's own tests, so this really was free to do now rather than later.

**Minor — FailedPrecondition doc-comment overclaim: fixed.** schedules.go:23-29 now correctly carves out `ListWorkflowExecutions`/`GetWorkflowExecution` as not gating on scheduler presence.

**Minor — `Wrap`'s nil contract undocumented: fixed.** client.go:174-177 now documents `Wrap(nil) == nil`.

**Minor — error-wrapping style split: left as-is, and I'm fine with that.** I called this "not a clean violation" originally since `client.go` was already split on convention before this PR touched it. Deferring the file-wide cleanup rather than picking a side as a one-off here is the right call — not asking for anything further.

No new issues from this push — the diff is scoped to exactly `client.go`, `schedules.go`, and `schedules_test.go`.

Re-verified independently: `go build/vet -tags fts5 ./...` clean, `gofmt -l` clean, `golangci-lint run ./pkg/tui/client/...` 0 issues, `go test -tags fts5 -race -count=1 -cover ./pkg/tui/client/...` all pass with both previously-0% methods now at 100%, and `gen/go/`+`proto/` still show zero diff against `main`.

Both majors resolved and independently verified. Approving.

ilsiepotamus (approved):
Re-checked at `d4263583` — empty diff against the head I approved (`61e9e136`), same content just re-committed with proper authorship (`josh.schoen@teradata.com`). Re-affirming approval, no fresh review needed since nothing about the reviewed content changed.

PR #366, opened 2026-08-25
fix(sqlite): render per-connection PRAGMAs into the DSN so every pooled connection gets them

ilsiepotamus:
## The bug (measured at 512-agent scale)

`PRAGMA busy_timeout` is **per-connection**, but it was being set via a one-shot `db.Exec` on a pooled `*sql.DB` — configuring exactly one connection and leaving the rest of the pool at 0, failing instantly with SQLITE_BUSY under write contention. Tonight's az512m fleet run (the first at full concurrency after the scheduler-stack merge removed the accidental serialization that had been masking this) recorded **291 lost message rows** ("failed to save message: database is locked") and **~27,000 eval-run write failures**.

Two empirical findings that shaped the fix:
- The CGO driver (go-sqlcipher) **defaults** every connection to busy_timeout=5000, which is why dev machines never reproduced this — only `CGO_ENABLED=0` builds (i.e., exactly what fleets ship) run the modernc driver, which has **no default busy handler**.
- modernc only started honoring mattn-style DSN params (`_busy_timeout`, `_journal_mode`, `_fk`) in **v1.55.0** — the mattn-style DSNs already in `pkg/storage/backend` were silently inert on older nocgo builds.

## The fix

- New `sqlitedriver.DSN(path, Options{BusyTimeoutMS, WAL, ForeignKeys})` with **build-tagged rendering**: mattn syntax under CGO, modernc-native `_pragma=...` under nocgo (the guaranteed contract on every modernc release). Preserves caller query params, handles `:memory:`, and leaves an empty path untouched (an empty path is SQLite's private temp DB; appending params would create a literal `?_busy_timeout=...` file).
- All 14+ production open sites converted at busy_timeout=5000 with each site's existing WAL/FK intent moved into the DSN; per-connection-broken `Exec` pragmas removed (journal_mode Execs kept where genuinely correct — it's a persistent database-level setting). `:memory:` payload store pinned to `SetMaxOpenConns(1)` (a pooled `:memory:` DSN is a different empty DB per physical connection). `NewMigrator`'s exported signature untouched (loom-cloud implements against it); its Exec stays as a documented best-effort net for external callers.

## Regression evidence

`TestDSNBusyTimeoutOnEveryPooledConn` (uses 7500 so the CGO default can't mask the old behavior) pins 4 distinct pooled connections: old approach under nocgo reads `7500, 0, 0, 0` — the exact production failure; the DSN approach reads 7500 + WAL + FK on all four. `TestDSNConcurrentWritersSurviveContention`: 4 writers × 50 inserts, zero SQLITE_BUSY. Both verified under CGO (`-race`) and `CGO_ENABLED=0`.

Follow-ups noted in code, deliberately not done here: the observability store runs a 10-connection pool in rollback-journal mode (writers serialize on the file lock — WAL or a single-writer split would cut contention structurally), and the encrypted-DB `PRAGMA key` Exec has the same pooled-connection hazard for pools >1.

## Validation

`gofmt`/`go vet` clean; `go build -tags fts5 ./...` under both CGO and `CGO_ENABLED=0`; `-race -count=2` across all 19 touched-store packages; sqlitedriver suite green under both drivers.

### DIFF
--- a/cmd/looms/cmd_eval.go
+++ b/cmd/looms/cmd_eval.go
@@ -25,7 +25,7 @@ import (
 
 	"github.com/spf13/cobra"
 	loomv1 "github.com/teradata-labs/loom/gen/go/loom/v1"
-	_ "github.com/teradata-labs/loom/internal/sqlitedriver" // SQLite driver
+	"github.com/teradata-labs/loom/internal/sqlitedriver"
 	"github.com/teradata-labs/loom/pkg/agent"
 	"github.com/teradata-labs/loom/pkg/evals"
 	"github.com/teradata-labs/loom/pkg/evals/judges"
@@ -222,8 +222,10 @@ func runEval(cmd *cobra.Command, args []string) {
 	// Create pattern tracker for judge metrics if multi-judge is configured
 	var patternTracker *learning.PatternEffectivenessTracker
 	if suite.Spec.MultiJudge != nil && judgeOrch != nil {
-		// Open database connection for pattern tracking
-		db, err := sql.Open("sqlite3", evalStoreDB)
+		// Open database connection for pattern tracking. busy_timeout rides in
+		// the DSN so every pooled connection waits on lock contention instead
+		// of failing instantly with SQLITE_BUSY.
+		db, err := sql.Open("sqlite3", sqlitedriver.DSN(evalStoreDB, sqlitedriver.Options{BusyTimeoutMS: 5000}))
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "⚠️  Warning: Failed to open database for pattern tracking: %v\n", err)
 			fmt.Fprintf(os.Stderr, "    Judge metrics will not be recorded for learning loop\n")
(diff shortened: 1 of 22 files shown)


ilsiepotamus:
**Verified under real fleet load before merge** (az512o: 384 seat-holding agents + 128 one-shot agents against a live Teradata, same workload that produced the original failure).

| | az512n (main, no fix) | az512o (this PR) |
|---|---|---|
| `Failed to persist message` — lost conversation rows | **291** | **0** |
| `failed to store eval run` — observability writes | ~27,000 | 1,486 |

**The data-integrity failure class is eliminated**, and total lock errors fell 94%. Every remaining failure is the observability store — exactly the follow-up this PR documents in code: it runs a 10-connection pool in rollback-journal mode, so writers serialize on the file lock and can still exhaust even a 5s `busy_timeout` under fleet load. The DSN fix means they now *wait* rather than fail instantly (94.5% of them succeed); the structural fix is WAL and/or a single-writer split, filed separately.

Merging on that evidence.

PR #367, opened 2026-08-25
feat(artifacts): put session scoping on the wire

Josh-Schoen:
Loom keeps **one artifact catalog per server**, shared by every surface that talks to it — and the store has been session-aware since sessions were introduced: `artifacts.Filter` has a `SessionID` field, `GetByName` takes a session, and every artifact row records which session produced it.

**None of that is on the wire.** `ListArtifactsRequest` cannot express a session filter, `GetArtifactRequest` can only scope a name lookup through the session carried in the call context, and the `Artifact` message itself has no `session_id` — so a remote client cannot even filter a full listing for itself.

The consequence: a surface talking to a shared `looms` cannot answer *"which files did this session produce?"* — the first question every file panel asks. This applies to `loom-mcp` too: `loom_list_artifacts` gives MCP hosts the same session-blind view.

## Changes

`session_id` added to `Artifact`, `ListArtifactsRequest`, and `GetArtifactRequest`, wired through:

- **`ListArtifacts`** maps a non-empty `session_id` onto the store filter it already had. Unset means unfiltered — existing behaviour exactly.
- **`GetArtifact`** prefers an explicit `session_id` over the call-context session for name lookups. Names are only unique per session, so a bare name from a remote caller was ambiguous. **ID lookups ignore session** — IDs are global, and scoping them would only manufacture spurious not-founds.
- **A request-supplied `session_id` is authorized before it selects anything.** It is caller-declared, unlike a context session id, so both RPCs run it through `MultiAgentServer.authorizeSessionScope`, which applies `sessionAccessibleBy` — the same per-user isolation predicate every other session-scoped RPC uses. Denial is `NotFound`, so "exists but not yours" is indistinguishable from "no such session". Under ownership enforcement the session must be the caller's; single-tenant deployments stay permissive, exactly as before. The context-derived session id remains trusted and unchecked.
- **A soft-deleted session still resolves for its owner.** Postgres `DeleteSession` only sets `deleted_at` and the artifact rows outlive it, while `LoadSession` and `SessionExists` both filter `deleted_at IS NULL` — so ownership falls back to `SessionStore.CallerOwnsSession`, an owner-scoped existence check that ignores `deleted_at`. Without it the gate refused an owner the session-filtered view of artifacts the store still returns unfiltered, for the whole grace window. It is wired as an optional capability rather than a `SessionStorage` method, since that interface is implemented outside this repo.
- **`artifactToProto`** carries `SessionID` out, so responses say which session each artifact belongs to.

SDK gains the two session-scoped calls (`ListSessionArtifacts`, `GetArtifactByName`); the general artifact methods already existed.

`loom-mcp` now advertises `session_id` on the `loom_list_artifacts` and `loom_get_artifact` schemas, so a model driving them can actually reach the field. `loom_get_artifact` also declared `id` **required**, which made the name lookup undrivable from the schema even though the RPC has always accepted either — a name lookup is precisely what `session_id` is for, so `id` is now optional.

**Not in scope**, and now stated in the architecture doc rather than left implied: ID lookups (`GetArtifact` by id, `GetArtifactContent`, `DeleteArtifact`) are unscoped by design, and the write path (`UploadArtifact`) is not session-aware. Session scoping here is a filtering and name-resolution boundary, not an access-control boundary on IDs.

## Testing

The fixture seeds two sessions that **both contain a `report.md`** — the name collision is precisely the case the explicit session exists for — plus a session-less user upload. Tests assert filtered, unfiltered (unchanged), scoped-name, and by-ID behaviour against a real SQLite store built through the storage backend, FK constraints and all.

Three denial tests cover the boundary the happy-path cases cannot: a foreign session is refused on both the listing and the name path, and a blank identity is refused under enforcement. Each was confirmed to **fail without the authorization check**. Alongside them: the owner still reads her own session (so the check is an isolation boundary, not a feature switch), single-tenant mode stays permissive (regression guard for the default), and the ID carve-out is pinned as deliberate.

`pkg/storage/postgres/artifact_session_scope_test.go` pins that `session_id` composes with — never substitutes for — the store's unconditional `user_id = $1` predicate, on both `List` and `GetByName`. It is deliberately **untagged**: every `//go:build integration` file in that package is invisible to CI, which runs `-tags fts5` only, so a tagged test would have been dead coverage. It follows `human_request_store_test.go` instead and skips without `TEST_POSTGRES_URL`, so CI is the first place it runs.

The two SDK wrappers are covered against a bufconn stub that records the request, since putting `session_id` on the wire is the only reason they exist: a wrapper that dropped it would still compile and still return artifacts. Those tests also pin that `limit: 0` stays 0 (the server applies its default, not the client) and that the name path sends no `id`, which would otherwise make the server ignore the session.

`gofmt`, `go vet`, `golangci-lint` (0 issues), `buf lint` clean. **`buf breaking --against main` clean** — three new optional fields, nothing modified. `buf generate` re-run with no diff against the committed generated files. `pkg/server`, `pkg/artifacts`, `pkg/tui/client`, `pkg/storage/postgres`, and `pkg/mcp/server` pass under `-race` with `-count=1`.

## Rebase

Replayed onto current `main` (was 20 commits behind) — no conflicts. The three field numbers (`Artifact` 17, `ListArtifactsRequest` 7, `GetArtifactRequest` 3) were re-verified as still next-available on current `main`, and the base's own changes to `pkg/artifacts/store.go` (a SQLite DSN/pragma refactor) and `loom.proto` (comments on `WeaveProgress`) leave this change's dependencies intact.

Independent of #363/#364 — branches directly from `main`.

### DIFF
--- a/docs/architecture/artifacts.md
+++ b/docs/architecture/artifacts.md
@@ -1306,6 +1306,19 @@ session-scoped RPC uses.
 Denial is reported as `NotFound`, not `PermissionDenied`, so a caller cannot
 distinguish "exists but not yours" from "no such session" by probing.
 
+**Soft-deleted sessions stay readable by their owner.** Postgres `DeleteSession`
+only sets `deleted_at`, and a session's artifacts outlive it — `purge_soft_deleted`
+reaps artifacts by their *own* `deleted_at`, and the `artifacts.session_id`
+CASCADE fires only on the later hard purge. Because `LoadSession` and
+`SessionExists` both filter `deleted_at IS NULL`, neither can tell an owner's
+just-deleted session from one that never existed. Ownership therefore falls back
+to `SessionStore.CallerOwnsSession`, an owner-scoped existence check that ignores
+`deleted_at`, so the owner keeps the session-filtered view of artifacts the store
+still returns unfiltered. It is exposed as an optional capability
+(`sessionOwnershipProbe`) rather than a `SessionStorage` method, because that
+interface is implemented outside this repo; a store without it keeps the
+fail-closed behaviour.
+
 Two limits worth stating plainly, because "session scoping" reads broader than what
 is enforced:
 
--- a/docs/architecture/artifacts.md
+++ b/docs/architecture/artifacts.md
@@ -1319,6 +1319,14 @@ still returns unfiltered. It is exposed as an optional capability
 interface is implemented outside this repo; a store without it keeps the
 fail-closed behaviour.
 
+Only the Postgres store needs this, and so only it implements the capability.
+SQLite hard-deletes sessions and its `artifacts.session_id` foreign key is
+`ON DELETE CASCADE`, so a deleted session's artifacts are already gone — there is
+nothing to be locked out of. The two backends therefore answer the same request
+differently in *shape*: with enforcement on, scoping to a deleted session yields
+an empty list on Postgres and `NotFound` on SQLite. Both mean "no artifacts";
+a client that distinguishes the two should treat `NotFound` as empty here.
+
 Two limits worth stating plainly, because "session scoping" reads broader than what
 is enforced:
 
--- a/docs/architecture/artifacts.md
+++ b/docs/architecture/artifacts.md
@@ -823,7 +823,9 @@ LIMIT 100;
 **Rationale**:
 - **Search**: Session-scoped by default for isolation
 - **List**: Only show artifacts in current session (avoid clutter)
-- **Read**: Session-scoped for security (no cross-session access)
+- **Read**: Session-scoped for the agent tool path (no cross-session access); on the
+  gRPC API an explicit `session_id` is accepted and authorized against the caller
+  (see [Session Isolation](#session-isolation))
 - **API-level**: Cross-session search available for server-side operations
 
 ---
@@ -1274,9 +1276,9 @@ db.Exec("PRAGMA foreign_keys=ON")
 3. **Storage-level**: SQLite queries filtered by session_id
 4. **Filesystem-level**: Paths constructed from validated session ID only
 
-**Bypass Prevention**:
+**Bypass Prevention (agent tool path)**:
 ```go
-// Agent cannot override session ID via tool parameters
+// The agent cannot override session ID via tool parameters
 // Session ID comes from context, not user-controllable input
 
 func (t *WorkspaceTool) Execute(ctx context.Context, params map[string]interface{}) (*shuttle.Result, error) {
@@ -1284,6 +1286,34 @@ func (t *WorkspaceTool) Execute(ctx context.Context, params map[string]interface
     // params["session_id"] is ignored - session ID always from context
 }
 ```
+
+This holds for the in-process agent tool (`pkg/shuttle/builtin/workspace.go`), which
+is the path an agent drives. It is **not** a statement about the gRPC API — see below.
+
+**gRPC API path (`ListArtifacts`, `GetArtifact`)**: these accept an explicit
+`session_id` on the request, because a remote surface has no session in its call
+context and still has to answer "which files did this session produce?". A
+request-supplied session id is caller-declared rather than server-derived, so it is
+authorized before it selects anything: `MultiAgentServer.authorizeSessionScope`
+runs the same per-user isolation predicate (`sessionAccessibleBy`) that every other
+session-scoped RPC uses.
+
+| Deployment | Effect of an explicit `session_id` |
+|---|---|
+| Ownership enforcement on (authenticated) | Must resolve to a session the caller owns, or the call returns `NotFound`. A blank identity is never a wildcard. |
+| Single-tenant (default, unauthenticated) | Permitted, matching the mode's documented trust model — `looms serve` states all data is accessible to all callers and there is no per-user isolation. |
+
+Denial is reported as `NotFound`, not `PermissionDenied`, so a caller cannot
+distinguish "exists but not yours" from "no such session" by probing.
+
+Two limits worth stating plainly, because "session scoping" reads broader than what
+is enforced:
+
+- **ID lookups are not session-scoped.** `GetArtifact` by `id`, `GetArtifactContent`,
+  and `DeleteArtifact` ignore session entirely — artifact IDs are global, and
+  scoping them would only manufacture spurious not-founds. Session scoping is a
+  filtering and name-resolution boundary, not an access-control boundary on IDs.
+- **The write path is not session-scoped.** `UploadArtifact` has no `session_id`.
 
 **Coordinator Exception**: Agents configured with `restrictReads=all_sessions` can read across sessions for research purposes, but write access remains session-scoped.
 
--- a/gen/go/loom/v1/loom.pb.go
+++ b/gen/go/loom/v1/loom.pb.go
@@ -9443,7 +9443,12 @@ type Artifact struct {
 	// Additional metadata (JSON-serializable key-value pairs)
 	Metadata map[string]string `protobuf:"bytes,15,rep,name=metadata,proto3" json:"metadata,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
 	// Soft delete timestamp (Unix seconds, optional)
-	DeletedAt     int64 `protobuf:"varint,16,opt,name=deleted_at,json=deletedAt,proto3" json:"deleted_at,omitempty"`
+	DeletedAt int64 `protobuf:"varint,16,opt,name=deleted_at,json=deletedAt,proto3" json:"deleted_at,omitempty"`
+	// Session this artifact belongs to. Empty for artifacts created outside a
+	// session (user uploads, pre-session tooling). The store has recorded this
+	// since sessions were introduced; it was simply never put on the wire, which
+	// left remote clients unable to tell whose files they were looking at.
+	SessionId     string `protobuf:"bytes,17,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
@@ -9590,6 +9595,13 @@ func (x *Artifact) GetDeletedAt() int64 {
 	return 0
 }
 
+func (x *Artifact) GetSessionId() string {
+	if x != nil {
+		return x.SessionId
+	}
+	return ""
+}
+
 // ListArtifactsRequest lists artifacts with optional filtering.
 type ListArtifactsRequest struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
@@ -9605,8 +9617,14 @@ type ListArtifactsRequest struct {
 	Offset int32 `protobuf:"varint,5,opt,name=offset,proto3" json:"offset,omitempty"`
 	// Include soft-deleted artifacts
 	IncludeDeleted bool `protobuf:"varint,6,opt,name=include_deleted,json=includeDeleted,proto3" json:"include_deleted,omitempty"`
-	unknownFields  protoimpl.UnknownFields
-	sizeCache      protoimpl.SizeCache
+	// Filter to artifacts belonging to one session (optional). Unset means no
+	// session filtering, preserving existing behaviour. This is what lets a
+	// remote surface render "the files this session produced" — the local
+	// artifact store has always supported it (artifacts.Filter.SessionID); the
+	// request message could not express it.
+	SessionId     string `protobuf:"bytes,7,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *ListArtifactsRequest) Reset() {
@@ -9681,6 +9699,13 @@ func (x *ListArtifactsRequest) GetIncludeDeleted() bool {
 	return false
 }
 
+func (x *ListArtifactsRequest) GetSessionId() string {
+	if x != nil {
+		return x.SessionId
+	}
+	return ""
+}
+
 // ListArtifactsResponse returns a list of artifacts.
 type ListArtifactsResponse struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
@@ -9742,7 +9767,12 @@ type GetArtifactRequest struct {
 	// Artifact ID (primary lookup)
 	Id string `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty"`
 	// Artifact name (alternative lookup if ID not provided)
-	Name          string `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
+	Name string `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
+	// Session to scope a name lookup to (optional). Names are only unique within
+	// a session, so a bare name lookup from a remote client is ambiguous without
+	// this. Ignored when id is set. Unset falls back to the session carried in
+	// the call context, preserving existing behaviour for in-session callers.
+	SessionId     string `protobuf:"bytes,3,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
@@ -9787,6 +9817,13 @@ func (x *GetArtifactRequest) GetId() string {
 func (x *GetArtifactRequest) GetName() string {
 	if x != nil {
 		return x.Name
+	}
+	return ""
+}
+
+func (x *GetArtifactRequest) GetSessionId() string {
+	if x != nil {
+		return x.SessionId
 	}
 	return ""
 }
@@ -11539,7 +11576,7 @@ const file_loom_v1_loom_proto_rawDesc = "" +
 	"\vtotal_count\x18\x02 \x01(\x05R\n" +
 	"totalCount\x12\x1f\n" +
 	"\vserver_name\x18\x03 \x01(\tR\n" +
-	"serverName\"\xb2\x04\n" +
+	"serverName\"\xd1\x04\n" +
 	"\bArtifact\x12\x0e\n" +
 	"\x02id\x18\x01 \x01(\tR\x02id\x12\x12\n" +
 	"\x04name\x18\x02 \x01(\tR\x04name\x12\x12\n" +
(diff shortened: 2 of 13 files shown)


ilsiepotamus (changes requested):
# PR #367 Review: "feat(artifacts): put session scoping on the wire"

**Setup notes**: Single commit (`15ade12e`) on top of `main`, author Josh-Schoen. `gh pr view` shows `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` — not conflicting, just needs a rebase/update to pick up newer main commits; nothing structural. Verified in worktree: `go build -tags fts5 ./...`, `go vet -tags fts5 ./...`, `gofmt -l`, `golangci-lint run` (0 issues), `buf lint`, `buf breaking --against main`, and `buf generate` (re-run and diffed byte-for-byte against the PR's committed generated files — identical) all pass clean. `go test -tags fts5 -race -count=1` green on `pkg/server`, `pkg/tui/client`, `pkg/artifacts`, `pkg/storage/postgres`.

---

## 1. DESCRIPTION

"Session scoping" here means **filtering**, not access control — the PR itself frames it that way ("render the files this session produced"), which matters for section 3.

**Before**: `Artifact` carried no `session_id` on the wire at all — a caller could not tell which session owned a returned artifact. `ListArtifactsRequest` had no session filter — every `ListArtifacts` call already returned every artifact visible to the caller (all sessions), it just couldn't be narrowed. `GetArtifactRequest`'s by-name lookup scoped itself only via `session.SessionIDFromContext(ctx)` — and no interceptor in `pkg/server/interceptors.go` ever populates that context key from gRPC metadata (the only call site of `session.WithSessionID` in the whole repo is `pkg/agent/agent.go:1792`, inside the server-side LLM conversation loop). So for a real external gRPC/TUI/MCP caller, that value was always `""`, and a bare name lookup matched only session-less artifacts.

**After**: `Artifact.session_id` (proto field 17), `ListArtifactsRequest.session_id` (field 7), `GetArtifactRequest.session_id` (field 3) are added. `ListArtifacts` now optionally filters by session (unset = unfiltered, unchanged); every returned `Artifact` states its owning session; `GetArtifact`'s by-name lookup now accepts an **explicit, client-supplied** `session_id` that takes priority over the (previously sole, always-empty-for-remote-callers) context value — `pkg/server/artifacts.go:94-100`. ID-based lookups (`GetArtifact` by id, `GetArtifactContent`, `DeleteArtifact`) are untouched — they ignore session entirely, before and after.

**Files/subsystems touched**: `proto/loom/v1/loom.proto` (3 new fields), `gen/go/loom/v1/loom.pb.go` + `gen/openapiv2/loom/v1/loom.swagger.json` (regenerated), `pkg/server/artifacts.go` (`ListArtifacts`, `GetArtifact`, `artifactToProto`), new `pkg/server/artifacts_session_test.go`, new `pkg/tui/client/artifacts.go` (two SDK wrappers: `ListSessionArtifacts`, `GetArtifactByName`). **Not touched** (relevant by omission): `pkg/artifacts/store.go` and `pkg/storage/postgres/artifact_store.go` (both already had session-aware signatures pre-PR — this PR only exposes a wire path to values they already accepted); `pkg/shuttle/builtin/workspace.go` (the in-process agent tool — always derives session from trusted context, never a parameter); `pkg/mcp/server/bridge_tools.go`/`bridge_handlers.go` (the `loom-mcp` surface the PR's own description names as motivation); `docs/architecture/artifacts.md`, `docs/guides/artifacts-usage.md`.

---

## 2. SHAPE REVIEW

Read `CLAUDE.md` (root) and the nested `docs/architecture/CLAUDE.md` in full.

**Proto is law — followed well.** `.proto` changed first (`proto/loom/v1/loom.proto:2664,2692,2716`), each new field is correctly the next-available tag in its message (Artifact was at 16 → 17; `ListArtifactsRequest` at 6 → 7; `GetArtifactRequest` at 2 → 3) — verified by reading each message definition in full, no renumbering. I independently re-ran `buf generate` in the worktree and diffed the result against the PR's committed `gen/go/loom/v1/loom.pb.go` and `gen/openapiv2/loom/v1/loom.swagger.json`: **byte-identical**, confirming the generated code wasn't hand-edited/stale. `buf lint` and `buf breaking --against .git#branch=main` both ran clean, matching the PR body's claim — three purely additive optional fields, nothing removed or retyped.

**gRPC-over-HTTP preference**: respected — no new HTTP-only path; the swagger diff (`gen/openapiv2/loom/v1/loom.swagger.json:618-631,712-731,4226-4232`) is a mechanical, additive consequence of the proto change via grpc-gateway, not hand-authored REST.

**Reuse over new mechanism — done well.** The wire fields map directly onto pre-existing internal abstractions: `artifacts.Filter.SessionID` (`pkg/artifacts/store.go:77`) and `ArtifactStore.GetByName(ctx, name, sessionID string)` (`pkg/artifacts/store.go:405`, `pkg/storage/postgres/artifact_store.go:147`) both already existed before this PR. No parallel scoping mechanism was invented. The two new `pkg/tui/client/artifacts.go` SDK methods follow the exact same thin-wrapper convention as the eight pre-existing artifact client methods in `pkg/tui/client/client.go:639-773`.

**Table-driven tests**: partially followed. `TestGetArtifactByNameScopedToSession` (`pkg/server/artifacts_session_test.go:114-133`) is a proper `[]struct{session, wantID}` table; the other three tests are single-scenario, which is reasonable given each covers a distinct case.

**Nothing flagged against CLAUDE.md's proto/buf/build rules** — this is the strongest part of the PR. The one process gap is documentation currency, covered next.

---

## 3. COMPREHENSIVE REVIEW (security-focused)

**What's actually enforced, and untouched by this PR**: `docs/reference/security-model.md:190` documents `artifacts` as RLS-protected via a `user_id` column; I traced this to the live policy — `pkg/storage/postgres/migrations/000016_rls_honor_jwt_sub.up.sql:47-50` (`CREATE POLICY artifacts_user_isolation ON artifacts USING (user_id = loom_current_user_id())`), enforced with `FORCE ROW LEVEL SECURITY` at the database level, doubled by an app-level predicate in every Postgres query: `pkg/storage/postgres/artifact_store.go:131-133` (`Get`), `:162-175` (`GetByName`), `:196` (`List`, unconditional `WHERE user_id = $1` base clause — session_id is additive on top, never a substitute). `userID` there comes from `UserIDFromContext(ctx)`, populated by an auth interceptor (`pkg/server/interceptors.go`) from a verified JWT subject or `x-user-id` header — not from anything this PR added. **This boundary is completely unaffected by the new `session_id` field**: a caller cannot use it to cross a `user_id`/tenant boundary in Postgres mode, confirmed at both the query and RLS-policy level.

**What is *not* enforced anywhere — the actual finding.** `pkg/server/artifacts.go:55-57` and `:94-100` accept `req.SessionId` verbatim, with **no check that the caller has any relationship to that session** — no session-ownership lookup, no comparison against the caller's own context-derived session. This is contained today only because:
- In **Postgres/multi-tenant mode**, the independent `user_id` boundary above happens to sit underneath it — a caller can at most reach a *different session of their own account*, never another tenant's.
- In **SQLite/single-tenant mode** (`pkg/artifacts/store.go` — no `user_id` column at all), there is no containment whatsoever, but this matches the mode's own documented, explicit trust model: `cmd/looms/cmd_serve.go:2281-2295` states "All data is accessible to all callers... There is no per-user data isolation" for SQLite. Not a new hole — but the PR does newly make session ownership *wire-visible and directly addressable* (`ListArtifacts{}` unfiltered now returns every artifact's `session_id`, and `GetArtifact{name, session_id}` lets any caller target a specific foreign session by name) where before this information required inference.

This is a design smell independent of today's topology: a handler that trusts a client-declared scope value with zero ownership check is exactly the pattern that becomes a real vulnerability the next time it's reused somewhere lacking the (unrelated) `user_id` safety net underneath it.

**Documentation now states a false invariant.** `docs/architecture/artifacts.md:1268-1288` ("Session Isolation" / "Bypass Prevention") reads: *"Agent cannot override session ID via tool parameters / Session ID comes from context, not user-controllable input... params['session_id'] is ignored."* And `docs/architecture/artifacts.md:826`: *"Read: Session-scoped for security (no cross-session access)."* This PR adds precisely the override that text says doesn't exist — on the new gRPC `GetArtifact`/`ListArtifacts` path (the doc's invariant is true only for the untouched `WorkspaceTool`, `pkg/shuttle/builtin/workspace.go:83`, which still derives session solely from trusted context). The doc's own escape-hatch note at line 821 ("Cross-session search is available at the API level for coordinator agents or server-side operations") anticipated something narrower than what this PR ships (any remote gRPC/TUI/MCP caller, not just server-side coordinators). The doc is not updated, despite `docs/architecture/CLAUDE.md`'s explicit maintainer instruction: *"Keep current: Update architecture docs when design changes."*

**Tests are happy-path only — no denial assertions.** All four tests in `pkg/server/artifacts_session_test.go` (`TestListArtifactsFiltersBySession:77`, `TestListArtifactsUnfilteredIsUnchanged:99`, `TestGetArtifactByNameScopedToSession:114`, `TestGetArtifactByIDIgnoresSession:137`) assert a caller **can** get what it asks for; none assert a caller is **denied** cross-session access. `TestGetArtifactByIDIgnoresSession` (lines 140-146) in fact *proves* the opposite by design: a request tagged `SessionId: "sess-1"` successfully retrieves artifact `b1`, which belongs to `sess-2`. All four tests run only against `setupArtifactServer`'s `SQLiteStore` fixture (lines 31-71) — none exercise `pkg/storage/postgres/artifact_store.go`, so there is zero regression coverage confirming the new `session_id` parameter continues to compose safely with the `user_id` predicate that is the only thing actually containing it in production multi-tenant mode.

**MCP surface not actually fixed despite being named as motivation.** The PR body states the problem applies to `loom-mcp`'s `loom_list_artifacts` too, but `pkg/mcp/server/bridge_tools.go:378-386` (the `loom_list_artifacts`/`loom_get_artifact` tool schemas advertised to an LLM) was not updated to list `session_id` as a parameter. The generic `callGRPC` helper (`pkg/mcp/server/bridge.go:422-462`, using `protojson.Unmarshal` on raw args at line 437) means the field is *silently functional* if a caller happens to know to pass it — but undiscoverable to a model driving the tool from its published schema, so the PR's own stated problem remains practically unaddressed for that surface.

**Scope gaps (informational, not blocking)**: `UploadArtifact` (`pkg/server/artifacts.go:118-343`) still has no session awareness — no `session_id` on `UploadArtifactRequest`, nothing set on the created artifact struct; `GetArtifactContent` and `DeleteArtifact` (`:396-463`, `:346-362`) remain fully ID-only/unscoped, unchanged. Consistent with the PR's stated (read-only) scope, but worth naming since "session scoping" as a feature title could otherwise be read as broader than what shipped.

---

**Tally**: 0 blockers, 2 majors, 3 minors.

- Major 1: `session_id` on `GetArtifact`/`ListArtifacts` is an unauthenticated, self-declared scope with no ownership check (`pkg/server/artifacts.go:55-57,94-100`); safe today only via an unrelated, independent enforcement layer.
- Major 2: `docs/architecture/artifacts.md:1268-1288,826` now documents a security invariant this PR makes false for the gRPC surface, and isn't updated.
- Minor 1: Zero negative/denial test coverage; no test touches the Postgres-backed store with the new field.
- Minor 2: `loom-mcp` tool schemas (`pkg/mcp/server/bridge_tools.go:378-386`) not updated despite being named as motivation.
- Minor 3: Write path (`UploadArtifact`) and content/delete paths remain entirely outside session scoping — fine given stated scope, worth flagging for completeness.

VERDICT: CHANGES NEEDED

Josh-Schoen:
Thanks — the review found a real hole, and the fix turned out to be reuse rather than new mechanism. Rebased onto `main` (was 20 behind; no conflicts, and the three field numbers 17/7/3 are still next-available on current `main`, re-verified) and addressed all five points.

## Major 1 — self-declared session scope: **fixed**

Confirmed, and the sharpest form of it is that `sessionAccessibleBy` documents itself as *"the per-user isolation predicate shared by every session-scoped RPC"* — this PR added a session-scoped read path and was the one caller that skipped it.

So rather than invent a check, `ListArtifacts`/`GetArtifact` now go through a new `MultiAgentServer.authorizeSessionScope` (`pkg/server/multi_agent.go`), which resolves the session the way `DeleteSession` does (in-memory agents, then the session store) and runs `sessionAccessibleBy` on it. Details worth flagging:

- **Only the request field is authorized.** The context session id stays trusted — it is server-derived. An explicit `session_id` is the caller choosing whose namespace to read.
- **Denial is `NotFound`, not `PermissionDenied`**, matching `DeleteSession`, so a caller cannot separate "exists but not yours" from "no such session" by probing.
- **An unresolvable id defers to the tenancy mode** rather than a blanket allow or deny: enforcing deployments fail closed (the stance `findSessionOwner` already takes on ids it cannot verify), single-tenant stays permissive. This is what keeps the default local mode behaving exactly as before.

One trap worth recording, since it would have been a regression in its own right: `LoadSession` reports a miss as an **error** on SQLite and as a **nil session** on Postgres, and neither distinguishes foreign from unknown. Treating that error as `Internal` would have turned any unknown session id into a failed call, breaking filtering for exactly the callers this PR is for. A miss therefore falls through to the tenancy decision instead of surfacing.

Your framing that this is "safe today only via an unrelated enforcement layer" was the right read — the Postgres `user_id` predicate is now a second line rather than the only one.

## Major 2 — docs stated a false invariant: **fixed**

`docs/architecture/artifacts.md` now scopes the "Bypass Prevention" claim to the agent tool path where it is actually true (`WorkspaceTool`), and adds a gRPC section stating what the API accepts, the per-deployment effect in a table, and why denial is `NotFound`. Line 826's "Read: Session-scoped for security" is qualified the same way.

It also states the two limits plainly, since "session scoping" reads broader than what is enforced: **ID lookups are not session-scoped** (a filtering and name-resolution boundary, not an access-control boundary on IDs) and **the write path is not scoped**.

## Minor 1 — no denial coverage: **fixed**

Six tests added. The three denial tests were verified to **fail without the authorization check** — before the fix, alice reads bob's listing, resolves bob's `report.md`, and an anonymous caller reads a session under enforcement:

- `TestListArtifactsDeniesForeignSession`, `TestGetArtifactByNameDeniesForeignSession`, `TestAnonymousCallerDeniedUnderEnforcement`
- `TestOwnerStillReadsOwnSessionUnderEnforcement` — so the check is an isolation boundary, not a feature switch
- `TestSingleTenantSessionScopeStillPermitted` — regression guard for the default mode
- `TestGetArtifactByIDRemainsUnscopedUnderEnforcement` — records the ID carve-out as deliberate

On the Postgres request: added `pkg/storage/postgres/artifact_session_scope_test.go`, pinning that `session_id` composes with — never substitutes for — the unconditional `user_id = $1` predicate, on both `List` and `GetByName`.

**Worth knowing about this repo's gate:** I first wrote it as `*_integration_test.go` using the shared `testPool` harness, then found that **every `//go:build integration` file in `pkg/storage/postgres` is invisible to CI** — CI runs `go test -tags fts5` and the `integration` tag appears in no workflow (only in justfile `./test/e2e/...` targets). A tagged test would have been dead coverage. It is therefore untagged with a local fixture, following `human_request_store_test.go`, which is the one Postgres file that actually exercises the CI service container. Honest limitation: **I could not run it locally** (no Docker daemon, no local Postgres) — it skips without `TEST_POSTGRES_URL`, and CI is the first place it executes.

## Minor 2 — MCP surface: **fixed**

`session_id` added to the `loom_list_artifacts` and `loom_get_artifact` schemas in `bridge_tools.go`.

While there: `loom_get_artifact` declared `id` as **required**, which made the name lookup undrivable from a schema-reading model even though the RPC has always accepted either. Since a name lookup is precisely what `session_id` is for, `id` is now optional with "either id or name is required" in the description — otherwise adding the field would have changed nothing in practice.

## Minor 3 — write/content/delete paths: **documented, not expanded**

Left out of scope deliberately, now stated in the architecture doc rather than left implied. Happy to open a follow-up for `UploadArtifact` if you want the write path scoped too.

## One deliberate non-change

Unfiltered `ListArtifacts` still returns every artifact the caller can see, each now carrying its `session_id`. Narrowing that is a behaviour change beyond this PR's promise ("unset means unfiltered — existing behaviour exactly"), and under Postgres the listing is already `user_id`-scoped. Say the word if you'd rather it defaulted to the caller's sessions.

## Verification

`gofmt`, `go build -tags fts5 ./...`, `go vet`, `golangci-lint` (0 issues), `buf lint`, `buf breaking --against main` all clean. `buf generate` re-run: no diff against the committed generated files. `pkg/server`, `pkg/artifacts`, `pkg/tui/client`, `pkg/storage/postgres`, `pkg/mcp/server` pass under `-race -count=1`.

Josh-Schoen:
## Follow-up: an availability bug in the ownership gate, now fixed

A parallel review pass found a real defect in the authorization I added above, on the exact use case this PR exists to serve. Fixed in `f8c81a3b`.

**An owner was locked out of their own just-deleted session's artifacts, on every auth-enabled deployment.** `enforceOwnership` tracks `config.Server.Auth.Enabled` (`cmd/looms/cmd_serve.go:2340`), so this landed on the multi-tenant Postgres deployments the gate is for.

Postgres `DeleteSession` is a **soft** delete, and a session's artifacts outlive it: `purge_soft_deleted` reaps artifacts by their *own* `deleted_at`, and the `artifacts.session_id` CASCADE fires only on the later hard purge. But `LoadSession` (`session_store.go:142`) and `SessionExists` (`:232`) both filter `deleted_at IS NULL`, so neither could distinguish an owner's just-deleted session from one that never existed — both collapsed into my fail-closed branch.

The result was an asymmetry with no defensible reading: for the whole grace window the gate refused an owner the session-**filtered** view of artifacts the store still returned **unfiltered** and by ID. "Which files did this session produce?" failed for a session they had just deleted themselves.

**Fix** — `SessionStore.CallerOwnsSession`, an owner-scoped existence check that ignores `deleted_at`, following `RestoreSession`'s idiom in the same file. Notes on the shape:

- Wired in as an **optional capability** (`sessionOwnershipProbe`), not a `SessionStorage` method. That interface is implemented outside this repo, so adding a method to it would break those implementations; a store without the capability keeps the fail-closed behaviour unchanged.
- **Still owner-scoped**, so a true result proves the *caller's* ownership and never discloses a foreign session's existence. A probe error fails closed.
- `sessionAccessibleBy` now delegates to `ownerAccessibleBy(caller, owner)`, so the probe result runs the identical policy rather than a copy — "a blank identity is never a wildcard" stays in one place. Anonymous callers are refused twice over: the query cannot match `user_id = ''`, and the predicate rejects a blank caller regardless.
- A compile-time assertion pins the postgres store to the capability, so a signature drift fails the build rather than silently reverting the type assertion to `false` and re-locking owners out.

**Tests.** The server test fails without the fix (`owner denied the filtered view of their own soft-deleted session: NotFound`) and asserts the asymmetry — the unfiltered list succeeds in the same test — so it cannot pass for the wrong reason. A foreign id and an unknown id both stay denied. The Postgres test performs a real soft-delete, pins that `LoadSession` and `SessionExists` genuinely miss afterwards, and that a second user is still refused.

SQLite is unaffected: it hard-deletes sessions, so the artifacts cascade away and there is nothing to be locked out of.

One note on framing, since it affects how much the fail-closed branch was ever buying: falling through on an unresolvable id would *also* have been safe here, because the Postgres artifact queries carry an unconditional `user_id = $1`. The only store the branch protects against is one that ignores `user_id` — and such a store already leaks the ungated unfiltered listing. I kept the branch anyway, since it costs nothing now that ownership resolves correctly, but it is defence-in-depth rather than the thing standing between a caller and another tenant's data.

`gofmt`, `go build -tags fts5 ./...`, `go vet`, `golangci-lint` (0 issues) clean; `pkg/server`, `pkg/storage/postgres`, `pkg/artifacts`, `pkg/tui/client`, `pkg/mcp/server`, `pkg/agent` all pass under `-race -count=1`.

ilsiepotamus (approved):
Re-reviewed at `6e7fc143` (was `15ade12e`). This is clean now — approving.

## Major 1 — self-declared session scope: confirmed fixed

`authorizeSessionScope` (`pkg/server/multi_agent.go:675`) is wired into both
`ListArtifacts` (`artifacts.go:60`) and `GetArtifact`'s name branch
(`artifacts.go:111`), and it's a real check, not a cosmetic one: it runs the
same `sessionAccessibleBy` predicate `DeleteSession` uses, denies as
`NotFound`, and fails closed on unresolvable ids under enforcement. I pulled
the worktree and verified this myself rather than taking the commit message
at its word — reverted both call sites and reran
`TestListArtifactsDeniesForeignSession` / `TestGetArtifactByNameDeniesForeignSession`
/ `TestAnonymousCallerDeniedUnderEnforcement` /
`TestListArtifactsStillDeniesForeignUnresolvableSession`: all four failed
without the check, all four pass with it restored. Good catch on the
SQLite+auth generalization too — this isn't just a Postgres fix, since
`enforceOwnership` tracks `Auth.Enabled` independent of backend.

Also traced the soft-delete follow-up: confirmed the `sessions_user_isolation`
RLS policy filters only on `user_id`, not `deleted_at`, so `CallerOwnsSession`
genuinely sees through a soft-delete in real Postgres, not just in the
unit-test stub. The owner-scoped, capability-based design is the right call
given that interface lives outside this repo.

## Major 2 — docs: confirmed fixed

`artifacts.md` now correctly scopes "Bypass Prevention" to the agent-tool
path, adds a gRPC section with the actual enforcement behavior and a
deployment-mode table, and states the two remaining limits (ID lookups,
write path) plainly instead of leaving them implied.

## Minor 1 — denial coverage: confirmed fixed

12 tests now (was 4), 8 of them genuine denial/edge cases, verified
non-vacuous as above. The new Postgres integration test
(`artifact_session_scope_test.go`) is wired into CI's real postgres service,
not silently dead — good call catching the `integration`-tag gap yourself
before I had to.

## Minor 2 — MCP schemas: confirmed fixed

`session_id` advertised on both tools, plus the `id`-required bug fix is a
nice catch beyond what was asked. `bridge_artifact_session_test.go` checks
both the schema and the actual forwarding.

## Minor 3 — write/content/delete scope: documented as intended

No objection to leaving these out of scope — the doc now says so explicitly,
which was the actual ask.

## Verification

`go build/vet -tags fts5 ./...`, `gofmt -l`, `golangci-lint run` (0 issues)
all clean. `go test -tags fts5 -race -count=1` green on `pkg/server`,
`pkg/artifacts`, `pkg/storage/postgres`, `pkg/mcp/server`, `pkg/tui/client`.

One non-blocking nit for a follow-up if you want it: the line at
`artifacts.md:1327` ("empty list on Postgres vs `NotFound` on SQLite for a
deleted session") describes the fully-reaped end state, not the grace-window
state your own tests exercise (where Postgres still returns the artifacts).
Accurate, just a little easy to misread out of context — not blocking.

PR #365, opened 2026-08-25
feat(shuttle): backend-agnostic resource-lease contract drives RESOURCE_HOLDER scheduling

ilsiepotamus:
## Summary

Phase 3a of the fleet-contention work: a **backend-agnostic lease contract** on the tool-result path, and the machinery that turns it into scheduler priority. Loom never learns what a "DB handle" is — the backend declares what's scarce, loom reacts generically.

- **`shuttle.LeaseEvent`** (`acquired`/`released`, opaque backend-defined Kind/ID) riding well-known `Result.Metadata` keys, with tolerant parsing (malformed entries are data, never errors). Any adapter or tool can emit them.
- **`shuttle.BackpressureHint`** hoisted to the shuttle layer — `{code, retry_after_s, wait_param, max_wait_s}`, field-compatible with #355's MCP hint so that adapter's migration is a drop-in. Contract only; the wait loop migrates with #355.
- **`UnmarkResourceHolder` + `WithSlotInfoHolding` + `SlotInfo.Class()`** in the scheduler wiring — the mark is no longer a one-way, single-turn flag.
- **Per-session lease ledger in the agent** (mirrors the existing `sessionToolLedger` lifecycle, retired on session delete/clear): tool results flow through the single execution funnel where lease events update the ledger and mark/unmark the turn; turn start seeds RESOURCE_HOLDER when the session already holds leases — so a conversation that acquired a handle last turn keeps priority inheritance until it releases.

## The money test

`TestLeaseLifecycleAcrossTurns`: a mock tool's lease events move a conversation's observed slot class across four turns — `NEW → RESOURCE_HOLDER` (mid-turn acquire), `RESOURCE_HOLDER` (next turn seeded from the ledger), `→ IN_FLIGHT` (mid-turn release), `IN_FLIGHT` (no seeding after release). Plus per-session isolation, ledger retirement on delete/clear, 32-goroutine same-session race coverage, and tolerant-parse tables on both contracts.

## Status (per docs honesty rules)

✅ Contract types, ledger, mark/unmark/seed, tests — this PR
📋 MCP adapter migration to the shuttle contract — lands with #344/#355
📋 Express-lane per-call handle checkout — needs the server half (teradata-mcp-v2)

[...]

PR #376, opened 2026-08-27
fix(scheduler): aging advances a starvation tier, not a priority class

ilsiepotamus:
## Found by trying to measure the thing it broke

I set out to prove whether RESOURCE_HOLDER priority inheritance (#365/#369) actually helps. 384 agents, ceiling pinned below demand so admission genuinely queues, waiter classes sampled every 30s:

```
t+5m   382 holder    1 in-flight
t+10m  368 holder
t+15m  252 holder
t+20m  245 holder    2 in-flight
```

Essentially every parked request read as a holder, after **715 promotions**. Class ordering had degenerated to FIFO, so the mechanism could not be exercised at all — and a naive A/B would have reported "priority inheritance has no effect" for entirely the wrong reason.

**Cause:** aging promoted waiters up the *same ladder* priority inheritance orders (`NEW → IN_FLIGHT → RESOURCE_HOLDER`). A request holding nothing got relabelled a lease holder purely for having waited. The longer a scope stayed saturated, the less the class meant — so the dilution is worst exactly when scheduling matters most.

## The fix: two concerns, two mechanisms

- **Class says what a request IS.** `RESOURCE_HOLDER` is reachable only by actually holding a lease. Aging never writes it.
- **Starvation says how long it has WAITED.** Aging advances a tier (`promoted++`); dispatch serves waiters that have crossed a threshold ahead of every non-starved request in the band, oldest first.

Liveness is preserved rather than traded away: a starved `NEW` waiter is still served before a freshly-arrived holder. `dispatchLocked` now builds each band's order in a single pass (starved oldest-first, then class order, FIFO within class) and rebuilds the queues from what remains, so a waiter cannot be granted twice.

## One subtlety worth reviewing

Precedence is **sticky** (`promoted > 0`), not re-derived from the escalating threshold. My first draft used one predicate for both jobs, and because the threshold doubles per tier, a just-promoted waiter lost its precedence the instant it earned it — the new liveness test caught that, which is the main reason I trust it.

Two test expectations changed because they encoded the old ladder (`TestAgingPromotesStarvedWaiters`, `TestAgingStaysWithinBand`); both now assert the new contract, and the band test keeps its real subject — aging must never cross bands — which this change makes trivially true.

## Docs corrected

The proto promised "any waiter reaches the top class in at most 2 × starvation_age_s", which is no longer how liveness is delivered; `starvation_age_s` and the `RESOURCE_HOLDER` comment now describe the actual mechanism. Also fixed a stale claim in `band_test.go` that batch liveness comes from "its own budget share" — there is no such share; interactive outranks batch entirely.

## Validation

`gofmt`/`go vet` clean, `buf lint`/`generate` clean (comment-only proto diff), `go build -tags fts5 ./...`, scheduler suite `-race -count=3`, plus `-race` on pkg/agent, pkg/server, pkg/llm, cmd/loom, cmd/looms.

### DIFF
--- a/gen/go/loom/v1/llm_scheduler.pb.go
+++ b/gen/go/loom/v1/llm_scheduler.pb.go
@@ -40,7 +40,9 @@ const (
 	SlotPriorityClass_SLOT_PRIORITY_CLASS_IN_FLIGHT SlotPriorityClass = 2
 	// Conversation holding an external scarce resource (database session
 	// handle, MCP slot): priority inheritance — it is blocking other agents
-	// that are burning LLM budget trying to acquire what it holds.
+	// that are burning LLM budget trying to acquire what it holds. Reached
+	// ONLY by actually holding a lease; starvation aging never promotes a
+	// request into this class (see starvation_age_s).
 	SlotPriorityClass_SLOT_PRIORITY_CLASS_RESOURCE_HOLDER SlotPriorityClass = 3
 )
 
@@ -179,9 +181,13 @@ type LLMSchedulerConfig struct {
 	// The door gate is wired process-wide and configured at boot via looms
 	// config (llm.max_door_queue); this per-scope field is not yet read.
 	MaxDoorQueue int32 `protobuf:"varint,4,opt,name=max_door_queue,json=maxDoorQueue,proto3" json:"max_door_queue,omitempty"`
-	// Seconds after which a waiting slot request is promoted one priority
-	// class. Bounds starvation: any waiter reaches the top class in at most
-	// 2 * starvation_age_s.
+	// Seconds a request may wait before it earns starvation precedence, which
+	// puts it at the head of its band ahead of every non-starved request
+	// regardless of class (oldest starved first). Each further interval
+	// advances its tier. Bounds starvation without rewriting priority class:
+	// waiting longer does not make a request a lease holder, so the
+	// RESOURCE_HOLDER class keeps meaning what it says even in a scope that
+	// has been saturated for a long time.
 	StarvationAgeS int32 `protobuf:"varint,5,opt,name=starvation_age_s,json=starvationAgeS,proto3" json:"starvation_age_s,omitempty"`
 	// Target utilization of measured capacity, (0, 1]. 0 = default 0.8.
 	UtilizationTarget float32 `protobuf:"fixed32,6,opt,name=utilization_target,json=utilizationTarget,proto3" json:"utilization_target,omitempty"`
--- a/gen/openapiv2/loom/v1/llm_scheduler.swagger.json
+++ b/gen/openapiv2/loom/v1/llm_scheduler.swagger.json
@@ -83,7 +83,7 @@
         "starvationAgeS": {
           "type": "integer",
           "format": "int32",
-          "description": "Seconds after which a waiting slot request is promoted one priority\nclass. Bounds starvation: any waiter reaches the top class in at most\n2 * starvation_age_s."
+          "description": "Seconds a request may wait before it earns starvation precedence, which\nputs it at the head of its band ahead of every non-starved request\nregardless of class (oldest starved first). Each further interval\nadvances its tier. Bounds starvation without rewriting priority class:\nwaiting longer does not make a request a lease holder, so the\nRESOURCE_HOLDER class keeps meaning what it says even in a scope that\nhas been saturated for a long time."
         },
         "utilizationTarget": {
           "type": "number",
@@ -137,7 +137,7 @@
         "SLOT_PRIORITY_CLASS_RESOURCE_HOLDER"
       ],
       "default": "SLOT_PRIORITY_CLASS_UNSPECIFIED",
-      "description": "SlotPriorityClass orders slot requests. Higher classes are served first;\naging promotes a starved waiter one class (liveness, not throughput).\n\n - SLOT_PRIORITY_CLASS_NEW: First LLM call of a conversation: admitted as churn allows.\n - SLOT_PRIORITY_CLASS_IN_FLIGHT: Conversation mid-task: run-to-completion bias — finishing releases every\nresource the conversation holds.\n - SLOT_PRIORITY_CLASS_RESOURCE_HOLDER: Conversation holding an external scarce resource (database session\nhandle, MCP slot): priority inheritance — it is blocking other agents\nthat are burning LLM budget trying to acquire what it holds."
+      "description": "SlotPriorityClass orders slot requests. Higher classes are served first;\naging promotes a starved waiter one class (liveness, not throughput).\n\n - SLOT_PRIORITY_CLASS_NEW: First LLM call of a conversation: admitted as churn allows.\n - SLOT_PRIORITY_CLASS_IN_FLIGHT: Conversation mid-task: run-to-completion bias — finishing releases every\nresource the conversation holds.\n - SLOT_PRIORITY_CLASS_RESOURCE_HOLDER: Conversation holding an external scarce resource (database session\nhandle, MCP slot): priority inheritance — it is blocking other agents\nthat are burning LLM budget trying to acquire what it holds. Reached\nONLY by actually holding a lease; starvation aging never promotes a\nrequest into this class (see starvation_age_s)."
     },
     "v1SlotState": {
       "type": "object",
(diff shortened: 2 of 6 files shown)


Josh-Schoen:
The problem here is real and the fix's framing is right: aging was conflating "has waited a long
time" with "holds a scarce resource," and under sustained saturation that conflation actively
erased the distinction priority inheritance depends on. Separating class (what a request IS) from
starvation tier (how long it's WAITED) is the correct fix, and the `TestStarvedWaiterOutranksHolderWithoutBecomingOne`
test is a good proof that liveness survives the separation.

**Verdict: approve once F1 is addressed.**

## What this makes possible

RESOURCE_HOLDER priority inheritance (#365/#369) can now actually be exercised and measured under
sustained load — which is the thing your own load test set out to check and found it couldn't.
Verified with a discriminating test: overlaying your exact test files onto the pre-fix
`scheduler.go` reproduces the bug precisely (`TestStarvedWaiterOutranksHolderWithoutBecomingOne`
fails with "the freshly-arrived holder was served before the starved waiter" against the old code,
passes against the new code), so this isn't just a plausible-sounding fix — it's demonstrated.

## Where it could hurt

`dispatchLocked`'s own doc comment says "Below a non-starved head, backfill continues
(throughput)" — and that's no longer true. The old per-class loop broke on an unfit head but moved
on to the *next class*, so a large blocked `RESOURCE_HOLDER` didn't stop smaller `NEW`/`IN_FLIGHT`
waiters from getting granted. The new merged `starved`+`ordinary` loop does one unconditional
`break` at the first ungrantable waiter regardless of whether it's starved, which stops *all*
grants in the band until that head either fits or starves. I reproduced this directly: a
non-starved 400-token holder that can't yet fit blocks a 100-token `NEW` waiter that could —
passes against your base commit, fails against this PR's head. Under the exact saturated-load
conditions your PR's own test methodology targets, this means one stuck large reservation can now
stall a burst of cheap requests that used to sail through, for up to a full `starvation_age_s`
before the (correct) starvation halt would have kicked in anyway. Nothing in the suite would catch
a regression here — `TestStarvedTopClassHeadHaltsBackfill` only exercises the *starved*-head-halts
case, never the non-starved case that's supposed to keep flowing.

## Verified credit

- `buf lint`, `buf format --diff --exit-code`, and `buf breaking` (against `main`) all pass; the
  proto diff is confirmed comment-only, non-breaking.
- `buf generate` reproduces the committed `gen/go` and `gen/openapiv2` files byte-for-byte — zero
  diff. Proto/gen parity is exactly right.
- `go build -tags fts5 ./...`, `go vet -tags fts5 ./...`, `gofmt -l` on the touched files, and
  `go test -tags fts5 -race -count=3 ./pkg/llm/scheduler/...` all pass, matching your validation
  notes. Also reran `-race` on `pkg/agent`, `pkg/server`, `pkg/llm`, `cmd/loom`, `cmd/looms` —
  all clean.
- The "sticky precedence" subtlety you called out (`hasStarvedLocked` checks `promoted > 0`, not a
  re-derivation against the escalating threshold) is correct as implemented and is exactly what
  keeps a just-promoted waiter from losing precedence the instant its tier advances.

## Findings

**F1 — `dispatchLocked` backfill/throughput regression for non-starved heads.**
`pkg/llm/scheduler/scheduler.go`, `dispatchLocked()`. Affects: operators under saturated load (the
same persona your own load test exercises). The merged `starved`+`ordinary` grant loop breaks
unconditionally on the first ungrantable waiter instead of continuing past a non-starved one to
try the next class, as the old code did and as the function's own comment still claims. Suggest
restructuring so the `ordinary` portion still breaks per-class (skip to the next class on an unfit,
non-starved head) rather than across the whole flattened list, and adding a regression test — a
`TestNonStarvedHeadAllowsBackfill` alongside the existing `TestStarvedTopClassHeadHaltsBackfill` —
so a future change can't silently reintroduce this.

**F2 — Enum-level proto comment now contradicts the field-level fix.**
`proto/loom/v1/llm_scheduler.proto:20-21` (propagates to `gen/go/loom/v1/llm_scheduler.pb.go:31`
and the swagger doc). The `SlotPriorityClass` summary still says "aging promotes a starved waiter
one class (liveness, not throughput)" directly above the corrected `RESOURCE_HOLDER` comment
saying the opposite. One-line fix plus regenerate. Non-blocking, but worth catching before someone
reads the generated godoc and takes the wrong sentence at face value.

**F3 — Orphaned doc-comment fragment.**
`pkg/llm/scheduler/scheduler.go:665-667`. The old `starvedAtTopLocked` doc comment survived as
unchanged diff context and now sits above `agingDueLocked`'s own doc comment, describing a
"top class of its band" concept that no longer exists. Cosmetic; delete the three lines.

## Things I was wrong about

I initially suspected the `Acquire()` fast path (`queuesEmptyAtOrAboveLocked`) could let a fresh
higher-class arrival slip past an already-parked, already-starved lower-class waiter — a subtler
version of the same liveness gap. Tracing every code path that can free budget (`release`,
`UpdateFromHeaders`, `ObserveSuccess`, `SetConfig`, the 1s ticker), all of them call
`dispatchLocked()` before releasing the lock, so a starved queued waiter is always granted
synchronously before a concurrent `Acquire()` could observe the freed capacity. That invariant is
pre-existing, untouched by this PR, and holds. Not a real finding — withdrawn.

## Environment notes for next time

- `buf generate` here uses remote BSR plugins (`buf.build/protocolbuffers/go`, etc., per
  `buf.gen.yaml`) rather than local `protoc-gen-*` binaries — none of the local plugin binaries
  were installed in this environment and it still worked over the network. Worth knowing if a CI
  runner ever loses BSR connectivity.
- Overlaying the PR's test files onto the base commit's `scheduler.go` (same package, same test
  helpers) is a cheap way to get a true regression-test proof without needing `git bisect` or a
  synthetic harness — recommend it for future scheduler PRs in this package.

Josh-Schoen (changes requested):
Requesting changes on F1 only — see the comment above for full detail. The starvation/priority-class fix itself is verified (overlaid the PR's 3 new tests onto pre-fix code: all fail with the described symptom, all pass on the PR's code; proto/gen parity confirmed via a real buf generate diff). But a probe test shows the rewritten dispatchLocked introduces a real throughput regression not mentioned in the PR description — any single unfit waiter now halts its whole priority band instead of letting smaller waiters behind it get granted. Confirmed by running the identical probe against base (passes) and PR head (fails). Recommend holding until that's addressed.

ilsiepotamus:
F1 fixed in `44b199a4`. You were right, and pinning it turned up something worse.

**The fix**: dispatch is now two passes instead of one flattened loop.

- Starved waiters, oldest first across classes: an unfit head **halts the band**, because backfilling past it is exactly how it never accumulates the capacity it needs.
- Then the ordinary class order, FIFO within a class: an unfit head stops **only its own class**. A non-starved waiter has no claim on the band, so a large `RESOURCE_HOLDER` that cannot yet fit no longer stalls a burst of cheap `NEW` waiters — the throughput `dispatchLocked`'s own comment promises.

**The worse thing**: `TestStarvedTopClassHeadHaltsBackfill` was passing for the wrong reason and never exercised starvation at all. Starvation is `promoted > 0`, and `promoted` is bumped only by `ageLocked` on the **1s tick** — so the test's 300ms sleep with `StarvationAge: 200ms` left nothing starved. It passed purely because the old flattened break halted the band for *any* unfit head, starved or not.

In other words the suite was pinning the regression you found, and this PR's headline behaviour was untested. The test now waits for a real promotion (`waitUntilStarved`) before asserting the halt, so it tests what its name claims.

Both directions verified by removing each behaviour in turn: dropping the starved halt fails `TestStarvedTopClassHeadHaltsBackfill`, restoring the flattened break fails the new `TestNonStarvedHeadAllowsBackfill`, and both pass together.

`go test -tags fts5 -race -count=3 ./pkg/llm/scheduler` passes; golangci-lint 0 issues. CI green (18 pass).

ilsiepotamus:
Rebased cleanly onto current \`main\` at \`5d3d2001\` (fix commit unchanged: \`44b199a4\`, the two-pass dispatch fix for the F1 throughput regression).

Verified on the pushed head: \`go build -tags fts5 ./...\`, \`go test -tags fts5 -race ./pkg/llm/scheduler/...\` (all pass), \`gofmt -l\` clean, \`golangci-lint run --build-tags fts5 ./...\` (0 issues), \`buf lint\`/\`buf format --diff\`/\`buf breaking --against main\` all clean, CI green.

PR #382, opened 2026-08-31
HITL park-and-resume: a turn ends at a human decision and resumes on it

mrg-ray:
## What

Park-and-resume for HITL: a tool batch that needs a human decision no longer blocks inside the turn — the turn ENDS cleanly and resumes when the decision arrives, however much later.

- **Pre-scan admission** (`maybeParkBatch`): before anything in a batch executes, every call is preflighted (`Chain.Preflight` — same verdict surface as execution admission, side-effect free). Any Ask or registered `contact_human` question parks the WHOLE batch: one grouped `human_requests` row (per-call descriptors keyed by ToolCall.ID), durable tail `…user, assistant(ToolCalls)` with no tool rows, typed `TurnParkedError` terminal. A Deny is never lifted and never parks.
- **`ResumeChat`**: continues the parked turn without appending a user message. Turn-aware tail walk binds the decision to its exact batch (`ErrNothingParked` / `ErrNotParkedTail` / `ErrStaleDecision` typed terminals); rejected → uniform `permission_denied` with the human's verbatim reason; approved → question items synthesize the human's answer, everything else dispatches under a batch-scoped `AskGrant` (lifts Ask only — Deny still dominates, re-verified per call at completion time via the pre-scan's own classification). Loop then re-enters normally; a nested park chains.
- **Append-point guard**: a new user turn on a session with a pending parked decision is refused with the typed `SessionParkedError` before any turn-end side effect — closes the race between an embedder's admission-time probe and a park landing mid-turn.
- **`dispatchOneCall` extraction**: the conversation loop's dispatch body extracted verbatim (shared by the loop and resume completion); golden/differential tests pin identical behavior with the feature off, for ungoverned agents, and for ask-free batches.
- `shuttle`: `Preflight` on Executor/Chain, `AskGrant` context (batch-scoped Ask lift), `SummarizeCall`/`BoundParams` exports.

Embedder contract: `WithHITLPark(store, ttl, notifier)` — nil store leaves the legacy in-turn resolver path fully intact.

## Testing

`go test -tags fts5 ./...` green (the sole failure, `pkg/orchestration TestPipeline_PartialResultsCostAndDuration`, is a pre-existing timing flake present on main). Park behavior suite covers: pre-scan ordering/durability, block-all, question answers, deny-never-lifted (pure and mixed batches, forged-answer refusal), resume terminals, tail walk across in-turn user rows, the append-point guard, params bounding, and the no-op proofs. Proven live end-to-end (with the loom-cloud embedder) under real LLM traffic: approve / reject-and-replan / question-answers / janitor-driven resume with no client. Mixed multi-item batches are covered by the behavior suite and verified manually through the product UI.

### DIFF
--- a/cmd/looms/cmd_hitl.go
+++ b/cmd/looms/cmd_hitl.go
@@ -511,6 +511,24 @@ func runHitlRespond(cmd *cobra.Command, args []string) {
 
 	span.SetAttribute("responded_by", respondedBy)
 
+	// A PARKED request cannot be answered by a store write. Every other HITL
+	// request has a waiter inside the turn polling for exactly this row, so
+	// writing the verdict is the whole handoff. A park has no waiter — its
+	// turn ENDED — and only Agent.ResumeChat can continue it. Deciding the row
+	// here would mark it answered while the approved actions never run, and
+	// the model would never learn what the human said.
+	if before, gerr := store.Get(ctx, requestID); gerr == nil && before != nil && before.RequestType == "parked" {
+		fmt.Fprintf(os.Stderr,
+			"Request %s is a PARKED turn: its decision must be applied through the agent's\n"+
+				"ResumeChat entry point, which re-enters the conversation. Responding here would\n"+
+				"close the request without ever running the actions or telling the model.\n\n"+
+				"Use the embedder that raised it (session %s) to deliver this decision.\n",
+			requestID, before.SessionID)
+		span.SetAttribute("success", false)
+		span.SetAttribute("refused", "parked")
+		os.Exit(1)
+	}
+
 	// Respond to request. The store's conditional write is a deliberate no-op
 	// for an already-decided or expired request, so success is judged by
 	// reading the row back — never by the write returning nil.
--- a/docs/architecture/hitl-park-and-resume.md
+++ b/docs/architecture/hitl-park-and-resume.md
@@ -176,6 +176,22 @@ must not kill a session.
   missing from the metrics backend.
 - 📋 Loop re-entry re-runs graph-memory context injection for the same turn,
   duplicating the injected block and paying a recall round-trip per resume.
-- 📋 `Chain.Preflight` re-evaluates every matching hook, so a stateful host
-  hook (a quota counter) sees each parked call twice.
 - 📋 No proto/gRPC surface for resume; park is a library API for embedders.
+  Whether loom should expose resume over gRPC is a product decision, not an
+  oversight — every other conversation entry point reaches clients through the
+  Weave surface, so the asymmetry is worth a deliberate answer.
+
+## 8. A note on hook evaluation count
+
+The pre-scan asks the admission chain the same question execution asks, so a
+call that parks and then runs evaluates every matching hook **twice** (three
+times for a `contact_human` re-classified at resume). This is safe because
+`Hook.Evaluate` is required to be side-effect free — now stated on the
+interface — with `PostToolHook` as the seam for anything that records. A host
+hook that counts inside `Evaluate` will over-count; that was true before park
+existed, but park is the first caller that makes it likely.
+
+Dynamic registration is not doubled: `tryDynamicRegistration` ends in
+`registry.Register`, so the pre-scan's registration is a hit for the later
+`Execute` rather than a second discovery round-trip. The cost moves earlier,
+it does not repeat.
--- a/docs/architecture/hitl-park-and-resume.md
+++ b/docs/architecture/hitl-park-and-resume.md
@@ -1,0 +1,181 @@
+# HITL Park-and-Resume: A Turn That Ends at a Human Decision
+
+**Status**: 🚧 In Development — PR #382 plus the review fixes on `fix/hitl-park-review`. The park pre-scan, `ResumeChat`, the request lifecycle, and the append-point guard are implemented with tests in `pkg/agent/hitl_park_test.go` and `pkg/agent/hitl_park_lifecycle_test.go`. Park is opt-in per agent (`WithHITLPark`) and is consumed by an embedder — loom itself ships no resume RPC.
+**Related**: PR #374 (in-turn hold heartbeat), PR #382 (park-and-resume)
+
+---
+
+## 1. Two ways to wait for a human
+
+Loom has two answers for a tool call that needs a person, and they are not
+alternatives so much as different scopes.
+
+| | In-turn hold (#374) | Park-and-resume (#382) |
+|---|---|---|
+| The turn | stays open, blocked inside the batch | **ends**, `TurnParkedError` |
+| The caller's stream | kept alive by a 30s heartbeat | closes; the resume is a new call |
+| Bounded by | the resolver's timeout | the request row's TTL (default 168h) |
+| Enabled by | an ask resolver on the chain | `WithHITLPark(store, ttl, notifier)` |
+| Good for | a decision measured in seconds | a decision measured in hours or days |
+
+A nil park store leaves the in-turn path exactly as it was. An agent that
+wires both keeps the hold as the fallback for any ask the pre-scan did not
+classify as a park item.
+
+### 1.1 What the progress stream shows
+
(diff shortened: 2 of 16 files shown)


ilsiepotamus:
Reviewed this end to end. The design holds up — the turn-aware tail walk, the verbatim `dispatchOneCall` extraction pinned by golden tests, and Deny never being lifted are all well judged, and `-race` is clean on both touched packages.

I found three defects I could reproduce, plus several smaller ones, and rather than just list them I've put the fixes up against this branch: **mrg-ray/loom#2** (from `teradata-labs/loom:fix/hitl-park-review`, branched off this PR's head `bb2034f7`). `just check` green.

The three with repros:

1. **The happy path wedges the session.** `ResumeChat` uses `decision.RequestID` as a span attribute and nothing else — it never loads or resolves the row, and nothing else in loom closes a `"parked"` row (the `ExpireRequest`/`RespondToRequest` call sites are the legacy in-turn waiters, which park mode doesn't wire). `guardParkedTail` then refuses every later turn: approve → resume → turn completes → *next message rejected, permanently*.

2. **Empty `ItemIDs` bypasses the batch binding.** `locateParkedBatch` only checks that each supplied id is in the tail batch, so supplying none checks nothing. Since a nested park chains, a spent decision for request A — redelivered without item ids — executes batch B.

3. **The grant reaches calls that were never on the card.** `completeParkedBatch` dispatches every rowless call under one blanket `AskGrant`, including calls that preflighted `Allow` and so never appeared in the request's params. If such a call's verdict has since become `Ask`, it's silently admitted — and ungranted it would have failed closed, so the grant is strictly weaker than the default there.

Also fixed in that PR: the `assistantPersisted` gate is vacuous (`PersistMessage` returns nil with no store, so it reports success for a batch written nowhere — which is also why the park suite runs with its central precondition switched off); the park's deferred `ReleaseAll` really releases MCP session handles of a turn that hasn't finished; an approval past `ExpiresAt` executes; a turn that already produced its final reply is still resumable via the `MaxToolExecutions` break's rowless calls; resumed turns record none of the six conversation metrics; graph-memory context is re-injected per resume; and `looms hitl respond` silently no-op-approves a parked row (a store write can't answer a park — there's no waiter).

One thing worth your call, not fixed: there's no proto/gRPC surface for resume. Loom's HITL decision channel has always been the store, which is sufficient for a hold because the waiter polls from inside the turn — but a park needs something to actively call `ResumeChat`, and there's no in-loom caller. Whether that belongs on the Weave surface seemed like your decision rather than a review fix.

Happy to rework any of it if you'd rather solve one differently.

mrg-ray:
Thank you — this is exactly the review the entry point needed, and the fix branch is merged (bb2034f7..2c16b5d8) with everything taken as-is except two pieces I had to reconcile with the embedder flow this PR ships for (loom-cloud), both now proven against its live E2E:

**1. Decided rows resume (`d689c50b`).** The pending-only gate in `loadParkedRequest` closed the wedge for the standalone flow, but the loom-cloud embedder records the verdict at its own respond door FIRST — under that door's expiry CAS — and then resumes; every one of its resumes arrived at an already-decided row and got `ErrUnknownRequest`. `loadParkedRequest` now accepts both states: pending = your flow exactly (ResumeChat decides, applies the apply-time expiry override, closes the row); decided = the row's status is authoritative over the caller's payload (an approving payload cannot execute against a rejected row), expiry is not re-judged at apply time for a decision recorded in time, and the row is left closed as the embedder left it. Replay stays unexecutable in both flows — the tail-walk terminals are the guard, which is where your reapply/nested-batch tests now land (`ErrNothingParked`/`ErrStaleDecision`, runs still pinned at 1/0).

**2. Handles stay call-scoped (`d689c50b`).** The collector hand-off is only reachable when one `Agent` instance serves both halves of the turn. In the pooled lifecycle loom-cloud runs — a fresh `Agent` per call — a collector parked by a nested re-park is unreachable forever and its handles leak server-side, which is strictly worse than the re-mint-on-demand behavior every live run has exercised. I've reverted to release-at-every-park-exit and recorded cross-call handle continuity as an embedder-owned seam in your §4, next to the resume-transport question. If you see a shape that serves both lifecycles without the leak, I'd genuinely like it — this was the least-harm call, not a disagreement with the defect.

On the design question: agreed it's deliberate, not an oversight — leaving resume as a library API for now. The one in-product caller drives it through the store + `ResumeChat`, and whether a Weave-surface resume should exist is being taken as a product decision, tracked in your §7.

Everything else — request-authoritative binding, card-scoped grant, `HasStore` gate (and the fixture growing a real store), preflight normalization order, the expiry refusal, metrics on resumed turns, graph re-injection, the CLI guard, the Evaluate purity contract, the doc — landed verbatim. The kind/question card headline is a real UX improvement on the product surface too.

mrg-ray:
Follow-up on point 2 of my earlier comment: the revert is gone — replaced with the shape that serves both lifecycles (`c4e0266a`).

Your adoption model is restored and completed: `chat()`'s FIRST park now parks its collector too (the piece the fix branch didn't reach — only re-parks did), so a same-process resume adopts the whole turn's handles and the finishing call releases them once. For the pooled lifecycle, `Agent.ReleaseParkedHandles(sessionID)` is the explicit drain — the loom-cloud executor calls it at its park terminal, keeping call-scoped semantics with no unreachable collector. §4 of your doc rewritten accordingly; lifecycle tests pin first-park parking, nested re-park continuity, and the drain's idempotence.

Also two hardenings on the shared invariants since, both probe-driven (`4425e444`): an applied batch whose turn died before its final reply resumes to completion again (the unconditional `rowless==0 → ErrNothingParked` stranded that turn's answer — the case your `MaxToolExecutions` reasoning correctly refuses is kept, gated on `finalReply`), and item binding is now by content (`verifyItemBinding`: recorded tool + seq must still describe the tail call), since per-response counter call IDs can collide across nested batches and ID-only binding let a redelivered decision reach an unseen batch.

ilsiepotamus (approved):

ilsiepotamus:
Correcting my earlier comment on this PR, now that it's merged.

**I overstated one fix.** I said the handle-release bug was fixed by mrg-ray/loom#2. It wasn't, for the common case: `parkHandles` had a single caller — `ResumeChat`'s *nested* park — while every FIRST park exits through `chat()`, whose unconditional deferred `ReleaseAll` still ran. The mechanism was dead for the case it was written for, and my commit message claimed the general case. @mrg-ray fixed this properly in c4e0266a ("handle continuity across the park gap"), independently and more thoroughly than I had. Same for binding: 4425e444 binds by recorded tool and batch position, not just call ID, which closes a gap I had only flagged.

**And four defects survived the merge** — three of which I've now reproduced against `main`. I put them up as #383 rather than leave them implied here:

1. **A human-approved batch executes twice.** `loadParkedRequest`'s doc names the tail walk as the double-application guard, but that walk reads the rowless calls *before* they execute, so two overlapping resumes both act. Both return `nil`, so nothing detects it — and two tool rows land for one `tool_use_id`, which puts duplicate `tool_result` blocks in the next provider request.
2. **A TTL lapsing during the batch wedges the session.** `closeParkedRequest` ran after the batch and judged success by the write's error, but `RespondToRequest` returns `nil` when it refuses past expiry. Row stays `pending`; `guardParkedTail` refuses every later turn.
3. **An abandoned park wedges the session permanently** — `guardParkedTail` has no `ExpiresAt` check and nothing sweeps parked rows.
4. **Adopt-then-refuse destroys a live park's handles** — adoption ran before validation, so a resume refused with a stale id released the handles of a still-parked turn.

Three of those trace to my first round, not to this PR's original design — I introduced the close that silently fails and the adoption ordering, and I missed the concurrency window entirely while reviewing. #382's own approach was execute-then-nothing; my half-fix made it execute-then-silently-fail-to-close.

Worth recording for anyone reading this thread later: none of the four is caught by CI. #382 merged fully green, and my first round also passed `just check` twice while carrying them. They only surfaced under an independent re-review plus mutation testing — deliberately breaking each mechanism to see whether any test noticed. Several tests here, mine included, passed regardless.

Nothing in loom is affected today, since park is opt-in with no in-tree caller.

PR #383, opened 2026-09-01
fix(hitl): claim a parked decision, and bind it to what the human approved

ilsiepotamus:
Follow-up to #382. Ten defects in the merged park-and-resume path, found across three independent review passes. Several were reproduced against `main` before fixing; every mechanism is mutation-verified (removing it fails the suite). `just check` green.

Park is opt-in (`WithHITLPark`) with no in-tree caller, so nothing in loom is broken today — but an embedder turning it on gets all of them.

## Approval bypass

**A spent decision executed a different, unapproved batch.** A decided row still loads — the embedder-recorded flow needs that — so redelivery was refused only by `verifyItemBinding`, which compared tool name and batch position and never the arguments, though `buildParkParams` already records them. Those two facts can carry *zero* information: on Gemini `ToolCall.ID` **is** the function name (`pkg/llm/gemini/client.go` — "Gemini doesn't provide call IDs"), so for a single-call batch the id, tool and seq collapse into one fact.

```
turn 1 approves export_csv{v:"customers_sample"} → runs
turn 2 parks   export_csv{v:"ALL_CUSTOMERS_PII"} → undecided, card unseen
turn 1's decision redelivered → err=nil, the PII export ran
```

The binding now compares the recorded params against the tail call's.

**Unbindable call IDs no longer park.** A duplicate ID collapsed two calls into one card descriptor — the human approves one action, two run — and an empty ID could never be matched back, wedging the session with no library path out. Ollama can omit the id; Gemini parallel calls reuse the function name. Such a batch dispatches inline, where an Ask with no resolver fails closed.

## Double execution and wedged sessions

**A human-approved batch executed twice.** The tail walk was the only guard, and it reads the batch's rowless calls *before* they execute, so two overlapping resumes both acted — both returning `nil`, and landing two tool rows for one `tool_use_id`, which puts duplicate `tool_result` blocks in the next provider request (`compileLocked` emits every contiguous tool row with no dedupe). The decision is now **claimed before anything executes**: written, read back, with a token in `response_data` saying which resume won — because "no longer pending" does not say *who* closed it. That also makes a parked batch at-most-once.

**A TTL lapsing during the batch wedged the session.** The close ran after the batch and was judged by the write's error, but both store writes are documented no-ops returning `nil` when the row is no longer claimable.

**An abandoned park wedged the session permanently** — `guardParkedTail` had no `ExpiresAt` check and nothing sweeps parked rows.

**The append-point guard was blind in the embedder-recorded flow.** It matched on `Status == "pending"`, but that flow decides the row *before* calling `ResumeChat` — so in between, a new user turn was admitted and buried the unapplied batch, silently discarding the approval and leaving the history asserting the call never completed. The guard now also asks the tail: an assistant batch still missing tool rows is an unapplied park whatever the row says.

**A question park answered `responded` was a dead end** — the documented status for an input/decision request, rejected as "not a park verdict" while the row stopped holding the session. Accepted for question cards, where responding *is* the verdict; still refused on an approval card.

## Destructive and lifecycle

**A refusal that proved nothing erased the approval.** `abandonParkedRequest` fired on any `locateParkedBatch` error, but `ErrNothingParked` is absence of evidence — and an empty session yields it with the park intact, because `GetOrCreateSessionWithAgent` fails **open** (a load error builds a fresh empty session under the same ID) and `LoadSession` is user-scoped. A transient store error, or a resume whose ctx carried a different identity, expired the row and destroyed the human's decision record. Since `loadParkedRequest` authorizes only on `hr.SessionID`, a caller who knew a session id and request id but could not read the session could close someone else's approval. Now closes only on `ErrNotParkedTail` — the one signal that is positive evidence.

**Handle adoption ran before validation**, so a refused resume released the handles of a still-parked turn. Moved below the claim. **An abandoned turn leaked its collector**, which pins every later turn on that session to `RESOURCE_HOLDER` via the lease ledger.

**A cross-process resume lost its graph-memory context.** The injection is in-memory only and never persisted; suppressing it on "is a resume" left exactly the turn carrying the approved action with no recall.

**The CLI parked-row guard failed open** on a store error — the silent wrong outcome it was added to prevent. **`looms hitl expire`** now warns that expiring a park releases the session without finishing the turn. **Synthesized refusals** carry `admission_decision: deny`; a NULL is miscounted by the analytics arm keying on `IS NULL`. **The session was created before the request was validated**, so a bogus RequestID persisted a session row.

## Scope, stated honestly

The per-session lock is **per-Agent**, not per-process — `sessionLocks` lives on the `Agent`, and a pooled embedder builds a fresh one per call. The standalone flow does not depend on it (it claims the row, which holds across agents and processes); the embedder-recorded flow has no row left to claim, so an embedder that pools agents or spreads resumes across processes must deliver an already-decided resume at most once itself. That obligation is in the code and the doc.

Also: the noop differential gains a session store on its park-enabled arm — without one the pre-scan gate is off, so it compared park-off against park-off, and a `panic` in `maybeParkBatch` left it green.

## Left open, documented

`docs/architecture/hitl-park-and-resume.md` §8 names what is still missing rather than leaving it for the next reader: abandoned parks are never swept; restart between park and resume is untested although it is the feature's premise; the standalone flow's durable row is not the authorization record (`Approved` is a parameter, the approver is the literal `"human"`); `AskGrant` is a blanket context value a nested-execution tool would inherit; and there is no proto/gRPC surface for resume.

Two test-quality gaps on main I did not close: `TestPark_HandleContinuityAcrossTheGap` asserts map bookkeeping over a collector holding zero handles (nothing in the fixture mints one), and `TestPreflightAdmitParity` never configures a resolver, so `Preflight`'s non-blocking contract is unpinned. Both need fixtures this PR does not build.

### DIFF
--- a/cmd/looms/cmd_hitl.go
+++ b/cmd/looms/cmd_hitl.go
@@ -517,7 +517,21 @@ func runHitlRespond(cmd *cobra.Command, args []string) {
 	// turn ENDED — and only Agent.ResumeChat can continue it. Deciding the row
 	// here would mark it answered while the approved actions never run, and
 	// the model would never learn what the human said.
-	if before, gerr := store.Get(ctx, requestID); gerr == nil && before != nil && before.RequestType == "parked" {
+	// Fail CLOSED: the sqlite and in-memory stores return an error for both
+	// absence and store failure, so a Get hiccup must not skip this guard and
+	// decide a parked row anyway — that is the exact silent wrong outcome the
+	// guard exists to stop.
+	before, gerr := store.Get(ctx, requestID)
+	if gerr != nil || before == nil {
+		fmt.Fprintf(os.Stderr,
+			"Cannot read request %s to check whether it is a parked turn (%v).\n"+
+				"Refusing rather than risk deciding a park the agent must resume.\n",
+			requestID, gerr)
+		span.SetAttribute("success", false)
+		span.SetAttribute("refused", "unverifiable")
+		os.Exit(1)
+	}
+	if before.RequestType == "parked" {
 		fmt.Fprintf(os.Stderr,
 			"Request %s is a PARKED turn: its decision must be applied through the agent's\n"+
 				"ResumeChat entry point, which re-enters the conversation. Responding here would\n"+
--- a/cmd/looms/cmd_hitl.go
+++ b/cmd/looms/cmd_hitl.go
@@ -423,6 +423,21 @@ func runHitlExpire(cmd *cobra.Command, args []string) {
 	if operator == "" {
 		operator = "operator"
 	}
+
+	// Expiring a PARKED row is legitimate — it is the retirement route for a
+	// stranded park — but it does not finish the turn. The batch keeps its
+	// unpaired tool calls, and the model is later shown "the call did not
+	// complete" for actions a human may well have approved. Say so, rather
+	// than printing an unqualified success.
+	if before, gerr := store.Get(ctx, requestID); gerr == nil && before != nil && before.RequestType == "parked" {
+		fmt.Fprintf(os.Stderr,
+			"Note: %s is a PARKED turn. Expiring it releases the session, but does NOT\n"+
+				"complete the turn — its tool calls stay unanswered and the agent will report\n"+
+				"them as never completed. Prefer resuming through the embedder that raised\n"+
+				"it (session %s) when that is still possible.\n\n",
+			requestID, before.SessionID)
+	}
+
 	if err := store.ExpireRequest(ctx, requestID, "operator:"+operator); err != nil {
 		span.RecordError(err)
 		span.SetAttribute("success", false)
--- a/docs/architecture/hitl-park-and-resume.md
+++ b/docs/architecture/hitl-park-and-resume.md
@@ -124,12 +124,10 @@ The grant lives only on the derived context handed to the parked batch's
 dispatch. The loop re-entry that follows runs ungranted, so a new ask in the
 continuation parks again.
 
-### 3.3 Closing the row
+### 3.3 Claiming the decision
 
-In the standalone (pending-row) flow, the row is closed the instant its
-decision is applied, **before** the loop re-entry that may park a new one — a
-pre-decided row is already closed by the embedder's respond door and is left
-untouched:
+In the standalone (pending-row) flow the row is closed **before the batch
+runs** — it is a claim, not bookkeeping:
 
 - approved → `RespondToRequest(…, "approved", …)`
 - rejected → `RespondToRequest(…, "rejected", …)`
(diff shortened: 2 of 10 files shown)


bsahitya:
# Review: PR #383 — `fix(hitl): claim a parked decision before executing it`

**Author:** ilsiepotamus (+ Claude) · **Base:** main · `+1197 / -83`, 9 files, 6 commits

> **Verdict: Approve.** This is high-quality, correct concurrency work. It builds clean (`-tags fts5`), `go vet` is clean, `gofmt` is clean, and the new/changed tests pass under `-race`. Comments below are mostly nits plus a few things worth confirming before merge.

---

## What it does

Fixes four+ defects in the opt-in HITL "park & resume" feature that survived #382. The central change: on the standalone (pending-row) flow, a resume now **claims the row before executing the batch** instead of closing it as bookkeeping afterward.

---

## Verification performed

- Read the full diff + surrounding source in a throwaway worktree of the PR head.
- Confirmed the store CAS the design rests on: `RespondToRequest` is an atomic `UPDATE … WHERE id=? AND status='pending' AND expires_at > ?` under mutex (`human_store_sqlite.go:453`), and returns `nil` for both "landed" and "no-op refused" — so judging by error alone genuinely is unsafe, and the read-back + token is the right fix.
- Confirmed `buildParkParams` (`park.go:621`) always sets `desc["params"]` (bounded map or digest string), so the new `sameParkedParams` default-false branch cannot spuriously refuse every legitimate resume. Good — this was my main correctness worry and it's covered.
- Ran `TestPark*`, `TestClaimParkedRequest*`, `ParkNoop*` with `-race`: all green.

---

## Strengths

- **Correct serialization model.** Row-claim (cross-process) for the standalone flow + per-agent session lock for the pre-decided flow, with the scope limits stated honestly rather than overclaimed. `claimParkedRequest` (`park.go:300`) reads back and compares a token — exactly right given the store's silent-no-op contract.

- **Fail-closed discipline applied consistently.** The CLI guard (`cmd_hitl.go`) now refuses when it can't read the row instead of deciding a parked row anyway; the claim refuses when it can't verify. This matches the file's own stated principle.

- **The self-corrections across commits 3–5 are the real value.**
  - *Commit 3* catches that abandon-on-any-error was destructive (an empty session from a fail-open `GetOrCreateSessionWithAgent` yields `ErrNothingParked` with the park intact); the fix narrows abandonment to `ErrNotParkedTail` only — positive evidence the turn is dead.
  - *Commit 4* catches a genuine approval-bypass (Gemini ID == tool name, so id/tool/seq collapse) and binds on arguments.
  - *Commit 5* catches that suppressing graph-memory injection on every resume starved the highest-risk turn of recall in a fresh process.
  
  These are subtle and each is mutation-tested.

- **Tests are honest about their own limits** — e.g. `TestPark_ConcurrentResumesExecuteTheBatchOnce` explicitly says it's a smoke test and points to the unit test that actually contends the claim. That's rare and welcome.

---

## Things worth confirming (not blockers)

**1. Expired standalone path** (`park.go:316`) returns `nil` regardless of who won the close. Within one agent the session lock serializes, so no duplicate refusal rows. Across pooled agents, two concurrent expired resumes can each synthesize refusal tool rows → duplicate `tool_result` blocks for one `tool_use_id` — the exact failure mode the PR elsewhere fixes. It's documented as the embedder's at-most-once obligation, consistent with the pre-decided flow, so acceptable — but confirm you're comfortable that the expired path shares the pre-decided flow's weaker guarantee rather than the claim's strong one.

**2. `sameParkedParams` relies on JSON-marshal equality** (`park.go:266`). `json.Marshal` sorts map keys so ordering is fine, and `float64(5)`/`int(5)` both render `5`. Edge cases (large int precision after a sqlite JSON round-trip, `nil` vs empty-map from `BoundParams`) are unlikely to bite but aren't tested. A residual: if a later batch has identical args to a spent approval, it re-binds — arguably fine (same authorized action), just noting it's the boundary.

**3. Known gaps are documented, not fixed** (§7): abandoned parks still leak their row + handle collector; no restart/rehydration test coverage though cross-process survival is the feature's whole premise. Since the feature is opt-in with no in-tree caller, shipping with these documented is reasonable — but "restart is untested although it is the feature's premise" is a notable gap to track.

---

## Nits

- **`types.go:88`:** The doc comment is garbled — the `parkedHandles` prose and the new `sessionLocks` prose run together on one line (`…closing a dead turn's row. Pooled embedders — …`). Reads like a bad merge; please reflow. The `sessionLocks` sentence should sit directly above its fields.

- **`ErrClaimNotConfirmed`** is declared in a separate `var` block from the other terminals purely to attach a "retryable" comment — fine, but a caller who treats all `park.go` errors as terminal (the previous contract) will now silently drop a retryable case. Make sure the CLI/embedder callers actually branch on it, or the retryability is theoretical.

- Commit messages and PR body are excellent but very long; the prose style (heavy em-dashes, editorializing) leaks into some code comments (e.g. the `agentCtx` *"which is not a property the next person moving code through this function should have to preserve"*). Not wrong, just verbose for inline comments.

---

## Bottom line

Genuinely careful work with a strong test-to-claim ratio and honest scoping. I'd merge after:

- [ ] Fix the `types.go` comment reflow
- [ ] Quick confirmation that callers distinguish `ErrClaimNotConfirmed`

The expired-path and identical-args residuals are worth a tracking issue but don't block.

bsahitya (dismissed):

bsahitya (approved):
# Re-review: PR #383 — HITL park-and-resume defect batch

**Author:** ilsiepotamus (+ Claude) · **Base:** main · `+1540 / -93`, 10 files, 8 commits

**Context:** This is a re-review. Prior review (approved, then dismissed) covered through commit `7a9191d`. One new commit landed after it: `ec1c3a3` (18:05, ~30 min after the prior review at 17:09).

> **Verdict: Approve, unchanged.** The new commit is a real fix, correctly scoped, and mutation-verified. Build is clean (`-tags fts5`), `go vet` clean, `gofmt` clean, and the park/claim/concurrency/noop suites pass under `-race`. The two merge conditions from the prior review are still open — details below.

---

## What changed since last review (`ec1c3a3`)

The unapplied-tail guard from `37deae9` held the session for any non-pending parked row whose batch was still rowless. `looms hitl expire` writes `"timeout"` and leaves the batch rowless by definition — so the guard silently neutered the operator's only recovery route for a stranded park.

The fix narrows the decided-row branch of `guardParkedTail` to hold only on a verdict awaiting application (`approved`/`rejected`/`responded`), skipping `timeout`.

**Verifications:**

- **Status set is correct and exhaustive.** The store's only statuses are `pending`, `approved`, `rejected`, `timeout`, `responded` (`pkg/shuttle/human_tool.go:66`). The whitelist covers exactly the three verdicts; `timeout` is the sole closure status and is correctly excluded. No missing status (e.g. no separate `"expired"`/`"cancelled"` string that would slip through).

- **No invalid provider request from the release.** After expire, the parked batch keeps dangling `tool_use` blocks with no `tool_result`. Confirmed this is safe — `compileLocked` emits `syntheticFailedResult` for every call left without a result row (`pkg/agent/context_compilation.go:44`, §5.2 step 7). The model is told `"[no result recorded — the call did not complete...]"`, which matches the commit's stated behavior. No dangling-pair API break.

- **Mutation-verified.** Removing the whitelist (`if r.Status != "approved" && ...`) fails `TestPark_OperatorRetirementReleasesTheSession` with exactly `"session still held after the operator retired the park."` The test drives real `Chat → park → ExpireRequest → Chat` and asserts release.

This is the same class of careful, mutation-tested work as the rest of the PR.

---

## Follow-up if needed (not blocking merge)

The two items from the prior review checklist didn't make it into this PR — that's fine given none are correctness blockers. Please track them in a follow-up issue alongside the carried residuals if needed:

1. **`types.go` comment reflow** — the `parkedHandles` prose still runs on (`...a dead turn's row. Pooled embedders —`) with `sessionLocks` wedged inside it (`types.go:83–95`). Trivial reflow.
2. **`ErrClaimNotConfirmed` caller branching** — currently referenced only inside `park.go`. Wire one caller to branch on it and pin with a test, or drop the retryable/terminal split so it's not dead API surface.
3. **Expired standalone path weak guarantee** — two concurrent expired resumes across pooled agents can each synthesize refusal rows → duplicate `tool_result` for one `tool_use_id`. Documented as the embedder's at-most-once obligation; worth tracking explicitly.
4. **`sameParkedParams` JSON-marshal edge cases** (`park.go:266`) — large-int precision after a sqlite JSON round-trip and `nil` vs empty-map from `BoundParams` are untested.
5. **Restart/rehydration coverage** — cross-process resume was verified by hand but has no suite. Cross-process survival is the feature's whole premise; §8 documents the gap honestly.

---

## Bottom line

The delta since the last pass closes a real operator-recovery hole the earlier guard opened, and it's well-verified. Approval stands.

PR #384, opened 2026-09-02
fix(agent): stop the output-token breaker firing on turns that made progress

ilsiepotamus:
## Why

The output-token circuit breaker exists to stop an agent that is stuck emitting tool calls it can never finish. Two accounting defects made it fire on sessions that were **not** stuck, ending the whole message with a hard error.

This was found while investigating a Tera 2.0 demo environment where field users reported "tools loop, then it errors out" — including during a live customer demo.

## 1. A tool that takes no arguments read as truncated

`detectEmptyToolCall` treated an empty `Input` as evidence the model was cut off mid-generation:

```go
if len(tc.Input) == 0 {
    return true
}
```

Plenty of tools legitimately take no arguments — a list or status call — and are *always* invoked with `{}`. `list_skills` is one. On a turn that also hit `max_tokens`, every such call counted as a truncated turn.

It now consults the advertised tool set: a tool whose schema declares no required parameter is expected to be called with an empty input, so emptiness says nothing about truncation for it. A tool the advertised set cannot describe (unknown name, or no schema) keeps the previous heuristics, and passing a nil set is exactly the old behavior — so the conservative direction is preserved everywhere the schema does not clearly say otherwise.

## 2. A productive turn did not end the run of truncated ones

The accounting switch had a branch for `max_tokens` with **complete** tool calls that neither counted nor cleared:

```go
default:
    // max_tokens with non-truncated tool calls: agent may still make progress
    // on the next turn. Don't count, don't clear — let it continue.
```

Those calls are complete and executable, so the turn *is* forward progress. Leaving the counter untouched made `outputTokenExhaustions` a lifetime tally instead of the consecutive count the threshold is documented against. A session under sustained output pressure never got a clearing turn, so truncated turns scattered among productive ones still summed to the threshold — and the error then reported them as "consecutive", which they were not. It is also why the symptom looks like *many* tool calls before the failure rather than exactly `threshold` of them.

This is the same defect class as the verbose-text-response regression this switch was introduced to fix (`TestOutputTokenCB_SessionAccumulation_Regression`: *"the old code ... never cleared it between user turns when stop_reason was max_tokens"*), in the one branch that fix did not cover. It now clears, matching the zero-tool-call branch beside it.

## On the testing gap

Both defects lived in `conversationLoop`, where no tracker-level test could observe them. `TestOutputTokenCircuitBreaker_Integration` re-derives the branch decisions in the test body rather than driving the loop — and its own `else → clear` already encoded the corrected behavior for case 2, so the test and the code disagreed with nothing to notice.

The new tests drive the real loop through a scripted LLM that can express a stop reason (the shared `mockLLMResponse` cannot, which is likely why this path was never covered end to end).

**Each new test was confirmed to fail with its fix reverted:**

- `TestOutputTokenCB_CompleteToolCallEndsTheRun` → `circuit breaker fired on interleaved truncated/complete turns: ... triggered after 3 consecutive turns`
- `TestOutputTokenCB_NoArgumentToolIsNotTruncation` → `circuit breaker fired on a legitimately argument-less tool`
- `TestDetectEmptyToolCall_SchemaAware` → `detectEmptyToolCall() = true, want false`

`TestOutputTokenCB_ConsecutiveTruncatedStillFires` guards the other direction: the breaker still fires on genuinely consecutive truncated turns, so neither fix disarms it.

## Scope / compatibility

- `detectEmptyToolCall` is unexported; the signature change is internal to `pkg/agent`.
- The all-falsy-values heuristic is deliberately left alone — this PR only decides *whether that judgment applies to a given tool*, so every existing `TestDetectEmptyToolCall` expectation is unchanged (it now passes `nil` for the tool set).
- No config, API, or behavior change for an agent that is genuinely stuck.

## Verification

- `GOWORK=off go test -tags fts5 -race -short ./pkg/agent/...` — pass
- `gofmt -s -l`, `go vet ./pkg/agent/...`, `golangci-lint run ./pkg/agent/...` — clean (0 issues)

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -2530,9 +2530,13 @@ func (a *Agent) runConversationLoop(ctx Context) (*Response, error) {
 			}
 
 			if llmResp.StopReason == "max_tokens" {
-				// tools is this provider call's advertised set, so a tool that
-				// takes no arguments is not mistaken for a truncated call.
-				hasEmptyToolCall := detectEmptyToolCall(llmResp.ToolCalls, tools)
+				// tools is this provider call's advertised set, so a call can be
+				// checked against the schema it was advertised with: a tool that
+				// takes no arguments is not mistaken for a truncated call, and a
+				// call missing what its schema demands is not mistaken for a
+				// complete one.
+				toolState := classifyToolCalls(llmResp.ToolCalls, tools)
+				hasEmptyToolCall := toolState == toolCallStateIncomplete
 
 				switch {
 				case threshold < 0:
@@ -2579,12 +2583,12 @@ func (a *Agent) runConversationLoop(ctx Context) (*Response, error) {
 						"output_tokens": llmResp.Usage.OutputTokens,
 					})
 
-				default:
-					// max_tokens with non-truncated tool calls. These calls are
-					// complete and executable, so the turn IS forward progress
-					// and the run of consecutive truncated turns ends here —
-					// the same reasoning the zero-tool-call branch above
-					// applies to a complete text response.
+				case toolState == toolCallStateComplete:
+					// max_tokens, and every call carries everything its schema
+					// demands. These calls are executable, so the turn IS
+					// forward progress and the run of consecutive truncated
+					// turns ends here — the same reasoning the zero-tool-call
+					// branch above applies to a complete text response.
 					//
 					// This previously neither counted nor cleared, which made
 					// outputTokenExhaustions a lifetime tally instead of the
@@ -2597,9 +2601,30 @@ func (a *Agent) runConversationLoop(ctx Context) (*Response, error) {
 					// introduced to fix (see
 					// TestOutputTokenCB_SessionAccumulation_Regression), in the
 					// one branch that fix did not cover.
+					//
+					// Clearing is gated on COMPLETE rather than on "not
+					// visibly empty": a malformed call reaches this switch with
+					// a populated-looking input (the OpenAI and Azure clients
+					// store an unparseable arguments string as {"_raw": ...}),
+					// and treating that as progress would let an alternating
+					// stream of broken calls reset the counter forever and
+					// disarm the breaker on exactly the providers whose
+					// truncation is most visible.
 					failureTracker.clearOutputTokenExhaustion()
 
-					span.AddEvent("output_token.non_truncated_toolcall", map[string]interface{}{
+					span.AddEvent("output_token.complete_toolcall_cleared", map[string]interface{}{
+						"stop_reason":     llmResp.StopReason,
+						"tool_call_count": len(llmResp.ToolCalls),
+					})
+
+				default:
+					// toolCallStateUnknown: max_tokens with calls whose
+					// completeness cannot be established (an unadvertised tool,
+					// or one advertising no schema). Don't count — there is no
+					// evidence of truncation — but don't clear either, because
+					// clearing asserts progress we cannot demonstrate. This is
+					// the pre-existing conservative behavior of this branch.
+					span.AddEvent("output_token.indeterminate_toolcall", map[string]interface{}{
 						"stop_reason":        llmResp.StopReason,
 						"tool_call_count":    len(llmResp.ToolCalls),
 						"has_empty_toolcall": hasEmptyToolCall,
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -2530,7 +2530,9 @@ func (a *Agent) runConversationLoop(ctx Context) (*Response, error) {
 			}
 
 			if llmResp.StopReason == "max_tokens" {
-				hasEmptyToolCall := detectEmptyToolCall(llmResp.ToolCalls)
+				// tools is this provider call's advertised set, so a tool that
+				// takes no arguments is not mistaken for a truncated call.
+				hasEmptyToolCall := detectEmptyToolCall(llmResp.ToolCalls, tools)
 
 				switch {
 				case threshold < 0:
@@ -2578,8 +2580,25 @@ func (a *Agent) runConversationLoop(ctx Context) (*Response, error) {
 					})
 
 				default:
-					// max_tokens with non-truncated tool calls: agent may still make progress
-					// on the next turn. Don't count, don't clear — let it continue.
+					// max_tokens with non-truncated tool calls. These calls are
+					// complete and executable, so the turn IS forward progress
+					// and the run of consecutive truncated turns ends here —
+					// the same reasoning the zero-tool-call branch above
+					// applies to a complete text response.
+					//
+					// This previously neither counted nor cleared, which made
+					// outputTokenExhaustions a lifetime tally instead of the
+					// consecutive count the threshold is documented against: a
+					// session under sustained output pressure never got a
+					// clearing turn, so truncated turns scattered among
+					// productive ones still summed to the threshold and failed
+					// the whole message. That is the same defect class as the
+					// verbose-text-response regression this switch was
+					// introduced to fix (see
+					// TestOutputTokenCB_SessionAccumulation_Regression), in the
+					// one branch that fix did not cover.
+					failureTracker.clearOutputTokenExhaustion()
+
 					span.AddEvent("output_token.non_truncated_toolcall", map[string]interface{}{
 						"stop_reason":        llmResp.StopReason,
 						"tool_call_count":    len(llmResp.ToolCalls),
--- a/pkg/agent/circuit_breaker_test.go
+++ b/pkg/agent/circuit_breaker_test.go
@@ -252,7 +252,7 @@ func TestDetectEmptyToolCall(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := detectEmptyToolCall(tt.toolCalls)
+			result := detectEmptyToolCall(tt.toolCalls, nil)
 			assert.Equal(t, tt.expected, result)
 		})
 	}
--- a/pkg/agent/circuit_breaker_test.go
+++ b/pkg/agent/circuit_breaker_test.go
@@ -252,7 +252,7 @@ func TestDetectEmptyToolCall(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := detectEmptyToolCall(tt.toolCalls, nil)
+			result := classifyToolCalls(tt.toolCalls, nil) == toolCallStateIncomplete
 			assert.Equal(t, tt.expected, result)
 		})
 	}
--- a/pkg/agent/conversation_helpers.go
+++ b/pkg/agent/conversation_helpers.go
@@ -18,6 +18,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"time"
+
+	"github.com/teradata-labs/loom/pkg/shuttle"
 )
 
 // failureSignature uniquely identifies a tool call failure for tracking consecutive failures.
@@ -244,8 +246,23 @@ func (t *consecutiveFailureTracker) checkOutputTokenCircuitBreaker(threshold int
 // detectEmptyToolCall checks if any tool call has an empty Input object.
 // This is a strong indicator that the LLM output was truncated mid-generation.
 // Returns true if empty tool call detected.
-func detectEmptyToolCall(toolCalls []ToolCall) bool {
+//
+// tools is the advertised tool set for the current provider call, used to tell a
+// genuinely argument-less call apart from a truncated one: a tool that requires
+// no arguments is EXPECTED to be invoked with an empty input, so emptiness says
+// nothing about truncation for it. Pass nil to skip that check and apply the
+// emptiness heuristics to every call (the historical behavior).
+func detectEmptyToolCall(toolCalls []ToolCall, tools []shuttle.Tool) bool {
 	for _, tc := range toolCalls {
+		// A no-argument tool call is complete as it stands. Without this, every
+		// invocation of a tool like list_skills or a bare status/list call read
+		// as truncated, and on a turn that also hit max_tokens it counted
+		// toward the output-token circuit breaker — 8 of them failed the whole
+		// message even though nothing was ever truncated.
+		if !toolRequiresArguments(tc.Name, tools) {
+			continue
+		}
+
 		// Check if Input is empty or only has zero values
 		if len(tc.Input) == 0 {
 			return true
@@ -266,6 +283,28 @@ func detectEmptyToolCall(toolCalls []ToolCall) bool {
 	}
 
 	return false
+}
+
+// toolRequiresArguments reports whether the named tool's advertised schema
+// declares at least one required parameter.
+//
+// It answers true for a tool it cannot describe — an unknown name, or one with
+// no schema — so the emptiness heuristics in detectEmptyToolCall still apply
+// unchanged there. That is the conservative direction: a call we cannot
+// characterize keeps the pre-existing truncation detection rather than being
+// waved through.
+func toolRequiresArguments(name string, tools []shuttle.Tool) bool {
+	for _, t := range tools {
+		if t == nil || t.Name() != name {
+			continue
+		}
+		schema := t.InputSchema()
+		if schema == nil {
+			return true
+		}
+		return len(schema.Required) > 0
+	}
+	return true
 }
 
 // TokenBudgetConfig holds configuration for token budget management.
--- a/pkg/agent/conversation_helpers.go
+++ b/pkg/agent/conversation_helpers.go
@@ -243,50 +243,8 @@ func (t *consecutiveFailureTracker) checkOutputTokenCircuitBreaker(threshold int
 	return fmt.Errorf("%s", msg)
 }
 
-// detectEmptyToolCall checks if any tool call has an empty Input object.
-// This is a strong indicator that the LLM output was truncated mid-generation.
-// Returns true if empty tool call detected.
-//
-// tools is the advertised tool set for the current provider call, used to tell a
-// genuinely argument-less call apart from a truncated one: a tool that requires
-// no arguments is EXPECTED to be invoked with an empty input, so emptiness says
-// nothing about truncation for it. Pass nil to skip that check and apply the
-// emptiness heuristics to every call (the historical behavior).
-func detectEmptyToolCall(toolCalls []ToolCall, tools []shuttle.Tool) bool {
-	for _, tc := range toolCalls {
-		// A no-argument tool call is complete as it stands. Without this, every
-		// invocation of a tool like list_skills or a bare status/list call read
-		// as truncated, and on a turn that also hit max_tokens it counted
-		// toward the output-token circuit breaker — 8 of them failed the whole
-		// message even though nothing was ever truncated.
-		if !toolRequiresArguments(tc.Name, tools) {
-			continue
-		}
-
-		// Check if Input is empty or only has zero values
-		if len(tc.Input) == 0 {
-			return true
-		}
-
-		// Check if all values are empty/zero
-		allEmpty := true
-		for _, v := range tc.Input {
-			if v != nil && v != "" && v != 0 && v != false {
-				allEmpty = false
-				break
-			}
-		}
-
-		if allEmpty {
-			return true
-		}
-	}
-
-	return false
-}
-
 // toolRequiresArguments reports whether the named tool's advertised schema
-// declares at least one required parameter.
+// demands any property, at the root or through a composite branch.
 //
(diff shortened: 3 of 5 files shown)


Ed-Jaramillo (changes requested):
## Summary

The no-argument-tool correction is sound for the covered cases, and the new tests exercise the real conversation loop. However, the new reset path treats "not empty/all-falsy" as proof that a `max_tokens` tool call is complete and executable. Repository provider adapters show a concrete path where malformed or required-key-incomplete calls reset a genuinely consecutive truncation run, defeating the circuit breaker's containment behavior.

## Blocking finding: malformed tool calls can clear the truncation counter

At `pkg/agent/agent.go:2582-2600`, `hasEmptyToolCall == false` clears the counter. But `detectEmptyToolCall` at `pkg/agent/conversation_helpers.go:255-308` only checks map length and zero-like values; it does not verify required keys, parse success, or schema validity.

OpenAI converts invalid argument JSON to `Input: {"_raw": <partial>}` at `pkg/llm/openai/client.go:498-512` and `pkg/llm/openai/client.go:774-784`; Azure OpenAI does the same at `pkg/llm/azureopenai/client.go:395-410` and `pkg/llm/azureopenai/client.go:890-900`. Those clients can independently map `finish_reason: "length"` to `StopReason: "max_tokens"`, so this malformed non-empty input reaches the reset branch.

For a tool requiring `sql`, alternating these incomplete responses demonstrates the issue:

1. `max_tokens` with `{}` increments the counter.
2. `max_tokens` with `{"_raw":"{\"sql\":"}` clears it.
3. Repeating the sequence prevents the threshold from being reached.

Please clear the counter only after every returned call is positively established as complete. At minimum, account for missing required keys and provider argument-parse failure. Root `Required` alone also does not capture `oneOf`/`anyOf` requirements allowed by the MCP schema.

Please add an end-to-end regression that alternates empty and non-empty malformed or missing-required-key input for a required-argument tool and asserts the breaker still fires on the third consecutive incomplete turn. A root-`oneOf` case where each branch requires a different property should also verify that `{}` remains incomplete.

Reviewed at `a5782e89867a1fc78a06a120c3c6c6f3710927e1`.

ilsiepotamus:
Thanks — this was a real defect and you're right that round 1 made it worse than the branch it replaced. Pushed `4bf6b2f5`.

I verified the `_raw` claim before acting on it, and it holds on all four sites: `pkg/llm/openai/client.go:505` (non-streaming) and `:781` (streaming accumulator), `pkg/llm/azureopenai/client.go:402` and `:896`, with `"length" → "max_tokens"` mapped independently at `azureopenai/client.go:379` and `:923`. So a truncated call on those providers arrives as a populated-looking map, and gating the reset on "not visibly empty" would have let alternating broken calls reset the counter forever — disarming the breaker on exactly the providers whose truncation is most detectable, which is worse than the old branch that at least held the count.

## What changed

The emptiness predicate is replaced by a three-state classification (`classifyToolCalls`/`classifyToolCall`):

- **incomplete** — a `_raw` parse-failure marker, a missing demanded property, or the legacy empty/all-falsy input
- **complete** — every call carries everything its schema demands
- **unknown** — completeness cannot be established (unadvertised tool, or one advertising no schema)

Clearing now requires **complete**. An **unknown** turn neither counts nor clears — that is this branch's pre-existing conservative behavior, and clearing there would assert progress that hasn't been demonstrated.

Counting also now catches malformed and missing-required calls. Worth flagging as a deliberate behavior change beyond the reset path: on OpenAI/Azure, truncation produced `_raw` and so was *never* counted before, meaning the breaker could not fire for those providers at all. It can now.

On composite schemas — you were right that root `Required` is insufficient. `schemaRequirementsMet` handles `oneOf`/`anyOf` (one satisfied branch) and `allOf` (all of them), and `schemaDemandsProperty` walks the same branches so a tool whose branches each demand a different property is no longer classified as argument-less. Recursion is depth-bounded (`maxSchemaDepth`) since these schemas come from MCP servers and other third parties. One judgment call: `oneOf` is checked as "at least one" rather than JSON Schema's "exactly one" — the question here is only whether the model finished emitting arguments, and the permissive direction keeps a well-formed call from reading as truncated. Say the word if you'd rather it were strict.

I also removed `detectEmptyToolCall` rather than leaving it as a wrapper, so there's one source of truth; its table test now drives the classifier and every existing expectation still holds unchanged.

## Tests

Both requested cases are in, plus the third state:

- `TestOutputTokenCB_MalformedArgsDoNotClear` — alternating `{}` / `{"_raw": "{\"sql\":"}`, breaker still fires
- `TestOutputTokenCB_MissingRequiredKeyDoesNotClear` — alternating `{}` / `{"limit": 5}` (required `sql` absent)
- `TestOutputTokenCB_IndeterminateToolCallDoesNotClear` — an unknown call must not clear a run
- `TestClassifyToolCall_CompositeSchemas` — includes the root-`oneOf` case where each branch demands a different property, asserting `{}` stays incomplete
- `TestClassifyToolCall_RawMarkerAlwaysIncomplete` — including on a tool that demands nothing, where an empty input would otherwise be valid
- `TestClassifyToolCalls_Aggregation` — incomplete > unknown > complete

Each guard was confirmed load-bearing by reverting it and watching the paired test fail. One honest note: with only the required-key check removed, `MalformedArgsDoNotClear` still passed — the `_raw` guard catches that case independently. The two overlap for a required-argument tool, and `RawMarkerAlwaysIncomplete` is what isolates the marker.

`GOWORK=off go test -tags fts5 -race -short ./pkg/agent/...` passes; gofmt/vet/golangci-lint clean.

Ed-Jaramillo (changes requested):
## PR Re-review — loom #384

**https://github.com/teradata-labs/loom/pull/384** · fix(agent): stop the output-token breaker firing on turns that made progress · by Ilsun Park (`@ilsiepotamus`) · `fix/output-token-cb-accounting` → `main` · re-reviewed at `4bf6b2f57ef46aaca5c196732cd19fe79b3c21a7` · 2026-09-02

### Summary

The remediation commit `4bf6b2f5` replaces the previous negative "not empty" reset condition with a three-state tool-call classifier: incomplete, complete, or unknown. It explicitly handles OpenAI/Azure `_raw` parse-failure fallbacks, missing root required properties, root `oneOf`/`anyOf`/`allOf` branches, unknown or nil-schema tools, and aggregation of multiple calls.

The original malformed-root-input issue is fixed for the newly covered cases, and exact-head CI is green. The fix is not complete, however: the classifier does not traverse property or array-item subschemas. A call that omits required nested data can therefore be classified complete and clear the circuit-breaker run even though a real repository tool rejects the same input. The new exact-key checks can also classify an executable snake_case input as incomplete before the executor performs its supported snake_case-to-camelCase normalization.

Live GitHub state remained stable throughout this re-review. GitHub reports the PR as mergeable, but its review decision remains `CHANGES_REQUESTED`.

### Change Since the Previous Review

- **Previously reviewed head:** `a5782e89867a1fc78a06a120c3c6c6f3710927e1`
- **Current and final head:** `4bf6b2f57ef46aaca5c196732cd19fe79b3c21a7`
- **New commit:** `4bf6b2f57ef46aaca5c196732cd19fe79b3c21a7` — `fix(agent): clear only on a positively complete turn (review feedback)`
- **Incremental change:** 452 additions and 59 deletions across the same four PR files
- **Current cumulative diff:** 769 additions and 21 deletions

The author acknowledged the prior defect and identified `4bf6b2f5` as the remediation. No additional submitted review exists at the current head.

### Prior Blocker Disposition

**PARTIALLY RESOLVED**

Resolved:

- OpenAI and Azure OpenAI `_raw` malformed-JSON fallback maps are now incomplete.
- Populated calls missing a root-level required key are now incomplete.
- Root `oneOf`, `anyOf`, and `allOf` branches containing required keys are assessed.
- Unknown and nil-schema populated calls no longer clear an existing run.
- Incomplete dominates unknown and complete when aggregating multiple calls.
- Exact-head CI passed the new malformed-input, missing-required, composite, aggregation, consecutive-truncation, complete-call, and no-argument regressions.

Unresolved:

- The schema checker does not descend through `Properties` or `Items`, so nested missing-required input can still clear the run.
- Classification occurs before the executor's established parameter-key normalization, causing divergence between "complete" accounting and actual executability.

### Evidence Reviewed

- Live PR metadata, description, commits, cumulative and incremental diffs, reviews, author response, linked-issue state, merge state, check runs, workflow logs, and artifact metadata.
- Every changed file in full at the current head:
  - `pkg/agent/agent.go`
  - `pkg/agent/conversation_helpers.go`
  - `pkg/agent/circuit_breaker_test.go`
  - `pkg/agent/output_token_cb_accounting_test.go`
- Circuit-breaker tracker semantics, dispatch order, recovery, tool execution, schema types, parameter normalization, and provider response conversion.
- Real nested tool schemas and runtime validation paths, including `generate_visualization`.
- OpenAI, Azure OpenAI, Anthropic, Bedrock, Gemini, LiteLLM, Mistral, and Hugging Face response conversion.
- Exact-head CI:
  - 18 checks succeeded, including race-enabled unit tests, fuzz tests, lint, three platform builds, CodeQL, MCP SDK interop, proto checks, coverage, and version consistency.
  - One `gosec` check was neutral because two configurations were not found. The separate Security Scan passed but invokes `gosec` with `-no-fail`.
  - No check was failed or pending.

### Verification of Material Claims

| Claim | Classification | Evidence |
|---|---|---|
| Concrete zero-argument tools are no longer treated as truncated. | **VERIFIED** | Current implementation and exact-head conversation-loop regression. |
| Exact-schema complete calls clear a previous exhaustion run. | **VERIFIED** | Current accounting branch and exact-head regression. |
| OpenAI/Azure malformed JSON can be represented as `_raw` alongside `max_tokens`. | **VERIFIED** | Provider conversion and stop-reason mappings. |
| Root missing-required calls no longer clear the run. | **VERIFIED** | Current classifier and alternating conversation-loop regression. |
| Root composite cases covered by the tests are handled. | **VERIFIED** | Current walker and exact-head `oneOf`/`anyOf`/`allOf` tests. |
| Composite and nested schema requirements are handled generally. | **CONTRADICTED** | The walker follows current-node composites but never descends into property or item subschemas. |
| Every complete and executable call is positively identified. | **CONTRADICTED** | Nested invalid input can be marked complete; normalized executable input can be marked incomplete. |
| Unknown or nil-schema populated calls do not clear a prior run. | **VERIFIED** | Three-state classifier and exact-head loop regression. |
| Multiple calls aggregate as incomplete > unknown > complete. | **VERIFIED** | Aggregation implementation and exact-head test. |
| There is no behavior change for an agent genuinely stuck on truncated calls. | **CONTRADICTED** | Nested missing-required responses can still reset the run indefinitely. |
| The author's local mutation/revert checks passed. | **AUTHOR-REPORTED** | No mutation-run artifact was available; exact-head CI validates only the current implementation. |
| Equivalent exact-head race tests and lint passed. | **VERIFIED** | Current-head workflow logs. |

### Merge Recommendation

**REQUEST CHANGES**

The prior root-level malformed-input problem has been materially improved, but a real nested repository schema still reproduces the same counter-reset failure one level deeper. The normalization mismatch also risks firing the breaker on calls the executor supports and successfully executes. The PR is not ready to merge until accounting uses the same structural and normalization semantics as execution and includes the corresponding conversation-loop regressions.

### Blocking Findings

#### 1. Nested required properties can still clear a consecutive incomplete run

- **Severity:** BLOCKER
- **Location:** `pkg/agent/conversation_helpers.go:307-336`, `schemaRequirementsMet`; reset consumer at `pkg/agent/agent.go:2586-2613`
- **Issue:** `schemaRequirementsMet` checks `Required` and composites only on its current schema node. It does not descend through `schema.Properties` or `schema.Items`, so missing nested required data is positively classified as complete.
- **Evidence:** The requirement walker contains no traversal of property or item subschemas. The repository's `generate_visualization` schema has a root-required `datasets` array whose item objects require `name` and `data` (`pkg/visualization/tool.go:52-74`). Its runtime rejects items missing those fields (`pkg/visualization/tool.go:139-173`). The new composite tests cover root branches only.
- **Concrete trigger:** For `generate_visualization`, alternate `max_tokens` responses containing `{}` and `{"datasets":[{}],"title":"t","summary":"s","output_path":"x"}`. The first increments the run. The second has every root key and is classified complete, clearing the run, although the real tool rejects the dataset as `INVALID_DATASET`. Repetition prevents the threshold from being reached.
- **Consequence:** A genuinely consecutive malformed/incomplete loop can continue to evade the output-token circuit breaker.
- **Minimal fix:** Recursively inspect present object properties and array items, including composites under those subschemas. If nested completeness cannot be established, classify the call as unknown rather than complete.
- **Required regression:** Add a threshold-three conversation-loop test alternating `{}` with the nested `generate_visualization` input above. Assert the breaker fires and the nested partial turn never clears the run. Include a nested `oneOf` or `allOf` under a required property.

### Non-Blocking Findings

#### Major — Classification does not use the executor's supported key normalization

- **Location:** `pkg/agent/conversation_helpers.go:307-314`; `pkg/shuttle/executor.go:305-308,607-639`
- **Issue:** Required-property checks use raw provider key spelling, while `Executor.Execute` deliberately normalizes snake_case and camelCase before admission and execution.
- **Evidence:** `schemaRequirementsMet` requires an exact raw key. `normalizeParametersToSchema` exists because models commonly send snake_case for camelCase schemas. `TestExecutor_ParameterNormalization_SnakeToCamel` defines required `userId`, supplies `user_id`, and proves the tool receives normalized `userId`; that test passed at the current head. Circuit-breaker accounting occurs before dispatch and normalization.
- **Concrete trigger:** A tool requires `userId`, and repeated `max_tokens` responses contain `{"user_id":"123"}`. The classifier counts each response as incomplete. The executor would normalize and execute the call, but at the threshold the breaker returns before dispatch.
- **Consequence:** Productive, executable calls can reproduce the false-positive failure that this PR is intended to fix.
- **Minimal fix:** Classify the same normalized parameter map used by `Executor.Execute`; extract the normalization into shared code so accounting and execution cannot diverge.
- **Required regression:** Register a camelCase-schema tool requiring `userId`; provide more than the threshold of `max_tokens` calls containing `user_id`; assert no circuit-breaker error and that each call reaches the tool with normalized `userId`.

#### Minor — `_raw` conflicts with a legitimate tool property

Presence of `_raw` is treated as an out-of-band parse-error signal before consulting the advertised schema (`pkg/agent/conversation_helpers.go:385-392`). JSON Schema and the repository's `Tool` contract do not reserve that property name, so a tool legitimately requiring `_raw` cannot be distinguished from the provider fallback. Carry provider parse failure as explicit `ToolCall` metadata rather than encoding it in the user argument map.

#### Minor — Schema-depth exhaustion fails open as complete

Beyond `maxSchemaDepth`, `schemaDemandsProperty` returns false while `schemaRequirementsMet` returns true (`pkg/agent/conversation_helpers.go:275-309`). An unassessable deeply composed schema can therefore clear the run. Propagate a tri-state traversal result and map depth exhaustion to unknown.

### Suggested PR Comments (local draft — not posted)

**Nested schema blocker**

`schemaRequirementsMet` walks `Required` and composites only on the current schema node; it never descends through `Properties` or `Items`. A real `generate_visualization` input such as `{"datasets":[{}],"title":"t","summary":"s","output_path":"x"}` is therefore classified complete and clears the counter, even though each dataset requires `name` and `data` and the tool rejects it. Alternating that response with `{}` still disarms the breaker. Please recurse through nested property/item requirements and add this alternating conversation-loop regression.

**Normalization mismatch**

The completeness check uses raw provider keys, but `Executor.Execute` intentionally normalizes snake_case/camelCase before admission and execution. A schema requiring `userId` accepts and executes `user_id` under the existing executor contract, but this classifier counts it as incomplete; repeated `max_tokens` turns can hard-fail before dispatch. Please classify the same normalized input the executor uses and add an end-to-end regression.

**`_raw` ambiguity**

`_raw` is inferred solely from the argument map, so a tool that legitimately defines an `_raw` property is indistinguishable from the provider fallback. Please carry parse failure as explicit `ToolCall` metadata rather than reserving a user-schema key.

**Depth handling**

The traversal depth bound currently fails open: beyond the limit, the code treats requirements as absent or satisfied. Since completeness cannot be established there, this should produce unknown rather than complete.

### Positive Observations

- The author correctly understood and fixed the original `_raw` and root-required cases.
- The three-state aggregation is materially safer than treating "not visibly empty" as proof of completeness.
- Multiple-call aggregation and unknown-call hold behavior are explicit and correct.
- The new tests exercise the real conversation loop and passed under race detection at the exact head.
- The current head remained unchanged throughout the re-review.

### Questions and Assumptions

- No author clarification is necessary to reproduce the nested-schema or normalization findings; both follow from the current implementation and existing repository contracts.
- This review treats omitted nested `Required` fields as incomplete because the input schema is the tool contract and real tool code rejects the demonstrated shape. Semantic constraints such as patterns, ranges, and enum values were not treated as evidence of output truncation.
- Required all-falsy values remain classified incomplete. That behavior predates this PR and is explicitly unchanged, so it does not affect this verdict.

### Review Limitations

- No local test, build, lint, or mutation run was performed. Exact-head CI logs and check output were inspected instead.
- CI artifact contents were not downloaded; artifact metadata and targeted logs were reviewed.
- The author's mutation/revert claims remain author-reported.
- The original production incident was not independently reproduced from a trace or incident ticket.
- The neutral `gosec` result limits independent security-scan certainty but does not affect the deterministic correctness/reliability findings.

**Confidence:** High in the blocker, normalization finding, exact-head metadata, and CI state. The reviewed head remained `4bf6b2f57ef46aaca5c196732cd19fe79b3c21a7`.

ilsiepotamus:
Both findings reproduce. Pushed `d7bc6fe5`.

## Blocker — nested requirements

Confirmed against `pkg/visualization/tool.go:52-75`: root requires `datasets`, and the item schema requires `name` and `data`. The walker only looked at its current node, so `{"datasets":[{}],"title":"t","summary":"s","output_path":"x"}` was positively reported **complete** and cleared the run — the same containment hole as round 2, one level deeper.

`checkAgainstSchema` now descends through present properties and array items, carrying the three-state verdict, with composites evaluated at each level. An absent optional property carries no requirement; an absent required one is caught at its own level.

One deliberate boundary: only **structural presence** is checked. Enum/pattern/range violations are not evidence the model stopped emitting mid-call, and treating one as truncation would fire the breaker on a complete answer. That matches the assumption in your review.

## Major — normalization mismatch

Also confirmed: `Executor.Execute` calls `normalizeParametersToSchema` at `executor.go:264,308,387`, and `TestExecutor_ParameterNormalization_SnakeToCamel` proves `user_id` reaches a `userId`-requiring tool normalized. Judging raw spelling counted those executable calls incomplete and would hard-fail a productive session at the threshold — precisely the false positive this PR exists to remove.

Taken your suggested fix rather than duplicating the logic: `normalizeParametersToSchema` is now exported as `shuttle.NormalizeParametersToSchema`, the three executor call sites use it, and classification runs on the same map `Execute` will run. Root keys only, matching what the executor actually does — nested keys stay untouched on both sides, so the two can't disagree there either. I noted that constraint in the exported doc comment.

## Minor — `_raw` ambiguity

Agreed the name isn't reserved. It's now read as a parse-failure marker only when the advertised schema does not itself declare a `_raw` property, so a tool that legitimately takes one keeps its own meaning.

I did **not** move parse failure onto `ToolCall` metadata. That's the cleaner design and I think you're right about it, but it reaches into every provider client and their conversion paths, which is a wider change than this PR should carry. Happy to do it as a follow-up if you'd prefer it done properly rather than worked around.

## Minor — depth failing open

Fixed. Past `maxSchemaDepth` both walkers previously landed on complete; exhaustion now yields **unknown**, which neither counts nor clears.

## Tests

Each guard confirmed load-bearing by reverting it and watching the paired test fail:

- `TestOutputTokenCB_NestedMissingRequiredDoesNotClear` — alternating `{}` / the nested partial above, threshold 3, breaker fires
- `TestClassifyToolCall_NestedRequirements` — nested items, one bad item among good ones, empty array (vacuously fine), and a nested `oneOf` under a required property
- `TestOutputTokenCB_SnakeCaseInputIsNotIncomplete` — five `max_tokens` turns of `{"user_id":"123"}` against a required `userId`: no breaker, and the tool asserts it received `userId`
- `TestClassifyToolCall_RawIsAProperty`
- `TestCheckAgainstSchema_DepthExhaustionIsUnknown`

`toolRequiresArguments` is folded into the classifier so there's one path; its table test now drives `classifyToolCall` with an empty input and states the same rules.

`GOWORK=off go test -tags fts5 -race -short ./pkg/agent/... ./pkg/shuttle/...` passes; gofmt/vet/golangci-lint clean.

On your note that the mutation/revert checks are author-reported — that's fair, they're local and leave no CI artifact. The reverts are mechanical if you want to reproduce any of them: drop the `schema.Properties` descent loop in `checkObject`, or swap `shuttle.NormalizeParametersToSchema(tool, tc.Input)` back to `tc.Input`, and the named test goes red.

PR #284, opened 2026-08-01
feat(runtime): OTLP tracing, multi-arch Docker image, env placeholder expansion for MCP/LiteLLM

Vasu2803k:
## Description

Adds OTLP tracing, multi-arch runtime packaging, and safe environment placeholder support for Loom runtime deployments. It also hardens LiteLLM/OpenAI, MCP streamable HTTP sessions, SSE compatibility, tool alias governance, and ephemeral-agent opt-in behavior.

Closes [TER-583](https://teradata-pe.atlassian.net/browse/TER-583) | [TER-597](https://teradata-pe.atlassian.net/browse/TER-597)

### Breaking change

`manage_ephemeral_agents` is now opt-in. Deployments that spawn sub-agents must add it to `tools.builtin`; programmatic `MultiAgentServer` embedders without an agent registry can call `SetAgentSpawnEnabled(agentID, true)`.

### Included

- Standard and Loom OTLP endpoint/header precedence shared by serve, workflow, and auto-select paths.
- Single-pass `${VAR}` expansion that preserves literal dollars, supports `$$`, and does not recursively expand resolved secrets.
- LiteLLM proxy health via `GET /health/liveliness`, avoiding paid chat probes.
- OpenAI malformed tool-pair filtering and one shared retry path for typed pre-response transport failures.
- MCP session ID replay, expiry/reinitialization, teardown, and wrapped default transport safety.
- Canonical tool identity for governance/circuit breakers and alias cleanup on unregister.
- Legacy SSE snake_case/numeric JSON compatibility plus synchronized heartbeat writes.
- Multi-arch runtime build targets deriving their default tag/version from `VERSION`.

### Validation

- [x] `just fmt` / `just fmt-check`
- [x] focused `go test -tags fts5 -race` across all changed packages
- [x] `just build`
- [x] `just test`
- [ ] `just build-runtime-multiarch` (not run locally)
- [ ] `just lint` (repository-wide pre-existing QF1012 staticcheck baseline remains; no findings in changed files)

### Merge note

Please squash-merge this PR because its history contains review/WIP commits.

PR #385, opened 2026-09-02
chore(legal): ship NOTICE and THIRD-PARTY-NOTICES.md with every release

ilsiepotamus:
Closes a real compliance gap found while compiling an OSS inventory for legal review.

**Loom distributes no NOTICE file, while several of its dependencies ship one.** Apache-2.0 §4(d) requires that if a work you redistribute contains a NOTICE file, its attribution notices are reproduced in your own distribution. Six modules in the shipped tree have one — `aws-sdk-go-v2`, `smithy-go`, `grpc`, and three yaml modules — and nothing carried them.

Separately, every permissive license in the tree (MIT, BSD-2, BSD-3, ISC, Apache-2.0, MPL-2.0) requires its text to accompany the distribution. We shipped only our own `LICENSE`.

## What this adds

- `NOTICE` — Loom's copyright plus every bundled work's NOTICE, verbatim.
- `THIRD-PARTY-NOTICES.md` — the full license text of all 139 distributed components, deduplicated by identical text.
- `cmd/gen-notices` — generates both from the dependency tree.
- `just notices` / `just notices-check`, the latter wired into `just check`.
- Release archives now include `LICENSE`, `NOTICE`, and `THIRD-PARTY-NOTICES.md` beside the binary.
- `docs/reference/third-party-notices.md`.

## Scope

The generator lists what is **compiled into the binaries we distribute** — `loom`, `looms`, `loom-standalone` — with each binary's release build tags (`fts5`, plus `standalone`).

That is narrower than the module graph on purpose. `go list -m all` resolves 422 modules; 133 reach a user's machine. The rest are test-only or build-time. Listing them would overstate the obligations and make legal review harder rather than more thorough.

## Three deliberate properties

**License text is copied verbatim** from the dependency's own source. Nothing classifies or normalizes into an SPDX identifier — an identifier is a summary, and a summary is not what these licenses require us to ship.

**A package carrying its own LICENSE is attributed separately from its module.** Some modules vendor subcomponents under a different license; rolling those up to the module root would attribute them wrongly. This is why 139 components map to 133 modules.

**A component whose license cannot be found is a hard error.** An early draft of the generator skipped quietly — and silently dropped three modules that name their license `LICENSE-APACHE-2.0.txt`. That is precisely the failure these files exist to prevent, so filename matching is now by prefix and a miss refuses to write anything.

## Verification

Reconciled independently against Google's `go-licenses` over the same package set: **139 components both ways, none unaccounted for.** License distribution is MIT 63, Apache-2.0 43, BSD-3-Clause 29, MPL-2.0 2, BSD-2-Clause 1, ISC 1 — **no GPL, AGPL, or LGPL anywhere in the distributed tree.**

`notices-check` verified in both directions: passes on a clean tree, fails on a tampered one. `just check` green with it integrated.

## Two things worth a reviewer's eye

**MPL-2.0 × 2** — `go-sql-driver/mysql` and `r3labs/sse/v2`. File-level copyleft; we link both unmodified and there is no `vendor/` directory, so no source-disclosure obligation attaches to our code. Their license text ships.

**Statically linked C that module tooling misses.** Built with `CGO_ENABLED=1`, and `mutecomm/go-sqlcipher` is MIT as a Go module while its LICENSE is a composite covering three bundled C projects compiled into our binaries: SQLite (MIT), **SQLCipher (BSD-3, © ZETETIC LLC)**, and **LibTomCrypt (public domain)**. Module-level scanners report only the MIT wrapper, so these are called out explicitly.

## Not a legal opinion

These files are a factual inventory generated from the dependency tree. They cannot detect a dependency that has misdeclared its license upstream. The docs say so, and say to treat a release as the unit of accuracy — `notices-check` in `just check` is what keeps that true between releases.


Josh-Schoen:
## PR #385 — chore(legal): ship NOTICE and THIRD-PARTY-NOTICES.md with every release

Shipping binaries without a NOTICE file while several bundled dependencies (aws-sdk-go-v2, smithy-go, grpc, three yaml modules) carry one is a real Apache-2.0 §4(d) gap, and this closes it — for `loom`, `looms`, `loom-standalone` specifically, scoped correctly to what's actually compiled into those binaries rather than the full 422-module `go list -m all` graph.

**Verdict: approve.** No blockers — one non-blocking coverage gap noted below.

**What this makes possible:** anyone redistributing a `loom`/`looms`/`loom-standalone` release archive now ships `LICENSE`, `NOTICE`, and `THIRD-PARTY-NOTICES.md` alongside the binary (confirmed in the `release.yml` diff — both the Unix `tar` and Windows `7z` packaging steps were updated). An operator or downstream packager no longer has to assemble this themselves before redistributing.

**Where it could hurt:** two things worth knowing, both minor.
1. `just notices-check` is real and wired into `just check` (verified: running `go run ./cmd/gen-notices -check` against this branch exits 0, matching the committed `NOTICE`/`THIRD-PARTY-NOTICES.md`) — but it is **not** wired into any GitHub Actions workflow. `.github/workflows/ci.yml` was not touched by this PR and none of the workflow files reference `notices`. That means a future dependency bump that adds or drops a NOTICE-bearing module will only be caught if the author remembers to run `just check` (or `just notices-check`) locally — CI itself won't flag drift.
2. `cmd/gen-notices` ships with no test file. `-check` acts as its own regression guard for *drift*, but a bug in the generator's own dedup/formatting logic (not drift, just wrong output) would pass `-check` by comparing generator output to itself.

**Verified credit:** ran `go run ./cmd/gen-notices -check` directly against the PR branch — exits 0, confirming the committed notices genuinely match the dependency tree as of this branch. Confirmed `just check`'s recipe list includes the new `notices-check` step. Confirmed the release workflow's both packaging steps (Unix and Windows) include the three new files. Doc-honesty grep against CLAUDE.md's banned-word list on `docs/reference/third-party-notices.md` and `NOTICE` — clean.

**Findings:**
- F1 (next contributor, minor): `notices-check` isn't in any CI workflow, only `just check` — a dependency-bump PR that skips `just check` locally will merge with stale notices undetected until someone runs it later. Fix size: small — add a `notices-check` step (or fold `just notices-check` in) to `.github/workflows/ci.yml`'s existing lint or a new job.
- F2 (next contributor, minor): no unit test for `cmd/gen-notices`'s dedup/generation logic itself. Fix size: small — a table test over a synthetic module list would cover the dedup-by-identical-text claim in the PR body.

**Things I was wrong about:** I initially suspected from the PR title/size that "wired into just check" might be an overstated claim (large diffs sometimes overstate integration) — checked the actual `Justfile` diff and it's true, verified by running the check.

**Environment gotchas:** none — this build has no proto/fts5/hawk interaction; `go run ./cmd/gen-notices` runs standalone.

Josh-Schoen (approved):
Approve. Ran go run ./cmd/gen-notices -check directly against the PR branch — exits clean, confirming the committed NOTICE/THIRD-PARTY-NOTICES.md genuinely match the dependency tree. Confirmed the Justfile wiring and both release.yml packaging steps. One non-blocking gap noted in the comment above: notices-check isn't wired into any GitHub Actions workflow, only into local just check.

Josh-Schoen:
## Correction — the notices are already stale against current `main`

My "approve" verified `gen-notices -check` against this PR's own branch, which is fine as far as it goes — but this branch is 5 commits behind `main`, and one of those commits (#390, a Dependabot "bump the go-minor-patch group" PR) bumped 9 dependency versions. I merged this branch with current `main` in a scratch worktree and re-ran the check:

```
$ GOWORK=off go run ./cmd/gen-notices -check
gen-notices: THIRD-PARTY-NOTICES.md is out of date — run `just notices`
exit status 1
```

Confirmed at the content level too — regenerating shows 14 version-string line-pairs drifted (`anthropic-sdk-go` v1.68.0→v1.70.1, five `aws-sdk-go-v2` submodules, `go-sql-driver/mysql`, `jhump/protoreflect`, `klauspost/compress` ×3, `golang.org/x/crypto`). `NOTICE` itself is unchanged — same 8 NOTICE-bearing modules, same attribution text — so this isn't a missing-attribution problem, just stale version numbers in `THIRD-PARTY-NOTICES.md`.

This is a direct, concrete instance of the CI gap I flagged in my first comment: since `notices-check` isn't wired into any GitHub Actions workflow, nothing would have caught this before merge, and the notices would land already-inaccurate on day one.

**Revised verdict: approve once rebased and regenerated** (small: `git merge main && just notices && git commit`). Everything else in my original review stands — the generator itself works correctly and produced accurate output against the state it was run against; it was checked against a moving target that moved.

Josh-Schoen (changes requested):
Requesting changes — see the correction comment above. The notices are already out of date against current main (confirmed: merging this branch with main and re-running gen-notices -check fails, 14 version-string lines drifted from a dependency-bump PR that landed after this branch's base). Small fix: rebase, run just notices, recommit. Withdrawing my earlier approve.

ilsiepotamus:
Fixed in `dadf8aa8` + `211eaf70` — and your CI-gap note turned out to matter much more than either of us thought.

**The stale notices**: merged `main`, re-ran `just notices`. Exactly the 14 version-string line-pairs you identified, `NOTICE` unchanged (same 8 NOTICE-bearing modules), confirming your read that this was stale versions and not missing attribution.

**The CI gap**: wired into the lint job. It runs the generator's `-check` directly rather than `just notices-check`, since that job has no `just` installed and adding a toolchain for one step is not worth it. Verified the gate has teeth: it passes on the regenerated tree, fails when a single version string is perturbed, passes again once restored.

**Then the gate failed CI, and it was right to.** Chasing it found a real compliance bug, not drift.

`gen-notices` scopes itself with `go list -deps`, which honours build constraints for the **host** GOOS/GOARCH. So the committed notices covered darwin/arm64 — one of the four platforms `release.yml` ships. Two consequences:

1. A file generated on a mac could never satisfy `-check` on linux CI. The gate would have been unpassable rather than useful.
2. Worse: **linux and windows users were being shipped licenses with no attribution**. Regenerating across the release matrix added 9 components a mac-only run omitted — `danieljoos/wincred` (windows credential store), `godbus/dbus` (linux), and the `modernc.org/sqlite` pure-Go stack with `libc`/`mathutil`/`memory`. That is the exact Apache-2.0 §4(d)/license-inclusion failure this PR exists to close, hidden by a host-scoped generator.

`listDeps` now pins GOOS/GOARCH and `collect` unions the dependency sets across linux/amd64, darwin/amd64, darwin/arm64 and windows/amd64. Output is identical regardless of who runs it, and covers what each platform's users actually receive. `collect` splits into `collectPkgs` + `finish` since the loop is now targets × platforms. `NOTICE` is still unchanged.

Verified on both platforms rather than assumed: the committed file passes `gen-notices -check` on my darwin/arm64 host **and** inside a `golang:1.26-bookworm` linux container.

Worth noting for the record: your approve, your own correction, and my first fix all checked the notices on a single platform. It only surfaced because the gate ran somewhere other than where the file was generated — which is the argument for the gate.

CI: Go Lint (the job carrying the gate) passes; 18 green, no failures.

ilsiepotamus:
Rebased onto current `main` at `de0eebb1` (previous fix commits: `dadf8aa8`, `211eaf70`).

The rebase itself needed one hand-resolved conflict in `.github/workflows/release.yml` — #284 had added a `--binary` flag to the Unix `sha256sum` call and switched the Windows hash to a lowercased/no-newline format after this branch's base; kept both of those and layered the NOTICE/LICENSE/THIRD-PARTY-NOTICES.md packaging on top.

Regenerating turned up a second real bug: I'd run `just notices` in an earlier pass on this branch but never committed the output before pushing, so the first re-push still shipped stale content and CI's `gen-notices -check` (run with `GOWORK=off`, which resolves a slightly different module set than a plain `go run`) caught it. Fixed in `de0eebb1` — regenerated with `GOWORK=off go run ./cmd/gen-notices`, verified `-check` passes under those exact conditions, then committed and pushed.

Verified on the pushed head: `go build -tags fts5 ./...`, `gofmt -l` clean, `golangci-lint run --build-tags fts5 ./...` (0 issues), `buf lint`/`buf format --diff`/`buf breaking --against main` all clean, CI green.

PR #386, opened 2026-09-04
feat(agent): resource-await park — end the turn on a success result awaiting a resource

ilsiepotamus:
## Problem

A tool that starts a long-running job follows the MCP event-based resource pattern: it returns immediately with a success result naming a resource (`gdp://jobs/{id}`), and the caller subscribes and acts on `notifications/resources/updated`. Neither existing mechanism fits:

- **In-turn resource wait** (#343) triggers only on *failed* calls, blocks inside the turn (60s budget), and *retries the tool* on wake — re-running a job starter starts a second job.
- **HITL park** ends the turn durably but parks *pre-execution*, and an approved resume *re-executes* ordinary tools (same double-start hazard); its injection primitive was gated to `contact_human`.

## Mechanism

1. **Marker** — `protocol.MetaAwaitResource` (`com.teradata.loom/await-resource`), a result-`_meta` key a server stamps on a *successful* `tools/call` result: `{"uri": "..."}`. `protocol.CallToolResult` gains the `_meta` field it rides on; the MCP adapter stamps it onto the generic contract as `shuttle.Result.AwaitResource` (any embedder-owned bridge can stamp the field directly). Core 2026-07-28 has no native long-running-op marker — the tasks extension is a separate, unadopted spec — hence the reverse-DNS loom key.
2. **Hold** — a successful result carrying `AwaitResource` is withheld from the transcript (no tool row, the parked rowless-tail state) *after* the embedder's new `ResourceAwaitHandler.PrepareWait` accepts (subscribe-before-park; refusal or no handler degrades to pass-through). Gated by the same durability preconditions as the HITL pre-scan, conversation loop only (a resumed batch never nests a park).
3. **Park** — after the batch loop, ONE grouped parked `HumanRequest` (RequestType `parked`, new Kind `resource`, per-call descriptor `{seq, kind, tool, params, uri}`) and `TurnParkedError`. A persist failure fail-safes: every held call's original result commits as its normal tool row (new `commitToolRow` — dispatchOneCall's commit tail extracted verbatim) and `AbandonWait` fires.
4. **Resume** — `ParkDecision.Results` (call id → terminal payload) rides the caller's payload like `Answers`; the row's descriptor kind (not the payload) routes resource items away from **every** dispatch arm: approved+payload injects the payload verbatim as the call's successful result (`synthesizeParkedResult`), refusal synthesizes `resource_wait_failed` with the embedder's reason, approved-without-payload synthesizes `MISSING_RESULT`. **No arm re-executes the tool body.**

## Embedder contract

The embedder owns everything with a lifetime longer than a call: the subscription (`client.Subscribe`/`SubscribeResource`) and its reconnect/restart reconciliation, the terminal `ReadResource`, recording the decision, at-most-once resume delivery, and TTL policy. First consumer: tera-backend's MCP resource watcher (design in that repo, `docs/mcp-resource-subscription-park-design.md`).

## Verification

- 7 new `TestResourceAwait_*` tests through the real conversation loop (hold+park shape, sibling rows land, subscribe-before-park ordering, no-handler / refusal / store-failure degradations incl. `AbandonWait`, injection verbatim, no re-execution on any resume arm, `resource_wait_failed` / `MISSING_RESULT`).
- Adapter stamping tests (well-formed marker stamps; absent/malformed does not) + protocol table test.
- Full `pkg/agent`, `pkg/shuttle`, `pkg/mcp/...` suites green with `-race -count=1`; `golangci-lint` clean.
- Docs: `docs/architecture/resource-await-park.md` + the lifecycle guide's success-result carve-out.

### DIFF
--- a/docs/architecture/resource-await-park.md
+++ b/docs/architecture/resource-await-park.md
@@ -1,0 +1,92 @@
+# Resource-Await Park
+
+**Status:** Implemented (`pkg/agent/park_resource.go`)
+**Relates to:** HITL park-and-resume (`docs/architecture/hitl-park-and-resume.md`),
+in-turn resource wait (`pkg/mcp/adapter/resource_wait.go`, issue #343)
+
+## Problem
+
+A tool that starts a long-running job follows the MCP event-based resource
+pattern: it returns immediately with a success result carrying a
+`resource_link` (`gdp://jobs/{id}`), and the caller is expected to subscribe
+and act on `notifications/resources/updated`. Neither existing mechanism fits:
+
+- **In-turn resource wait** (issue #343) triggers only on *failed* calls,
+  blocks inside the turn (60s budget, bounded further by the embedder's
+  request watchdog), and *retries the tool* on wake — re-running a job
+  starter starts a second job.
+- **HITL park** ends the turn durably, but parks *pre-execution* on a human
+  gate, and an approved resume *re-executes* ordinary tools — same
+  double-start hazard — while its result-injection primitive was gated to
+  `contact_human`.
+
+## Mechanism
+
+```
+tool executes ── success result + _meta["com.teradata.loom/await-resource"]
+        │                          (adapter stamps shuttle.Result.AwaitResource)
+        ▼
+ResourceAwaitHandler.PrepareWait(session, call, uri)     ← embedder subscribes
+        │ error → result passes to the model unchanged (degrade)
+        ▼
+row withheld (call stays rowless) ─ after the batch loop, ONE parked
+HumanRequest row (RequestType "parked", Kind "resource", descriptor
+{seq, kind, tool, params, uri} per call) ─ TurnParkedError ends the turn
+        │
+        ▼  … resource reaches terminal state; embedder records the decision …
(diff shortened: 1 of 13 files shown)


InfinityOverflow (approved):
Approved just minor requests good to have in before merge specially untested branch

Fowler smells (judgement calls):
- Duplicated Code: maybeParkResourceBatch (park_resource.go:452-491) reconstructs the same HumanRequest/store.Store/notifier.Notify/TurnParkedError sequence as maybeParkBatch (park.go:628-660), differing only in Store-failure handling. Candidate: shared persistParkRequest(items, onFailure).
- Duplicated Code (minor): batch-wide batchIDCount (agent.go:2757-2759) and checkParkItemIDs (park.go:698-709) both do ID-uniqueness bookkeeping via different shapes (count-map vs. seen-set).

 Scope creep:
- park_resource.go:98-106 adds a guard rejecting holds on empty/duplicate toolCall.ID, mirroring checkParkItemIDs — not mentioned in the spec and untested (no test exercises this refusal path). Low risk, reasonable safeguard, but outside the stated verification list.

PR #387, opened 2026-09-04
fix(agent): give synthetic messages their own persisted role

InfinityOverflow:
## Description

Four kinds of agent-authored message content were persisted under the literal `Role: "user"` purely as an LLM-attention optimization (the model pays more attention to user-role turns): a loaded skill's full instructions, an end-of-turn hygiene-retry nudge, an empty-response retry nudge, and a final-answer synthesis prompt. Because all four used the literal `"user"` role, the TUI (and any other consumer reading persisted messages — `GetConversationHistory`, the live `SubscribeToSession` stream, the recall tool) could not tell them apart from something the human actually typed, so they rendered as ordinary human chat bubbles.

This PR gives each its own persisted role (`skill_body`, `hygiene_injection`, `empty_response_retry`, `synthesis_prompt`) while preserving the LLM-attention behavior unchanged: a single fold function, `normalizeWireRoles`, rewrites all four to `"user"` on a throwaway copy, called once at the top of `Agent.chatWithRetry` — the sole entry point every conversational LLM call (streaming, direct, retried) passes through. **No LLM provider client is touched** (Anthropic, Bedrock, OpenAI, Gemini, Ollama, Azure OpenAI all still only ever see `Role: "user"` on the wire, byte-for-byte as before). Persistence, the TUI, the live streaming gate, and internal "last user message" scans (lazy-tool triggers, graph-memory recall query, `firstUserLine`) all now see the distinct role.

## Type of Change

- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🔧 Refactoring (no functional changes)
- [ ] 🧪 Test update

## Related Issues

N/A — no tracked issue; this came out of a design exploration (`docs/architecture/skill-injection-flow.md`, not included in this PR — see Additional Context) into why the skill-body sidecar renders as a human message in the TUI.

## Changes Made

- **The fold** (`pkg/agent/llm_retry.go`): `IsSyntheticWireUserRole(role string) bool` — the single source of truth for which persisted roles are synthetic — and `normalizeWireRoles`, called at the top of `chatWithRetry`. No-op copy elided (returns the input slice unchanged, no allocation) when nothing needs folding, since most turns carry none of these roles.
- **The four producers flipped** to their own role: skill-body sidecar (`agent.go`, via `manage_skills`), hygiene-retry injection (`hygiene.go`), empty-response retry nudge (`agent.go`), synthesis prompt (`agent.go`, in `synthesizeFinalResponse`).
- **Live streaming gate** (`pkg/server/multi_agent.go`): `isNewMessageUpdateRole` widened to delegate to `IsSyntheticWireUserRole` for the four new roles, so `SubscribeToSession` still surfaces them (previously implicit since they were literally `"user"`).
- **Recall tool** (`pkg/agent/recall_tool.go`): the seq-range recall allowlist widened the same way — these are directly recoverable text content like `user`/`assistant` rows, not `tool` rows (whose exclusion is about needing to re-run the call).
- **TUI rendering** (`internal/tui/components/chat/messages/messages.go`, `internal/tui/components/chat/chat.go`): all four roles render as a muted, labeled note — visually distinct from both the human bubble and the assistant's content block — instead of falling through to either.
- **Docs**: `docs/reference/streaming.md`'s streamed-role list updated to match.
- **No changes**: proto files, database schema/migrations, any LLM provider client (`pkg/llm/anthropic`, `pkg/llm/bedrock`, `pkg/llm/openai`, `pkg/llm/gemini`, `pkg/llm/ollama`, `pkg/llm/azureopenai`).

### Test Checklist

- [x] All new and existing tests pass (`go test -tags fts5 ./...`) for every package this PR touches — see note below on environment limitations for the exact commands used
- [ ] Race detector shows zero race conditions (`just race-check`) — **could not be run locally**: `-race` fails to allocate ThreadSanitizer's shadow memory on the Windows machine this was developed on (a machine-specific limitation, reproduced consistently, unrelated to this code). Needs verification in CI before merge.
- [ ] All checks pass (`just check`) — `just check` reaches a pre-existing `proto-format-check` failure (CRLF/buf-toolchain drift on the Windows checkout, in files this PR never touches) before it gets to build/test; the build/test portions were run directly and are clean

### Test Coverage

- [ ] No tests needed — N/A, tests were added throughout

New/extended tests: the fold and its no-allocation fast path (`pkg/agent/llm_retry_wire_roles_test.go`), the streaming gate (`pkg/server/multi_agent_new_message_gate_test.go`), TUI rendering (`internal/tui/components/chat/messages/messages_test.go`), the recall tool (`pkg/agent/recall_tool_test.go`), the message-role constants (`internal/message/message_test.go`), and full `Chat()`-driven integration tests for all four producer flips (`pkg/agent/skill_role_separation_test.go`, `pkg/agent/hygiene_test.go`, plus two more in the same style).

Known, unrelated pre-existing flake observed during testing on this Windows dev machine: some tests using `t.TempDir()` for a context-dump or SQLite-registry sink fail on Go's automatic cleanup (`RemoveAll`: "process cannot access the file because it is being used by another process") — reproduced and confirmed across many tests spanning both files this PR touches and files it doesn't, always cleanup-only with zero assertion failures. Not caused by this change.

## Proto Changes

## Documentation

- [ ] README.md updated (if user-facing changes) — N/A, no user-facing README claims affected
- [ ] ARCHITECTURE.md updated (if architectural changes)
- [ ] TODO.md updated (if implementation plan changed)
- [x] No documentation changes needed beyond `docs/reference/streaming.md` (updated)

**Known gap**: `docs/architecture/skills-system.md`, `docs/architecture/agent-system-design.md`, and `docs/architecture/data-flows.md` still describe the old "skill body enters as a `Role="user"` sidecar" behavior and were not updated as part of this PR — flagged during review, left as a follow-up rather than expanding this PR's scope further.

## Security

## Performance

- [x] No performance impact — the fold's fast path avoids allocating when no message needs folding (the common case, most turns); a two-pass scan cost is accepted in exchange for that.

## Deployment Notes

- [x] No special deployment steps needed — no schema/migration, no config changes. Existing persisted rows already tagged `"user"` from before this change simply continue rendering as ordinary user messages; no backfill needed.

### Required

- [x] I have read the CONTRIBUTING.md guidelines
- [x] My code follows the project's code style (gofmt, proto format)
- [x] I have performed a self-review of my own code
- [x] I have commented my code in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (see Known gap above)
- [x] My changes generate no new warnings or errors
- [ ] All CI checks pass (proto, lint, test, build, security) — pending CI run; local verification is limited by the environment notes above
- [ ] Zero race conditions detected — pending CI (`-race` unavailable locally, see Testing)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with `go test -tags fts5 ./...` (not `-race`, see above)
- [x] Commit messages follow convention: `<type>(<scope>): <subject>`

### Optional

- [ ] I have updated the TODO.md with implementation notes
- [ ] I have added usage examples
- [ ] I have updated relevant diagrams
- [ ] I have tested on multiple platforms (Linux, macOS, Windows) — developed and tested on Windows only

## Additional Context

This branch was implemented via a written spec and implementation plan (task-by-task, with a per-task review and one final whole-branch review), including:
- One fix round during implementation (a `style()` case that was never actually wired up to the code that was supposed to use it).
- A final whole-branch review that found two more producers (`empty_response_retry`, `synthesis_prompt`) using the same "persisted as literal `user`" pattern that the original design hadn't scoped in — both are included in this PR.
- A final review fix wave for two other findings: the stale `docs/reference/streaming.md` entry, and the recall-tool exclusion bug (both described above, both fixed and re-reviewed).

---

**Before submitting:** Make sure all required checklist items are completed and all CI checks pass.

### DIFF
--- a/docs/reference/streaming.md
+++ b/docs/reference/streaming.md
@@ -1092,7 +1092,7 @@ message SessionStatusUpdate {
 Sent when a new message is added to the session conversation (e.g., from a sub-agent response, tool result, or user input).
 
 **Fields**:
-- `role`: Message role - `user`, `assistant`, or `tool` (only these roles are sent)
+- `role`: Message role - `user`, `assistant`, `tool`, `skill_body`, or `hygiene_injection` (only these roles are sent). `skill_body` and `hygiene_injection` are synthetic, agent-authored roles — a loaded skill's instructions and a hygiene-retry nudge, respectively — persisted separately from the literal end user; see `pkg/agent/llm_retry.go`'s `IsSyntheticWireUserRole`.
 - `content`: The message text
 - `message_timestamp`: When the message was created (Unix timestamp)
 - `tool_name`: 🚧 Defined in proto but not yet populated by the server implementation
--- a/docs/reference/streaming.md
+++ b/docs/reference/streaming.md
@@ -1092,7 +1092,7 @@ message SessionStatusUpdate {
 Sent when a new message is added to the session conversation (e.g., from a sub-agent response, tool result, or user input).
 
 **Fields**:
-- `role`: Message role - `user`, `assistant`, `tool`, `skill_body`, or `hygiene_injection` (only these roles are sent). `skill_body` and `hygiene_injection` are synthetic, agent-authored roles — a loaded skill's instructions and a hygiene-retry nudge, respectively — persisted separately from the literal end user; see `pkg/agent/llm_retry.go`'s `IsSyntheticWireUserRole`.
+- `role`: Message role - `user`, `assistant`, `tool`, `skill_body`, `hygiene_injection`, `empty_response_retry`, or `synthesis_prompt` (only these roles are sent). `skill_body`, `hygiene_injection`, `empty_response_retry`, and `synthesis_prompt` are synthetic, agent-authored roles — a loaded skill's instructions, a hygiene-retry nudge, an empty-response retry nudge, and a forced-final-answer synthesis prompt, respectively — persisted separately from the literal end user; see `pkg/agent/llm_retry.go`'s `IsSyntheticWireUserRole`.
 - `content`: The message text
 - `message_timestamp`: When the message was created (Unix timestamp)
 - `tool_name`: 🚧 Defined in proto but not yet populated by the server implementation
--- a/internal/message/message.go
+++ b/internal/message/message.go
@@ -38,6 +38,16 @@ const (
 	// HygieneInjection is an end-of-turn hygiene retry nudge. Same
 	// wire-folding treatment as SkillBody, different producer.
 	HygieneInjection Role = "hygiene_injection"
+
+	// EmptyResponseRetry is the one-shot nudge appended when the LLM
+	// returns an empty response, asking it to try again. Same
+	// wire-folding treatment as SkillBody, different producer.
+	EmptyResponseRetry Role = "empty_response_retry"
+
+	// SynthesisPrompt is the forced-final-answer request appended when the
+	// turn/execution budget is exhausted. Same wire-folding treatment as
+	// SkillBody, different producer.
+	SynthesisPrompt Role = "synthesis_prompt"
 )
 
 // FinishReason represents the reason a message finished.
--- a/internal/message/message.go
+++ b/internal/message/message.go
@@ -29,6 +29,15 @@ const (
 	Assistant Role = "assistant"
 	Tool      Role = "tool"
 	System    Role = "system"
+
+	// SkillBody is a loaded skill's full instructions, injected by
+	// manage_skills as a sidecar. Reaches the LLM as user-turn content
+	// (folded in pkg/agent.normalizeWireRoles) but is not the human's turn.
+	SkillBody Role = "skill_body"
+
+	// HygieneInjection is an end-of-turn hygiene retry nudge. Same
+	// wire-folding treatment as SkillBody, different producer.
+	HygieneInjection Role = "hygiene_injection"
 )
 
 // FinishReason represents the reason a message finished.
--- a/internal/message/message_test.go
+++ b/internal/message/message_test.go
@@ -1,0 +1,29 @@
+// Copyright 2026 Teradata
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package message
+
+import "testing"
+
+// The persisted-role string a producer writes (pkg/agent) and the TUI's
+// typed Role must be the same literal, or the bare cast in
+// internal/tui/adapter/messages.go:ProtoToMessage silently produces a role
+// the TUI never explicitly handles.
+func TestSyntheticRoles_MatchPersistedStrings(t *testing.T) {
+	if SkillBody != "skill_body" {
+		t.Fatalf("SkillBody = %q, want %q", SkillBody, "skill_body")
+	}
+	if HygieneInjection != "hygiene_injection" {
+		t.Fatalf("HygieneInjection = %q, want %q", HygieneInjection, "hygiene_injection")
+	}
+}
--- a/internal/message/message_test.go
+++ b/internal/message/message_test.go
@@ -26,4 +26,10 @@ func TestSyntheticRoles_MatchPersistedStrings(t *testing.T) {
 	if HygieneInjection != "hygiene_injection" {
 		t.Fatalf("HygieneInjection = %q, want %q", HygieneInjection, "hygiene_injection")
 	}
+	if EmptyResponseRetry != "empty_response_retry" {
+		t.Fatalf("EmptyResponseRetry = %q, want %q", EmptyResponseRetry, "empty_response_retry")
+	}
+	if SynthesisPrompt != "synthesis_prompt" {
+		t.Fatalf("SynthesisPrompt = %q, want %q", SynthesisPrompt, "synthesis_prompt")
+	}
 }
--- a/internal/tui/components/chat/chat.go
+++ b/internal/tui/components/chat/chat.go
@@ -376,7 +376,7 @@ func (m *messageListCmp) handleNewMessage(msg message.Message) tea.Cmd {
 		return m.handleNewUserMessage(msg)
 	case message.Assistant:
 		return m.handleNewAssistantMessage(msg)
-	case message.SkillBody, message.HygieneInjection:
+	case message.SkillBody, message.HygieneInjection, message.EmptyResponseRetry, message.SynthesisPrompt:
 		return m.handleNewSyntheticMessage(msg)
 	case message.Tool:
 		return m.handleToolMessage(msg)
@@ -611,7 +611,7 @@ func (m *messageListCmp) convertMessagesToUI(sessionMessages []message.Message, 
 			if msg.FinishPart() != nil && msg.FinishPart().Reason == message.FinishReasonEndTurn {
 				uiMessages = append(uiMessages, messages.NewAssistantSection(msg, time.Unix(m.lastUserMessageTime, 0)))
 			}
-		case message.SkillBody, message.HygieneInjection:
+		case message.SkillBody, message.HygieneInjection, message.EmptyResponseRetry, message.SynthesisPrompt:
 			// Synthetic content, not the human's turn — do not touch
 			// lastUserMessageTime, which anchors the assistant response timer
 			// to the real user's send time.
--- a/internal/tui/components/chat/chat.go
+++ b/internal/tui/components/chat/chat.go
@@ -376,6 +376,8 @@ func (m *messageListCmp) handleNewMessage(msg message.Message) tea.Cmd {
 		return m.handleNewUserMessage(msg)
 	case message.Assistant:
 		return m.handleNewAssistantMessage(msg)
+	case message.SkillBody, message.HygieneInjection:
+		return m.handleNewSyntheticMessage(msg)
 	case message.Tool:
 		return m.handleToolMessage(msg)
 	}
@@ -385,6 +387,12 @@ func (m *messageListCmp) handleNewMessage(msg message.Message) tea.Cmd {
 // handleNewUserMessage adds a new user message to the list and updates the timestamp.
 func (m *messageListCmp) handleNewUserMessage(msg message.Message) tea.Cmd {
 	m.lastUserMessageTime = msg.CreatedAt
+	return m.listCmp.AppendItem(messages.NewMessageCmp(msg))
+}
+
+// handleNewSyntheticMessage appends a skill-body or hygiene-injection
+// message to the list without treating it as the human's turn.
+func (m *messageListCmp) handleNewSyntheticMessage(msg message.Message) tea.Cmd {
 	return m.listCmp.AppendItem(messages.NewMessageCmp(msg))
 }
 
@@ -603,6 +611,11 @@ func (m *messageListCmp) convertMessagesToUI(sessionMessages []message.Message, 
 			if msg.FinishPart() != nil && msg.FinishPart().Reason == message.FinishReasonEndTurn {
 				uiMessages = append(uiMessages, messages.NewAssistantSection(msg, time.Unix(m.lastUserMessageTime, 0)))
 			}
+		case message.SkillBody, message.HygieneInjection:
+			// Synthetic content, not the human's turn — do not touch
+			// lastUserMessageTime, which anchors the assistant response timer
+			// to the real user's send time.
+			uiMessages = append(uiMessages, messages.NewMessageCmp(msg))
 		}
 	}
 
--- a/internal/tui/components/chat/messages/messages.go
+++ b/internal/tui/components/chat/messages/messages.go
@@ -140,10 +140,11 @@ func (m *messageCmp) View() string {
 		return m.style().PaddingLeft(1).Render(m.anim.View())
 	}
 	if m.message.ID != "" {
-		// this is a user or assistant message
 		switch m.message.Role {
 		case message.User:
 			return m.renderUserMessage()
+		case message.SkillBody, message.HygieneInjection:
+			return m.renderSyntheticNote()
 		default:
 			return m.renderAssistantMessage()
 		}
@@ -178,9 +179,14 @@ func (msg *messageCmp) style() lipgloss.Style {
 	}
 
 	style := t.S().Text
-	if msg.message.Role == message.User {
+	switch msg.message.Role {
+	case message.User:
 		style = style.PaddingLeft(1).BorderLeft(true).BorderStyle(borderStyle).BorderForeground(t.Primary)
-	} else {
+	case message.SkillBody, message.HygieneInjection:
+		// Always the plain, unfocused shape — a synthetic note is never the
+		// focus target a human bubble or assistant block can be.
+		style = style.PaddingLeft(2)
+	default:
 		if msg.focused {
 			style = style.PaddingLeft(1).BorderLeft(true).BorderStyle(borderStyle).BorderForeground(t.GreenDark)
 		} else {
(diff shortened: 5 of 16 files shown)


InfinityOverflow:
PR #388 closes the test gaps found in the development of this PR

ilsiepotamus (changes requested):
# PR #387 Review: `fix(agent): give synthetic messages their own persisted role`

**Branch**: `InfinityOverflow:...` → `main` @ merge-base `5d150b2e` (main has moved 3 commits ahead: 2 dependency/CI bumps + `fb727480` "resource-await park" in `pkg/agent`). `git merge-tree` simulation against current `origin/main` produced a clean merge with **zero conflicts**. Live GitHub state: `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND`, **all CI checks green** (build×3 platforms, Go Lint, Proto Lint & Breaking Changes, CodeQL, Fuzz Tests, Coverage Requirements, Unit Tests, Security Scan). The task brief's "mergeable: UNKNOWN" was stale/uncomputed, not a real conflict.

Verified locally in this worktree: `go build -tags fts5 ./...` clean, `go vet -tags fts5 ./...` clean, `gofmt -l` clean on every changed file, `go test -tags fts5 -race ./pkg/agent/...` **PASS** (30s, incl. all new tests), `go test -tags fts5 -race ./pkg/server/...` **PASS** (21s). This independently discharges the author's own unchecked checklist items ("race detector... could not be run locally" / "pending CI").

---

## 1. DESCRIPTION

**What a "synthetic message" is here**: four kinds of agent-generated conversational content that must reach the LLM as `Role: "user"` (an attention optimization — models weight user turns more heavily) but are not text the human typed:
- **Skill-body sidecar** — a loaded skill's full instructions, injected by `manage_skills(load)`, queued in `batchState.pendingSidecars` and flushed after all tool results in the batch (`pkg/agent/agent.go:3039-3053`, drained at `:2776-2784`).
- **Hygiene-retry injection** — an end-of-turn task-board hygiene nudge (`pkg/agent/hygiene.go:70-83`, `runEndOfTurnHygiene`).
- **Empty-response retry nudge** — a one-shot "your previous response was empty" prompt (`pkg/agent/agent.go:2645-2656`).
- **Synthesis prompt** — the forced-final-answer request fired when turn/tool budget is exhausted (`pkg/agent/agent.go:3075-3084`, `synthesizeFinalResponse`).

**The bug being fixed**: all four were persisted under the *literal* string `Role: "user"`. Every consumer that reads persisted history — `GetConversationHistory` (`pkg/server/server.go:546-567`), the live `SubscribeToSession` stream, the TUI, and the recall tool — could not distinguish them from genuine human input, so e.g. the TUI rendered a skill's multi-thousand-token instruction dump as an ordinary chat bubble the user appeared to have typed. Internal "find the last user message" scans (lazy-tool-disclosure trigger at `agent.go:2317-2326`, graph-memory recall query at `agent.go:3365-3373`) were also vulnerable to picking up a stray sidecar as "the current user message" if a `Chat()` call ended right after one was appended (e.g., an error path with no trailing assistant reply) — a real, if narrow, cross-turn misattribution bug, not an LLM-API-rejection issue.

**The fix**: four new persisted roles (`skill_body`, `hygiene_injection`, `empty_response_retry`, `synthesis_prompt`, added as both `internal/message.Role` consts and raw strings in `pkg/agent`), plus one fold function `normalizeWireRoles` (`pkg/agent/llm_retry.go:51-71`) that rewrites them all back to `"user"` on a throwaway copy, called once at the top of `chatWithRetry` (`llm_retry.go:81`) — confirmed by grep to be the sole `a.llm.Chat(...)` gateway inside `pkg/agent` (all direct call sites are internal to `llm_retry.go`'s retry/streaming branches). Provider clients (anthropic/bedrock/openai/gemini/ollama/azureopenai) are untouched — verified zero diff in `pkg/llm/*`.

**Subsystems touched**: `internal/message` (role consts), `pkg/agent` (4 producers + the fold + recall tool), `pkg/server` (live-stream role gate), TUI (`internal/tui/components/chat/...`), `docs/reference/streaming.md`. No proto, no DB schema/migration.

---

## 2. SHAPE REVIEW

**Proto-first**: correctly judged not to apply. `proto/loom/v1/loom.proto:1437` (`NewMessageUpdate.role`) is already a free-form `string`, and both SQL schemas store `role TEXT NOT NULL` with no `CHECK` constraint (`pkg/storage/sqlite/migrations/000001_initial_schema.up.sql:58`, `pkg/storage/postgres/migrations/000001_initial_schema.up.sql:24`) — a new string *value* needs no schema change. Good call, consistent with "proto is law" (a schema change would be needed only for a new *field*).

**Reuse vs. parallel mechanism**: the task brief asks whether this reuses the existing `LLMRole` machinery. It's important to note **`loomv1.LLMRole` is an unrelated enum** (`pkg/agent/agent.go:3962-4061`) — it selects which LLM backend serves a purpose (judge/orchestrator/classifier/compressor/agent), not who authored a message. There is no pre-existing typed enum for message-sender role anywhere in `pkg/agent`/`pkg/types` — `pkg/types.Message.Role` (`pkg/types/types.go:100`) has always been a bare `string`, and raw string literals for `"user"`/`"assistant"`/`"tool"` are the established, repo-wide convention (dozens of sites in `pkg/llm/*`, `pkg/server/*`, `pkg/task/*`, etc.). The PR extends that same field with new agreed values and centralizes the "is this synthetic" question behind one exported predicate — that's reuse, not a bolted-on parallel system, and it's the correct choice.

**Central-predicate goal only half-achieved** — `IsSyntheticWireUserRole`'s doc comment claims to be "the single source of truth," and it *is* reused correctly across packages (`pkg/server/multi_agent.go:816-822` delegates to it; `pkg/agent/recall_tool.go:127` delegates to it). But **within `pkg/agent` itself the four role strings are hardcoded independently in five places** with no compiler-enforced link:
- `pkg/agent/llm_retry.go:38` — the switch's case literals
- `pkg/agent/agent.go:2650` (`empty_response_retry`), `:3047` (`skill_body`), `:3080` (`synthesis_prompt`)
- `pkg/agent/hygiene.go:76` (`hygiene_injection`)

None of the four producer sites reference a shared Go constant (`pkg/agent` doesn't define one), so a future rename at a producer site without a matching update to the `llm_retry.go:38` switch would silently stop folding that role to `"user"` on the wire — no compile error, just a role string reaching the LLM raw. This is exactly the "hardcoded strings that should go through an existing enum/constant" pattern the project's CLAUDE.md flags ("Dont take shortcuts: No hardcodes"). Note: it would be a *worse* choice architecturally for `pkg/agent` to import `internal/message` for its constants — that package is explicitly TUI-domain ("Package message provides message types compatible with Crush's interface", `internal/message/message.go:14`), and a core-framework package depending on it would be a backwards dependency. The fix is a small local const block in `pkg/agent`, not a cross-package import — worth a note for a follow-up, not severe enough to block.

**Table-driven tests**: partially followed. `TestNewMessageUpdateGate_IncludesSyntheticRoles` (`pkg/server/multi_agent_new_message_gate_test.go`) is properly table-driven. But `pkg/agent/skill_role_separation_test.go` has three near-identical `Test*_PersistedUnder*Role` functions, and `internal/tui/components/chat/messages/messages_test.go` has four near-identical `TestMessageCmp_*_RendersAsSyntheticNote` functions, each differing only in role/content/expected-label — these are prime table-driven candidates per CLAUDE.md's explicit testing-pattern requirement. Minor style nit, not a correctness issue.

**What's done well**: the PR's own commit history shows genuine self-review catching real problems before submission (a dead `style()` case, wired up in `e870157c`; a recall-tool exclusion bug found on final pass, fixed in `9d749ecb`) — this is exactly the "test current → improve → test again" discipline the project values. The `normalizeWireRoles` no-allocation fast path is a nice, correctly-tested efficiency touch (`llm_retry_wire_roles_test.go`'s `require.Same` pointer-identity check is the right tool for that assertion, not `require.Equal`).

---

## 3. COMPREHENSIVE REVIEW

**Producer coverage**: all four producers correctly flipped, and I traced every call site. `dispatchOneCall` (the skill-body producer) is a **shared** function also called from the HITL-park resume path (`pkg/agent/park.go:1180-1182`), and that path's own sidecar-drain loop (`park.go:1186-1190`) calls the same `a.appendMessage` with the same struct — so the parked/resumed-batch path is covered for free, not missed.

**Tool-call/tool-result adjacency (the specific concern raised)** — verified safe, not merely assumed:
- Sidecars are queued in a per-batch buffer and explicitly drained only *after every tool result of the batch is in place* (`agent.go:2776-2784`, comment: "Order within the batch preserved"; `agent.go:3039-3043` states the invariant directly: "Sidecars ... are appended AFTER every tool_result in the batch, so tool_use↔tool_result adjacency is preserved (Anthropic pairing)"). This logic is untouched by the PR — only the `Role:` value at the append site changed, not the ordering.
- The compaction boundary's pair-atomicity guard, `adjustCompressionBoundary` (`pkg/agent/segmented_memory.go:378-424`), keys strictly on `Role == "assistant"` with `ToolCalls` and `Role == "tool"` with `ToolUseID` — none of the four new roles are ever assistant-with-toolcalls or tool-with-tooluseid, so this guard is structurally unaffected.
- `normalizeWireRoles` runs once, at the top of `chatWithRetry`, before any provider client sees the slice — so by the time messages reach the Anthropic/Bedrock/etc. client code that groups consecutive `tool` rows into one grouped turn, every synthetic role already reads `"user"`, identical to pre-PR wire shape for that dimension.

**Two real gaps the PR's own "final whole-branch review" pass missed**, both inside the conversation-history compaction pipeline the brief asked me to scrutinize:

1. **(Major) Silent content loss in the compaction fallback path.** `pkg/agent/memory_compressor.go:92-127` (`LLMCompressor.simpleCompress`) and `:206-239` (`SimpleCompressor.CompressMessages`) both `switch msg.Role` over exactly `"user"`/`"assistant"`/`"tool"`/`"system"` with **no `default` case** — any message carrying one of the four new roles produces nothing and is silently dropped from the emergency/heuristic summary. This fallback fires whenever the LLM-backed compactor errors or returns empty (`memory_compressor.go:76-84`) — a real production path (`context_compilation.go:696-710` calls it with the lock released across a 120s network call). Pre-PR, these rows were literally `"user"` and would have hit `case "user":` and been preserved (truncated) in the fallback summary; post-PR they vanish with **no log line**, since the outer fold code (`context_compilation.go:739-748`) only logs when the *whole* compressor call yields an empty string, which `simpleCompress` never does (it floors at `"Previous exchanges"`). This directly contradicts the PR's "no functional changes to wire content" framing for this one path. Partial mitigation exists only for `skill_body`: `foldedSkillLoads` (`context_compilation.go:757-773`, keyed off the *tool-result* row's "Skill loaded: " prefix, not the sidecar) independently tracks which skills were active and appends a `"[Folded active skill(s): ...]"` note regardless of compressor success — so the fact a skill was loaded survives, but the skill's actual instructional text does not. The other three roles (`hygiene_injection`, `empty_response_retry`, `synthesis_prompt`) have no such safety net. No test in this PR exercises fold/compaction with any of the new roles present — the "conversation-history compaction/compression pipeline" mentioned in the brief is exactly where this PR needed, and lacks, a round-trip test.

2. **(Minor) Temporal-grounding stamp silently dropped on the wire for all four roles.** `renderLocked` (`pkg/agent/context_compilation.go:186`) prepends a `"[Mon 2026-09-09 HH:MM MST] "` stamp to compiled content **only when `m.Role == "user"`**, and this runs (via `compileLocked` → `GetMessagesForLLM` → `Session.GetMessages`) *before* `normalizeWireRoles` ever sees the message, since `SegmentedMemory` is the default memory manager for every session (`pkg/agent/memory.go:589-609`, `ensureSessionMemory`, called unconditionally). So a message persisted as `skill_body`/`hygiene_injection`/`empty_response_retry`/`synthesis_prompt` no longer receives this stamp — it is later renamed to `"user"` by the fold with no stamp added. Pre-PR, since these rows *were* literally `"user"` at persist time, they received the stamp. This is a small, likely-harmless behavior change (these rows are agent-generated at the instant they're sent, so a wall-clock stamp is close to redundant) but it does mean the PR's repeated claim of "byte-for-byte as before" on the wire is not strictly true for the default (segmented-memory) configuration. Untested by any of the new tests, none of which construct an `Agent` with `SegmentedMem` populated when checking role/content.

**Backward compatibility**: correctly described. Old rows already persisted as `"user"` are never rewritten and will forever render as ordinary user messages (no migration, no backfill) — verified accurate against the unconstrained `TEXT` column, and a reasonable go-forward-only fix.

**Concurrency**: no new races. `normalizeWireRoles` builds a private copy from an already-private slice (`session.GetMessages()` returns a copy under `RLock`, or a freshly-built `compileLocked()` slice) — the range-loop `m := messages[i]; m.Role = "user"` mutates only the local copy, `messages[i].Role` (an immutable string field) is never touched, verified by direct code reading. Confirmed empirically too: `pkg/agent` and `pkg/server` full suites both pass `-race` locally (0 races, exit 0).

**Test coverage**: strong for the producers/predicate themselves — `llm_retry_wire_roles_test.go` genuinely exercises the wire behavior (fake "provider" records actual roles received, not just constructed structs) and `skill_role_separation_test.go` drives real `Agent.Chat()` end-to-end rather than only unit-constructing messages. Gaps: (a) the compaction-pipeline interaction above is entirely untested; (b) `internal/tui/components/chat/chat.go` — the file containing the new `handleNewSyntheticMessage` dispatch and the `convertMessagesToUI` synthetic-role case — has **no test file at all** (`ls internal/tui/components/chat/*_test.go` → no matches), so this routing glue is unverified beyond the (well-tested) rendering logic one layer down in `messages/messages.go`.

**Documentation accuracy**: `docs/reference/streaming.md:1092-1093` was correctly and accurately updated. The PR's own "Known gap" is honest and verified accurate — `docs/architecture/skills-system.md` (lines 5, 57, 157, 771, 898), `agent-system-design.md` (lines 369, 408), and `data-flows.md:465` all still explicitly state the skill body is delivered/persisted as a plain "user-role message" / `Role="user"`, confirmed stale by grep. Two smaller staleness spots the author's self-audit missed: the code comment at `pkg/agent/agent.go:2687` ("runEndOfTurnHygiene already appended the synthetic **user** message") still says "user" rather than "hygiene_injection", and the proto comment at `proto/loom/v1/loom.proto:1436` ("Message role (user, assistant, tool, system)") was left listing only the old four roles even though the wire-visible role set (per this PR's own doc update elsewhere) now has seven values.

---

### Summary

No blockers. Build/vet/gofmt clean; both changed packages pass `-race`; adjacency invariant for tool_calls/tool-results explicitly verified safe by design and unaffected; proto/schema correctly judged unnecessary; mergeable with all CI green. One genuine, verifiable, silent-data-loss regression in the compaction fallback path (`pkg/agent/memory_compressor.go:92-127,206-239`) that the PR's stated scope ("audit every consumer of the role string") should have caught but didn't — straightforward to fix (add the four cases, or better, route compaction input through the same `IsSyntheticWireUserRole` idea). Remaining findings are minor: a wire-content claim that's not quite byte-for-byte in the default config, within-package string-literal triplication with no compiler tie-back, two stale comments, three self-disclosed stale docs, and two test-coverage soft spots (compaction round-trip, TUI dispatch glue).

**Majors/Blockers: 1 major, 0 blockers.**

VERDICT: CHANGES NEEDED

ilsiepotamus (approved):
Fixes a real bug: four kinds of agent-authored content (skill body, hygiene-retry nudge, empty-response retry nudge, forced-synthesis prompt) were persisted under the literal `Role: "user"` purely for LLM attention, making them indistinguishable from real human input in the TUI, conversation history, live streaming, and the recall tool.

Traced every call site that reaches an LLM client to confirm nothing bypasses the fold (`normalizeWireRoles` in `chatWithRetry`) — the two other direct-`Chat()` callers (memory compressor, reranker) build their own fresh single-role message slices and never forward session messages, so there's no leak path around it. The no-allocation fast path is real, not just claimed. Wire format to every provider is confirmed byte-for-byte unchanged; no proto/DB changes.

Ran `go build -tags fts5 ./...` and `go test -tags fts5 -race ./pkg/agent/... ./pkg/server/... ./internal/tui/... ./internal/message/...` — all green, including the one checkbox the author flagged they couldn't verify locally (Windows TSan limitation blocked `-race` on their machine). Diff is surgical: the actual behavior-changing lines are ~15 one-word role swaps plus one well-tested fold function.

Approving.

PR #396, opened 2026-09-14
fix(usage): price from the registered catalog; report the turn's total usage alongside the final call's

ilsiepotamus:
## Why

Found while chasing a Loom Cloud incident (Teradata-PE/avmo-tera-cloud#418): users were exhausting a $10/month cost quota on cheap models while their per-session cost display summed to cents. Both halves traced to loom.

1. **Pricing ignored the registered catalog.** `catalog.LookupPricing` built its index from `BuildCatalog()` only — the static table — so a `catalog.Register`ed DB/gateway source could never influence cost, even though `catalog.Lookup` honoured it for context windows and output limits. Every provider client's `calculateCost` therefore fell to its hardcoded fallback for any id the static table did not list. On the OpenAI client, which embedders point at OpenAI-compatible gateways (LiteLLM, vLLM, …), that fallback is the gpt-4o rate ($2.50/M in, $10/M out) applied to **every** gateway alias regardless of what it is — 5–10× too high for DeepSeek/Nova/Haiku-class models. And the client hardcoded the `"openai"` namespace, so even a correctly registered gateway catalog keyed under its own provider could not be reached.

2. **`Response.Usage` is the final LLM call's share, not the turn's.** The conversation loop returned `llmResp.Usage` of whichever call ended the turn; nothing summed across tool-loop iterations, empty-response retries, hygiene retries or the synthesis call. An embedder metering a turn from `Response.Usage` saw ~1/N of a tool-heavy turn while its own per-call provider wrapper billed all N. The agent-level span attributes and `agent.cost.usd` / `agent.tokens.total` metrics had the same 1/N bias.

## What

- `pkg/llm/catalog/pricing.go`: `LookupPricing` consults `DefaultSource()` first (provider normalised), then the static index. A registered entry with no positive rate is treated as unpriced and falls through, so a metadata-only row cannot zero out a price. Behaviour is byte-identical when nothing is registered (the default source *is* the static source).
- `pkg/llm/openai`: new `Config.CatalogProvider` (default `"openai"`, exported `DefaultCatalogProvider`) — the catalog namespace `calculateCost` prices under, so a client fronting a gateway can be told to look where the embedder registered the gateway's ids. The gpt-4o fallback is unchanged.
- `pkg/agent`: new `Response.TurnUsage` and `TurnParkedError.TurnUsage` — the sum of every successful LLM call the turn made (loop calls, empty-response and hygiene retries, final synthesis; a synthesis *failure* still reports the loop's calls). `Response.Usage` keeps its final-call meaning **deliberately**: it is what the persisted assistant row carries as its own `TokenCount`/`CostUSD`, and each tool-calling assistant row in the loop already carries its own, so summing into `Usage` would double count `Session.TotalCostUSD`. The conversation span attributes (`conversation.tokens.*`, `conversation.cost.usd`) and the two agent-level metrics now read `TurnUsage`. `maybeParkBatch` / `synthesizeFinalResponse` gain a `turnUsage` parameter (internal).

No new goroutines or locks; the loop does six integer/float adds per LLM call. `LookupPricing` now takes the `defaultMu` read lock and one source-chain lookup per call — the same cost `catalog.Lookup` already pays on the request path.

## Tests

- `catalog`: registered source wins for a gateway alias and for an override of a static id; an unpriced registered row falls through to the static rate; provider aliases normalise. Existing `MatchesCatalog` / `NotFound` unchanged.
- `openai`: `CatalogProvider: "litellm"` prices a LiteLLM alias from the registered catalog; the same client without it lands on the gpt-4o default; a real OpenAI id is unaffected.
- `agent`: a three-call turn (two tool calls + text) reports `Usage` = final call and `TurnUsage` = sum incl. cache fields and cost; single-call turns report equal fields; the empty-response retry is included; `addUsage` unit.
- `pkg/agent/...` and `pkg/llm/...` green with `-race -short`.

## Embedder note

Loom Cloud already ships its own turn ledger and LiteLLM repricer (avmo-tera-cloud#418) that are correct with or without this change: its ledger counts at the provider boundary rather than reading `Response.Usage`, and its repricer overrides cost unconditionally, so neither double counts once loom does the right thing. A follow-up there can switch to `TurnUsage` + `CatalogProvider` and delete the local copies.


bsahitya (approved):
## Approve

Two real accounting bugs, cleanly fixed with tight tests. I checked out the branch, built with `-tags fts5`, and ran the catalog/openai/agent tests under `-race -short` — all green.

### What I verified

**Pricing from the registered catalog** (`LookupPricing`)
- Now mirrors `catalog.Lookup`: `NormalizeProvider` + `DefaultSource()` first, static `priceIndex` as fallback. The normalization is a nice side-fix — `azureopenai` now resolves against the static index too, which it didn't before.
- The "unpriced registered row falls through" guard (`CostPer_1MInputUsd > 0 || CostPer_1MOutputUsd > 0`) correctly prevents a metadata-only row from zeroing a price. Test covers it.

**Turn aggregation** (`TurnUsage`)
- `addUsage(&turnUsage, llmResp.Usage)` sits at the single post-call success site (agent.go:2512), so every billed call — tool-loop iterations, the empty-response retry, hygiene retries — is counted exactly once. Synthesis is added separately in `synthesizeFinalResponse`, no double count.
- The decision to keep `Response.Usage` as the final call is correct and load-bearing: it's what the persisted assistant row records as its own `TokenCount`/`CostUSD` (agent.go:1989-1990), and each tool-calling row already carries its own, so folding into `Usage` would double-count `Session.TotalCostUSD`. Good call documenting this in the field comment.
- `TurnParkedError.TurnUsage` and the synthesis-failure path (`Usage{}` + loop's `turnUsage`) are both consistent with that split.

Backward compatible (`Usage` semantics unchanged), no new locks/goroutines, and the CHANGELOG entry is accurate.

### Non-blocking notes

1. **`context.Background()` in `LookupPricing`** bypasses caller cancellation/timeout for a DB-backed source. It matches the existing `catalog.Lookup` precedent so I'm not blocking on it, but if a registered source ever does real I/O on the request path, both call sites will want a real ctx.
2. **MultiSource shadowing edge case**: if a registered chain places a *metadata-only* dynamic source ahead of a *priced* dynamic source (both before static), `MultiSource.Lookup` returns the first (unpriced) hit, and the fallthrough then re-checks only the static `priceIndex` — not the later priced dynamic source. Narrow (real chains front-load priced sources), but worth a one-line caveat on `LookupPricing` since the fallthrough intentionally re-queries static rather than continuing the chain.
3. Synthesis-failure drops the failed call's usage — fine, since usage isn't available on error.

None of these block. Nice work isolating the 1/N metering bias and the gateway overbilling.

PR #397, opened 2026-09-16
fix(llm): report tool-input stream deltas as activity

ilsiepotamus:
## Description

A streaming provider reports **tool-input deltas as stream activity**, and the agent turns them into a throttled, content-free progress event (`ProgressEvent.IsToolInputStream`), so a long tool-argument generation is no longer indistinguishable from a stalled stream.

Today every provider invokes the `TokenCallback` for text deltas only — OpenAI `Delta.Content`, Anthropic/Bedrock `text_delta` — and buffers `Delta.ToolCalls` / `input_json_delta` silently. That is correct for the callback's contract (its tokens become the visible partial response), but it means a model emitting a large tool argument (an HTML dashboard for a UI tool, a long document for a file tool) produces **no progress event at all** for the whole generation. Anything watching for activity — an idle-based turn deadline, a proxy inactivity timer — sees silence and cuts the turn. Loom Cloud hit exactly this: `create_ui_app` argument generations of 45–90s+ were killed by its sliding `SEND_MESSAGE_TIMEOUT` with `error reading stream: context deadline exceeded`, on three environments.

## Type of Change

- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🔧 Refactoring
- [x] 🧪 Test update

## Related Issues

Downstream: Teradata-PE/avmo-tera-cloud#441 (the embedder-side deadline keepalive; this PR supplies the honest signal it defers to).

## Changes Made

- `pkg/types/stream_activity.go`: `WithStreamActivity(ctx, fn)` / `NotifyStreamActivity(ctx)` — a context-carried hook, deliberately **not** a change to `StreamingLLMProvider`'s signature (nine providers plus downstream wrappers implement it). Re-exported from `pkg/llm/types` for providers.
- `pkg/types/types.go`: `ProgressEvent.IsToolInputStream` — a liveness pulse that carries no content and is always `Droppable`.
- Providers call `NotifyStreamActivity(ctx)` where they buffer tool input: anthropic (`input_json_delta`), openai + azureopenai (`Delta.ToolCalls`), bedrock legacy + SDK (`input_json_delta`), gemini (`FunctionCall` part), ollama (`Message.ToolCalls`). mistral delegates to the OpenAI-compatible stream; the Converse path has no delta loop.
- `pkg/agent/llm_retry.go` `chatWithStreaming`: installs the hook for the duration of `ChatStream` and emits `IsToolInputStream` events at most once per second (`toolInputActivityInterval`). Tool-input bytes never enter `PartialContent`. Text streaming is untouched.

## Testing

- `go build ./...` and `go vet` on all touched packages.
- `go test -race ./pkg/types/ ./pkg/agent/ ./pkg/llm/...` — pass.
- New tests: hook semantics (`pkg/types`); agent throttle + event shape — 50 rapid deltas → exactly one droppable, content-free `IsToolInputStream` event, and a text-only stream emits none (`pkg/agent/stream_activity_test.go`); one notification per `input_json_delta` with text still flowing through the token callback (anthropic); one per `tool_calls` chunk (openai).
- `golangci-lint run` on the touched packages.

### Test Checklist

- [x] All new and existing tests pass for the touched packages (`go test -race`)
- [x] Race detector shows zero race conditions on the touched packages
- [x] Code builds successfully (`go build ./...`)
- [ ] Full `just check` not run as a single command — build, vet, lint and `-race` tests were run directly on every touched package


bsahitya (approved):
## Review: Approve ✅

Solid, well-scoped fix. The core insight is correct: `TokenCallback` is text-only by contract, so tool-input JSON deltas produced zero progress signal, making a long `create_ui_app` argument generation indistinguishable from a stalled stream to any idle-based deadline or proxy inactivity timer.

### What I verified
- **Build**: `go build -tags fts5 ./...` passes.
- **Tests**: `go test -tags fts5 -race ./pkg/types/ ./pkg/llm/anthropic/ ./pkg/llm/openai/` pass, plus the new agent tests (`TestChatWithStreaming_ToolInputDeltasBecomeThrottledActivityEvents`, `TestChatWithStreaming_NoToolInputDeltasNoActivityEvents`) pass under `-race`. The only failure in `pkg/agent` was `TestDynamicToolDiscovery_Integration_Search`, an unrelated MCP filesystem-server timeout with no connection to this change.
- **Design**: The context-carried hook (`WithStreamActivity`/`NotifyStreamActivity`) deliberately avoids touching the `StreamingLLMProvider` signature — the right call given nine providers plus downstream wrappers implement it.
- **Threading**: The activity hook runs inline on the stream-reading goroutine, same as `tokenCallback`, sharing `progressCallback`, the `ctx.Done()` guard, and `Progress: 50`. No new concurrency surface — confirmed clean under the race detector.
- **Event shape**: `IsToolInputStream` carries no `PartialContent` and is always `Droppable`, reusing the existing heartbeat/backpressure semantics correctly, so JSON never leaks into visible text.
- **Provider coverage**: `NotifyStreamActivity` is placed exactly where each provider buffers tool input (anthropic `input_json_delta`, openai/azureopenai `Delta.ToolCalls`, bedrock legacy + SDK `input_json_delta`, gemini `FunctionCall`, ollama `Message.ToolCalls`). Mistral genuinely delegates to the OpenAI client's `ChatStream`, so it inherits the behavior — verified in `pkg/llm/mistral/client.go`.

### Non-blocking suggestions
- Provider-level tests exist only for anthropic and openai. bedrock (both paths), gemini, ollama, and azureopenai get the notification call but no dedicated test pinning "one notification per delta." Since they share the pattern this is acceptable, but a lightweight table test would guard against future regressions in those less-exercised paths.
- The throttle interval (`toolInputActivityInterval = 1s`) is fixed. If a downstream inactivity timer is ever configured tighter than ~1s this could be too coarse — worth a brief comment noting the assumed relationship to consumer-side deadlines.

Nothing here blocks merge. Nice, honest PR — the description accurately reflects the code.

